PR #18, opened 2026-02-26
Add prettier config and pre-commit hook for code formatting

dtodorowa:
**Motivation:**
To make UI PRs easier to review and remove visual clutter, we want to standardize the formatting when we push.

**Goal:**
Everyone should be able to use whatever formatting locally, but commits are going to be standardized

**Changes:**
1. Added .prettierrc.mjs: Single prettier config at the git root using createRequire to resolve plugins from the comhairle package's node_modules
2. .prettierignore: Excludes Rust, data, and build output from formatting
3. package.json: Added prettier, pretty-quick as devDeps + prepare script for auto-installing the git hook
4. install-hooks.mjs: Creates the pre-commit hook on pnpm install
5. package.json: Updated prettier-plugin-tailwindcss to ^0.7.2 (old one was incompatible with prettier 3.8)
6. Removed ui/.prettierrc and ui/packages/comhairle/.prettierrc -> single source of truth at the root now

**How it works:**
* On pnpm install in ui/: the prepare script auto-installs the pre-commit git hook
* On git commit: the hook runs pretty-quick --staged from the git root, formatting only staged files and re-staging them
* Existing scripts (pnpm format, pnpm lint in comhairle) still work — they find the root config by searching upward

Note: The tricky part was that the git root ≠ JS workspace root. pretty-quick must run from the git root so file paths match, but plugins need to be resolvable. The .prettierrc.mjs with createRequire bridges this gap by resolving plugin paths from the comhairle package where all peer dependencies (svelte, prettier, tailwindcss) are satisfied.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,19 +1,11 @@
-# Rust
-api/
-adaptors/
-comhairle_macros/
-data_model/
-target/
+# Ignore everything
+*
 
-# Data
-fixtures/
-pg_data/
-open-api-spec.json
-Cargo.lock
+# Allow ui directory and its contents
+!ui/
+!ui/**
 
-# Package managers
-pnpm-lock.yaml
-
-# Build output
+# Still ignore build output and lock files within ui
 ui/packages/comhairle/.svelte-kit/
 ui/packages/comhairle/build/
+pnpm-lock.yaml


.prettierignore
```diff
@@ -0,0 +1,19 @@
...
+adaptors/
+comhairle_macros/
+data_model/
+target/
+
+# Data
+fixtures/
+pg_data/
+open-api-spec.json
+Cargo.lock
```
rosswilson2306: @dtodorowa It might be better to try and ignore everything expect the ui directory so that if we add new rust modules or other files / directories etc to the monorepo, we don't have to update this file.

The docs say:
> .prettierignore uses gitignore syntax.

and the gitignore docs say:
> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again

So I think if you did something like:
```.prettierignore
# ignore everything first
. 

# cancel ignoring ui
!ui/
!ui/**
```
That might work. Not sure. Worth a try maybe?

PR #35, opened 2026-03-03
413 Add live jitsi calls for events

dtodorowa:
Motivation: We want to support live meetings
Changes: Added Jitsi support with a collapsible panel (Will probably change as designs come in)

Path to test: `/conversations/{id}/events/{id}/live`
Requirements: add `JITSI_DOMAIN` to your .env file

image
image


stuartlynn (approved):
This looks pretty good to get us going. Happy to merge so that @shuyanglin can start playing around with it and we can iterate from there.

PR #56, opened 2026-03-10
53/optional bot service

rosswilson2306:
Actions:

- refactor config so that bot service vars are optional
- refactor state so that bot service is optional based on config
- update usage of bot service
- test for creation of conversation with and without bot service

Motiviation:

- allow custom server instances to choose whether to add services

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -54,6 +54,14 @@ pub struct ComhairleState {
     pub jobs: Arc<JobQueues>,
 }
 
+impl ComhairleState {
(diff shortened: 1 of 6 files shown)


stuartlynn (changes requested):
Hey looks great. Just a couple of small things 

1. I think we should make a explicit struct for listing Services 
2. I think there is a way to reduce some of the code repetitions by having a convenience method on ComhairleState to get the BotService as a Result<BotService, ComhairleError> (or what ever the services is called) So we can just call that method and use ? to return the error

PR #67, opened 2026-03-12
48/conversation org region

rosswilson2306:
PR includes:

- migration to add region_id to workflow table
- migration to add organization_id to coversation
- filtering conversations by organization
- filtering regions by organization
- filtering organizations by region


rosswilson2306:
@stuartlynn 
Let me know what you think about the filtering in the organization and region models. It might not be too bad now that I've got it working. If we stick with the array columns I'll add a GIN index on `organization.regions`.

PR #90, opened 2026-03-18
85/convo consent field

rosswilson2306:
Actions:

- refactoring of Translatable derive macro to allow usage of Option<TextContentId> fields
- migration to add privacy_policy column to conversation table
- conversation model update to include privacy_policy
- refactoring of frontend translatable components to allow `value` to be null

Motiviation: 

- store translatable, rich text privacy policy on the conversation


api/src/models/conversation.rs
```diff
@@ -154,6 +160,12 @@ impl PartialConversation {
         if let Some(value) = &self.enable_qa_chat_bot {
             values.push((ConversationIden::EnableQaChatBot, (*value).into()))
         };
+        if let Some(value) = &self.privacy_policy {
+            values.push((ConversationIden::PrivacyPolicy, (*value).into()))
+        };
+        if let Some(value) = &self.faqs {
+            values.push((ConversationIden::Faqs, (*value).into()))
+        };
```
rosswilson2306, line 168: This is a little awkward how we have to handle this. As these translatable fields are optional we need to initialise the translation on the frontend, get the TextContentId from the response and update the conversation field with the TextContentId.

Luckily, there was already functionality in the FE TranslatableField component to run an onChange callback when the field doesn't have a TextContentId (unsaved) and once it is saved and has a TextContentId that callback is never run again.

I don't really like the idea of updating uuid fields from the frontend like this. Might be one for the next tech debt to see if there is a way to handle this all on the backend so that the translation is created and the conversation field is automatically updated. Might require rethinking the frontend a bit as well. Not sure.

dtodorowa:
frontend looks okay and makes sense to me, gonna approve and we can test on staging.
Do we want to release this while testing is going on? Probably okay? Wdyt?

dtodorowa (dismissed):

rosswilson2306:
> frontend looks okay and makes sense to me, gonna approve and we can test on staging. Do we want to release this while testing is going on? Probably okay? Wdyt?

Yeah, let's discuss in stand up. I still need Stuart to have a look at the macro changes. But this PR will change where the tutor bot displays so maybe we discuss with Shu about the user testing before merging.

stuartlynn (approved):
I think this looks fine for now. Lets get it merged so we can have it good to go for Waves and we can refine if we need to

PR #97, opened 2026-03-19
fix: redirect users back to conversation after signup

dtodorowa:
**the problem**
user clicks "take part" on a conversation → gets sent to signup → creates account → lands on homepage instead of the conversation they wanted to join

**ze fix**
added a backTo query param (that used to be there until I changed it recently - oops - sorry)

main change:
anonymous signup code page now uses backTo instead of hardcoded /conversations link

also did while i was here:
1. url-encoded backTo everywhere to prevent query string weirdness
2. switched from full urls to pathnames for cleaner params
3. threaded backTo through verification flow for non-anon signups
4. bonus**: login/signup forms now submit on enter key

is this overkill?
could simplify to only encode at entry points but then:
- breaks if someone adds a direct link without using helpers
- server-side redirects still need manual encoding anyway
- auth is critical, better paranoid than debugging "works 99% of the time" bugs

thoughts? keep it paranoid or simplify?

### DIFF
--- a/ui/packages/comhairle/src/lib/urls.ts
+++ b/ui/packages/comhairle/src/lib/urls.ts
@@ -1,59 +1,79 @@
-import { redirect } from "@sveltejs/kit"
-import { notifications } from "./notifications.svelte"
-import { goto } from "$app/navigation"
+import { redirect } from '@sveltejs/kit';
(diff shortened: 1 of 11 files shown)


stuartlynn (approved):
This looks good i think

PR #98, opened 2026-03-19
409/jitsi jwt auth

rosswilson2306:
Actions:

- refactor of `generate_jwt` to use builder pattern
- jitsi jwt route with required claims
- config update
- update frontend to use jwt authentication

### DIFF
--- a/api/src/config.rs
+++ b/api/src/config.rs
@@ -52,6 +52,14 @@ pub enum TranslatorConfig {
 }
 
 #[derive(Clone, Debug, Deserialize)]
+pub struct JitsiConfig {
+    pub jwt_app_id: String,
+    pub jwt_app_secret: String,
+    pub jwt_accepted_issuers: String,
+    pub jwt_accepted_audiences: String,
+}
+
+#[derive(Clone, Debug, Deserialize)]
 pub struct ComhairleConfig {
     pub database_url: String,
     pub jwt_secret: String,
@@ -68,4 +76,5 @@ pub struct ComhairleConfig {
     pub enable_rate_limiting: bool,
     pub heyform_url: String,
     pub polis_url: String,
+    pub jitsi: Option<JitsiConfig>,
 }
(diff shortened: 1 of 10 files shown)


api/src/config.rs
```diff
@@ -62,4 +71,5 @@ pub struct ComhairleConfig {
     pub default_knowledge_base_id: Option<String>,
     pub elicitation_bot_agent_id: Option<String>,
     pub whitelisted_domains: Option<Vec<String>>,
+    pub jitsi: Option<JitsiConfig>,
```
rosswilson2306: @stuartlynn Do we need a more generic name for this than jitsi? video_service or something?

api/src/routes/events.rs
```diff
@@ -154,6 +154,70 @@ async fn delete(
...
+        .ok_or(ComhairleError::NoVideoServiceConfigured)?;
+
+    let claims = VideoEventJwtClaims {
+        iss: &jwt_config.jwt_app_id,
+        aud: &jwt_config.jwt_app_id,
+        room: &video_meeting_id.to_string(),
+        context: VideoEventJwtContext {
+            user: VideoEventJwtUser {
+                name: user.username.as_deref(),
+                id: &user.id.to_string(),
```
rosswilson2306, line 205: @stuartlynn I've put this on the `event` router for now. The one thing I haven't done as add the `moderator` claim as I wasn't sure how we are going to handle this. I had originally thought we'd mark this on the event attendance, in which case I could move this route there. But I think you mentioned moderators being handled separately from attendees.

api/src/config.rs
stuartlynn: VideoCall service would be good I think to keep with the general pattern

api/src/routes/events.rs
stuartlynn: This is fine here just now.

stuartlynn (changes requested):

stuartlynn (approved):

PR #116, opened 2026-03-24
83/revisit steps

rosswilson2306:
Actions:

- migration to add can_revisit column to workflow_step table
- model update to allow can_revisit to be updated
- admin toggle on workflow step metadata to update can_revisit
- functionality to return workflow steps with active user's progress status for convenience

Motiviation:

- allow steps to be marked as revisitable


stuartlynn (approved):
Hey @rosswilson2306 Will the changes to the API client break anything? Just for the bloom frontends

PR #120, opened 2026-03-24
Waves: Add branding (colors and logo) & make buttons boxy

dtodorowa:
To test add this to your frontend env file
`PUBLIC_THEME="waves"`

image


stuartlynn (approved):
This is great. I think that our env variable should pick this up right?

ui/packages/comhairle/src/app.html
```diff
@@ -9,6 +9,10 @@
 			href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700&display=swap"
 			rel="stylesheet"
 		/>
+		<link
```
rosswilson2306, line 12: @dtodorowa Is it possible to wrap this and the other font `<link>`s in if conditions? Otherwise we are loading fonts that we're not using.

ui/packages/comhairle/src/lib/components/ComhairleLogo.svelte
```diff
@@ -30,17 +32,26 @@
...
 </script>
 
 {#snippet logoContent()}
-	<span
-		role="img"
-		aria-label="Comhairle Logo"
-		class="inline-block shrink-0 bg-current {sizeClass}"
-		style="-webkit-mask-image: url({logoSrc}); mask-image: url({logoSrc}); -webkit-mask-size: contain; mask-size: contain; -webkit-mask-repeat: no-repeat; mask-repeat: no-repeat; -webkit-mask-position: center; mask-position: center;"
-	></span>
+	{#if isWaves}
```
rosswilson2306, line 41: @dtodorowa Maybe a more scalable way of doing this is to have a mapper object, something like:
```ts
const themeLogoMapper = {
  waves: WavesLogo,
  scotGov: ScotGovLogo,
  fallback: ComhairleLogo,
};

...

<img src={themeLogoMapper[theme] ?? themeLogoMapper.fallback} ...
```

so that we don't end up with loads of else if blocks if this grows.

Probably not a concern for right now and I don't even know if the code I wrote works in svelte but it might be worth doing something like that if we have more cases like this in the future.

PR #124, opened 2026-03-25
fix casing of loginBehaviour on invite page so that correct buttons a…

rosswilson2306:
fix casing of loginBehaviour on invite page so that correct buttons appear


dtodorowa (approved):
You are the best Ross!! TYSMM <3

PR #125, opened 2026-03-25
Remove pid persistence from polis vote store

### LINKED ISSUE #123: Custom Polis: Use user id for the current user rather than getting it from local storage
Following up on https://github.com/crownshy/comhairle/pull/118/

we want to just grab the user_id from the current user (assuming it's the comhairle_id) rather than setting it in localstorage. If it's the PID we are saving then the polis initParticipation should return the currrent on for the comhairle user automatically

Why:
localstorage can get out of sync with the user.
we should have a single source of truth for it and that should probably be the currently logged in user

dtodorowa:
- do not persist pid to localStorage
- ensure pid is fetched before attempting to fetch statements


stuartlynn (approved):
This looks good. Going to test this ahead of the waves launch to ensure that everything is saving properly

PR #137, opened 2026-03-25
95 polis client refactor

rosswilson2306:
Actions:

+ refactor wiki_poll_service and polis client implementation to not rely on reqwest cookie_store but instead return formatted auth cookie and directly pass to subsequent polis admin requests to be added as a Cookie header
+ update polis client tests to make real requests but mark as ignored

Motivation:

+ reliance on reqwest cookie_store automatically picking up auth cookies from login response could create a potential race condition on subsequent requests at scale
+ use real requests in polis client tests to correctly test auth flow

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1312,6 +1312,7 @@ dependencies = [
  "clap",
  "comhairle-macros",
  "config",
+ "cookie",
  "dashmap",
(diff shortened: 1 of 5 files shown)


stuartlynn (approved):
Looks good

PR #138, opened 2026-03-25
Fix debouncing bug

### LINKED ISSUE #71: When editing the metadata text, it often reverts back to what it was before as you type
The metadata text input seems to be not correctly debouncing on save. Often when I type into it, it jumps between what I am typing and the previous text.

dtodorowa:
**Motivation:**
Do not overwrite what the user has typed in the field. Improve UX by providing visual feedback during autosave

**Actions:**
+ prevent dialog close on outside click/escape, use explicit close handler
+ add ui feedback to the user so they know something is still saving (autosave status indicator)
+ add save status state ('idle', 'saving', 'saved') to TranslatableField
+ display loading spinner when saving translations
+ display checkmark when save completes successfully
+ auto-hide saved status after 2 seconds
+ show status indicators alongside translation badges


stuartlynn (approved):
Awesome!

PR #140, opened 2026-03-26
136/tutor bot tab

rosswilson2306:
Actions:

- move tutor bot into own tab in the conversation support sidebar
- fix styling in light / dark mode


dtodorowa:
Looks great and I like the tab thingy! It's less in your face

Random thoughts while I was testing:

at first I thought I would prefer the input field to be at the very bottom but then we end up with a lot of negative space in the blue area between the first message and the input field.
image

I thought what you have is probably the better solution. 

When i started testing it tho, it started expanding and then i lost the input field
image

maybe we should give the chat a min and max height? Not super sure about an optimal solution? What do you think?
possible thing is to expand it to start with, so it doesn't grow as the chat gets longer?

dtodorowa (dismissed):

PR #152, opened 2026-03-30
Show thank you page in preview

### LINKED ISSUE #132: Add thank you page to "preview flow"
When we preview a conversation, we should show the thank you page in the end.

Current behaviour: exits conversation after last step is completed

dtodorowa:


ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte
```diff
@@ -101,7 +101,7 @@
 	}
 
 	function goToThankYouPage() {
-		goto(thank_you_page(conversation.id, workflowStep.id));
+		goto(thank_you_page(conversation.id, workflow_id, !conversation.isLive));
```
Comment (rosswilson2306) [inline ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte:104]: @dtodorowa If we want to show the thank you page in preview as well as live mode then we probably want to configure the `thank_you_page` function? https://github.com/crownshy/comhairle/blob/9c451b306c089930e9d3d7f461a8b6bf2b3863cd/ui/packages/comhairle/src/lib/urls.ts?plain=1#L32

It now won't require the `isPreview` param as we should reach the thank you page regardless, is that right?

rosswilson2306 (commented):
Left a comment / question?

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte:104]: I might be misunderstanding what `isPreview` does. Does it just configure the path to include `/preview` and not redirect?

Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte:104]: If so, it looks like `/preview` will be added to the path if the conversation is live?

stuartlynn (approved):

PR #174, opened 2026-03-31
Polis report data api

stuartlynn:
Adds an API to get access to the polis report data in a more sane format.

Adds ```/api/tools/polis/report_data?workflow_step_id=``` endpoint.

### DIFF
--- a/api/src/tools/polis.rs
+++ b/api/src/tools/polis.rs
@@ -1,6 +1,9 @@
 use std::sync::Arc;
 
-use aide::axum::{routing::post_with, ApiRouter};
+use aide::axum::{
+    routing::{get_with, post_with},
+    ApiRouter,
+};
 use async_trait::async_trait;
 use axum::{
     extract::{Json, Query, State},
@@ -90,6 +93,15 @@ impl ToolImpl for PolisTool {
                         .description("Logs into Polis as admin and returns session cookie")
                 }),
             )
+            .api_route(
+                "/polis/report_data",
+                get_with(get_report_data, |op| {
+                    op.id("PolisGetReportData")
+                        .tag("Tools")
+                        .summary("Get Polis report data for a workflow step")
+                        .description("Fetches the polis data export for a given workflow step")
+                }),
+            )
             .with_state(state.clone())
     }
 }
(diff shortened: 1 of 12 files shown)


api/src/tools/polis.rs
```diff
@@ -118,11 +130,101 @@ pub enum PolisError {
...
     pub workflow_step_id: Uuid,
 }
 
+#[derive(Serialize, Deserialize, JsonSchema)]
+struct ReportDataQuery {
+    pub workflow_step_id: Uuid,
+}
+
+/// Gets the polis report data for a workflow step
+async fn get_report_data(
```
rosswilson2306, line 198: @stuartlynn Could add tracing here.

api/src/tools/polis.rs
```diff
@@ -118,11 +130,101 @@ pub enum PolisError {
...
+    State(state): State<Arc<ComhairleState>>,
+    Query(ReportDataQuery { workflow_step_id }): Query<ReportDataQuery>,
+) -> Result<(StatusCode, Json<WikiPollReport>), ComhairleError> {
+    let workflow_step = models::workflow_step::get_by_id(&state.db, &workflow_step_id).await?;
+
+    println!("{workflow_step:#?}");
```
rosswilson2306, line 204: @stuartlynn print statement left behind from debugging?

api/src/tools/polis.rs
```diff
@@ -118,11 +130,101 @@ pub enum PolisError {
...
+
+    let config = match (workflow_step.tool_config, workflow_step.preview_tool_config) {
+        (Some(ToolConfig::Polis(config)), _) => config,
+        (None, ToolConfig::Polis(config)) => config,
+        _ => return Err(ComhairleError::WorkflowStepHasWrongType("Polis".into())),
+    };
+
+    // Login as admin to access the data export
+    state
```
rosswilson2306: @stuartlynn Do we not need to return the auth cookies from this and add to the next request? If not then we might not need to login at all as we're no longer storing the cookies on the polis client automatically.

api/src/wiki_poll_service/polis_service.rs
```diff
@@ -29,6 +103,275 @@ impl PolisClient {
...
+        );
+
+        info!("Getting PCA data from {url}");
+
+        let response = self.client.get(&url).send().await.map_err(|e| {
+            warn!("Failed to get PCA data: {e}");
+            PolisError::FailedToGetComments(format!("Failed to get PCA data: {e}"))
+        })?;
+
+        info!("Response {response:#?}");
```
rosswilson2306: @stuartlynn Can these debugging logs be removed?

api/src/wiki_poll_service/polis_service.rs
```diff
@@ -193,6 +536,20 @@ impl WikiPollService for PolisClient {
...
+
+    async fn get_report_data(&self, poll_id: &str) -> Result<WikiPollReport, WikiPollServiceError> {
+        info!("Getting full report data for poll_id: {poll_id}");
+
+        // Fetch all the data that powers the report page
+        let math_pca = self.get_math_pca(poll_id).await?;
+        let comments_data = self.get_comments_with_voting(poll_id).await?;
+
+        info!("math pca {math_pca:#?}");
+        info!("comments {comments_data:#?}");
```
rosswilson2306: @stuartlynn Same question here, can the debug logs be removed?

api/src/tools/polis.rs
stuartlynn: I think your right that we don't need to be logged in to get this data.

rosswilson2306 (approved):

PR #208, opened 2026-04-08
206 heyform launch

rosswilson2306:
Actions:

+ add new graphql query for updateFormSchemas
+ fix FORM_DETAIL_QUERY
+ update launch method to create new form and credentials, get preview form data and update live form with preview data

Motivation:

+ allow updating form fields data
+ FORM_DETAIL_QUERY should use formDetail instead of form
+ allow preview forms and live forms to be on isolated accounts similar to polis

### DIFF
--- a/adaptors/heyform-rust-sdk/src/client.rs
+++ b/adaptors/heyform-rust-sdk/src/client.rs
@@ -245,6 +245,25 @@ impl HeyFormClient {
         Ok(embed_url)
     }
 
+    pub async fn update_poll(&self, form_id: &str, form: Form) -> Result<bool> {
+        let variables = json!({
+            "input": {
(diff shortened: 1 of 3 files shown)


rosswilson2306:
@stuartlynn This seems to work the same as the polis tool now. Creates a new heyform survey with new credentials, gets the form data from the preview survey then updates the new, live form with the same data. 

I've tested it out with every form element type that heyform has and seems to be working okay.

stuartlynn:
This looks good.  We should check that things like hidden_field_ids are cloned when we use this method but I think this looks good otherwise

stuartlynn (approved):
Lets make sure that hidden fields are cloned with this approach  but if they are lets merge

PR #220, opened 2026-04-13
Adding a button to download a list of user emails from admin

stuartlynn:
This PR adds a button for admins to download a list of email address for users who have opted in to the admin page.

### DIFF
--- a/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte
+++ b/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte
@@ -6,6 +6,8 @@
 	import GenderComparison from '$lib/components/GenderComparison.svelte';
 	import GeoComparison from '$lib/components/GeoComparison/GeoComparison.svelte';
 	import * as Card from '$lib/components/ui/card';
+	import { Download } from 'lucide-svelte';
+	import { Button } from '$lib/components/ui/button';
 
 	import type { PageProps } from './$types';
 	import { BreadcrumbItem } from '$lib/components/ui/breadcrumb';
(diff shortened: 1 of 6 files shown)


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte
```diff
@@ -14,6 +16,33 @@
 	let { data }: PageProps = $props();
 	let { workflowSteps, workflowStats } = data;
 
+	async function downloadContacts() {
```
Comment (rosswilson2306) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte:16]: @stuartlynn This might be doing something I'm missing, but for a download route with `Content-Disposition: attachment;` you can just have a link with an href to the endpoint. Eg https://github.com/crownshy/comhairle/blob/685c9cb251448e5705075182d15896fcb0b876ba/ui/packages/comhairle/src/lib/components/KnowledgeBase/ParsedFileBadge.svelte?plain=1#L72-L78

Is this just for better error handling?

rosswilson2306 (dismissed):

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte
Comment (dtodorowa) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte:16]: i vote for the try/catch just so we get some info when something goes wrong. 
Not urgent at all, but is it worth starting to think about logging errors?
Could help us catch when stuff is broken in prod environments

Comment (stuartlynn) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/monitor/+page.svelte:16]: Nice yeah just changed this up

PR #224, opened 2026-04-15
Waves Report: Update styling and copy to match latest design

dtodorowa:
+ add executive summary section with structured content about agreement, differences, and next steps
+ add "Back to Top" navigation option with up arrow icon
+ restructure engagement statistics into 3x2 grid layout
+ update demographic section formatting to use inline text instead of separate paragraphs
+ remove bridging strength labels from statement cards
+ make bee swarm plot look more like polis


ui/packages/comhairle/src/lib/components/report/beeswarm/BeeswarmDots.svelte
```diff
@@ -77,8 +77,19 @@
 				b.y = Infinity;
 				do {
 					if (a) {
```
dtodorowa, line 79: this is to make it look like polis. Before and after attached:

Image

Image

PR #226, opened 2026-04-15
polis report data fixes

stuartlynn:
This PR fixes a bug where we calculated the wrong number of participants in groups and in total n the polis report. This stems from the PCA calculation in polis being capped at 100 individuals for small polls and the resulting data returned from the API showing only these. 

Instead we use the summary stats to get accurate info.

This also filters out the moderated statements

- **Updating the polis report with more accurate info**
- **Getting the right values to work around the PCA limit**


rosswilson2306 (approved):
There's a print statement left behind but not a big deal if this is urgent. Approving.

PR #233, opened 2026-04-16
Websocket refactor

stuartlynn:
This PR does a lot of work fixing up the websocket implementation 

We now have 
- A way to have different backend services that use the websocket connection 
- A setting up two services, one to monitor workflow processes another to manage notifications 
- A frontend service for websockets that creates a single connection but lets other systems hook into it 
- Frontend services for workflow monitoring and notifications 

The code is mostly refactored to use this with the exception of the workflow monitoring which is coming soon.

### DIFF
--- a/ui/packages/comhairle/src/lib/services/notifications.svelte.ts
+++ b/ui/packages/comhairle/src/lib/services/notifications.svelte.ts
@@ -1,0 +1,112 @@
+import { ws } from '$lib/api/websockets.svelte';
+import { notifications, type NotificationPriorities } from '$lib/notifications.svelte';
+
+interface ServerNotification {
+	id: string;
+	title: string;
+	message: string;
+	level: string;
+}
+
+export class NotificationService {
+	private _unreadCount = $state(0);
+	private isListening = false;
+
+	constructor() {
+		this.setupListeners();
+		// Request initial count after a short delay to ensure WebSocket is connected
+		setTimeout(() => this.requestUnreadCount(), 100);
+	}
+
+	get unreadCount() {
+		return this._unreadCount;
+	}
+
+	private setupListeners() {
+		if (this.isListening) return;
+		this.isListening = true;
+
+		// Listen for new notifications from server
+		ws.on('custom', (payload) => {
+			if (payload.event === 'notification:new') {
+				this.handleNewNotification(payload.data as ServerNotification);
+			} else if (payload.event === 'notification:unread_count') {
+				this._unreadCount = payload.data.count;
+			} else if (payload.event === 'notification:marked_read') {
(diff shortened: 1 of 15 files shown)


rosswilson2306 (approved):

ui/packages/comhairle/src/lib/services/notifications.svelte.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export class NotificationService {
+	private _unreadCount = $state(0);
+	private isListening = false;
+
+	constructor() {
+		this.setupListeners();
+		// Request initial count after a short delay to ensure WebSocket is connected
+		setTimeout(() => this.requestUnreadCount(), 100);
```
rosswilson2306, line 18: @stuartlynn Is this going to be called from a svelte component? If so I think it has the potential to cause a memory leak if the timeout isn't cleared on unmount. We'd maybe need an `onDestroy()` method that can be called on unmount to clear the timeout.

ui/packages/comhairle/src/lib/services/notifications.svelte.ts
```diff
@@ -0,0 +1,112 @@
...
+		ws.sendCustom('notification:mark_all_read', {});
+	}
+
+	// Request current unread count
+	requestUnreadCount() {
+		if (ws.connectionStatus === 'connected') {
+			ws.sendCustom('notification:get_unread_count', {});
+		} else {
+			// Retry when connected
+			setTimeout(() => this.requestUnreadCount(), 1000);
```
rosswilson2306, line 106: Same question here. This might be alright if this class is only instantiated once but otherwise might create a memory leak if the timeout isn't cleared at some point.

ui/packages/comhairle/src/lib/services/notifications.svelte.ts
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+}
+
+// Singleton instance
+export const notificationService = new NotificationService();
```
rosswilson2306, line 112: Just saw this, so I guess it is only instantiated once?

ui/packages/comhairle/src/lib/services/notifications.svelte.ts
stuartlynn: It's not. We are using a singleton method here where we instantiate and return an instance of the class on the export. This means that this gets called once on creation and never again. 

Also I think your thinking of setInterval. setTimouts clear themselves I think

ui/packages/comhairle/src/lib/services/notifications.svelte.ts
stuartlynn: Same as above

PR #248, opened 2026-04-21
Video call state

stuartlynn:
Adds a websocket service for video calls to coordinate a given call. There is a backend implementation and a frontend service that work together

It syncs the room state to all participants automatically 
It gives events for users joining / leaving 
It automatically generate breakout rooms and gives a function to regenerate them if needed. 
It allows the moderator to start the call / end the call 
It allows the moderator to set the current agenda item.

### DIFF
--- a/api/src/websockets/handlers/notifications.rs
+++ b/api/src/websockets/handlers/notifications.rs
@@ -5,7 +5,6 @@ use tracing::info;
 use uuid::Uuid;
 
 use crate::{
-    error::ComhairleError,
     models::{notification, notification_delivery},
     websockets::{
         error::WebsocketError, messages::WebSocketMessage, WebSocketConnection,
@@ -15,7 +14,19 @@ use crate::{
 };
 
 #[derive(Error, Debug)]
-pub enum NotificationWSError {}
+pub enum NotificationWSError {
+    #[error("Invalid or missing delivery ID")]
+    InvalidDeliveryId,
+
+    #[error("User not authorized to access this notification")]
+    Unauthorized,
+
+    #[error("Database error: {0}")]
+    DatabaseError(String),
+
+    #[error("Failed to send WebSocket message: {0}")]
+    SendError(String),
+}
 
 /// Handler for notification-related WebSocket messages.
 ///
@@ -129,7 +140,7 @@ impl NotificationMessageHandler {
         title: &str,
         message: &str,
         level: &str,
-    ) -> Result<(), ComhairleError> {
+    ) -> Result<(), NotificationWSError> {
         let ws_message = WebSocketMessage::Custom {
             event: "notification:new".to_string(),
             data: serde_json::json!({
@@ -140,7 +151,11 @@ impl NotificationMessageHandler {
             }),
         };
 
-        let sent_count = state.websockets.send_to_user(user_id, &ws_message).await?;
+        let sent_count = state
+            .websockets
+            .send_to_user(user_id, &ws_message)
+            .await
+            .map_err(|e| NotificationWSError::SendError(e.to_string()))?;
 
         if sent_count > 0 {
             info!(
@@ -212,7 +227,7 @@ impl NotificationMessageHandler {
         notification_type: notification::NotificationType,
         context_type: notification::NotificationContextType,
         context_id: Option<&Uuid>,
-    ) -> Result<notification::Notification, ComhairleError> {
+    ) -> Result<notification::Notification, NotificationWSError> {
         // Create the notification in the database
         let new_notification = notification::CreateNotification {
             title: title.to_string(),
@@ -222,7 +237,9 @@ impl NotificationMessageHandler {
             context_id: context_id.copied(),
         };
 
-        let created_notification = notification::create(&state.db, &new_notification).await?;
+        let created_notification = notification::create(&state.db, &new_notification)
+            .await
+            .map_err(|e| NotificationWSError::DatabaseError(e.to_string()))?;
 
         // Create delivery record
         let delivery = notification_delivery::CreateNotificationDelivery {
@@ -231,7 +248,9 @@ impl NotificationMessageHandler {
             delivery_method: Some(notification_delivery::DeliveryMethod::InApp),
         };
 
-        let _created_delivery = notification_delivery::create(&state.db, &delivery).await?;
+        let _created_delivery = notification_delivery::create(&state.db, &delivery)
+            .await
+            .map_err(|e| NotificationWSError::DatabaseError(e.to_string()))?;
 
         // Send via WebSocket if user is connected
         Self::send_notification_to_user(
@@ -264,19 +283,19 @@ impl WebSocketMessageHandler for NotificationMessageHandler {
             WebSocketMessage::Custom { event, data } if event.starts_with("notification:") => {
                 match event.as_str() {
                     "notification:mark_read" => {
-                        self.handle_mark_read(data, connection, state).await
+                        self.handle_mark_read(data, connection, state).await?
                     }
                     "notification:mark_all_read" => {
-                        self.handle_mark_all_read(connection, state).await
+                        self.handle_mark_all_read(connection, state).await?
                     }
                     "notification:get_unread_count" => {
-                        self.handle_get_unread_count(connection, state).await
+                        self.handle_get_unread_count(connection, state).await?
                     }
                     _ => {
                         info!("Unhandled notification event: {}", event);
-                        Ok(())
                     }
                 }
+                Ok(())
             }
             _ => Ok(()),
         }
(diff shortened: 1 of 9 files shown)


api/src/websockets/handlers/notifications.rs
```diff
@@ -1,15 +1,22 @@
...
-    websockets::{messages::WebSocketMessage, WebSocketConnection, WebSocketMessageHandler},
+    websockets::{
+        error::WebsocketError, messages::WebSocketMessage, WebSocketConnection,
+        WebSocketMessageHandler,
+    },
     ComhairleState,
 };
 
+#[derive(Error, Debug)]
+pub enum NotificationWSError {}
```
rosswilson2306: @stuartlynn It doesn't look like this is being used?

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+#[derive(Serialize, Debug)]
+pub struct VideoCallState {
+    /// Unique identifier for the video call
+    pub video_call_id: Uuid,
+    /// Current status of the call (Waiting, InProgress, or Ended)
+    pub status: VideoCallStatus,
+    /// Map of participant user IDs to their participant data
+    pub participants: HashMap<Uuid, VideoCallParticipant>,
+    /// List of breakout room assignments
+    pub breakout_rooms: Vec<BreakoutRoomAssignments>,
```
rosswilson2306, line 50: @stuartlynn Is there any benefit in trying to store this as HashMap with the breakout room's id as the key?

api/src/websockets/handlers/notifications.rs
rosswilson2306: I guess that's not really what this PR is about.

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+    UnauthorizedStateChange,
+}
+
+impl VideoCallState {
+    /// Creates a new video call state with default values.
+    ///
+    /// # Arguments
+    ///
+    /// * `video_call_id` - Unique identifier for this video call
+    /// * `jitsi_call_id` - Jitsi meeting room identifier
```
rosswilson2306, line 105: @stuartlynn What is the difference between these? We have `video_meeting_id` on the event model, which is used for the jitsi id https://github.com/crownshy/comhairle/blob/2d2a170bce2081865ea0691ace784a9e10e49c1c/api/src/models/event.rs?plain=1#L36
https://github.com/crownshy/comhairle/blob/2d2a170bce2081865ea0691ace784a9e10e49c1c/ui/packages/comhairle/src/routes/(public)/conversations/%5bconversation_id%5d/events/%5bevent_id%5d/live/+page.svelte?plain=1#L28

rosswilson2306: Also, we've tried to keep the config naming non-jitsi related so maybe not naming things `jitsi_...` here would be better https://github.com/crownshy/comhairle/blob/2d2a170bce2081865ea0691ace784a9e10e49c1c/api/src/config.rs?plain=1#L74

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+/// Data structure for broadcasting messages to call participants.
+#[derive(Serialize, Deserialize, Debug)]
+struct BroadcastMessageData {
+    pub event_id: Uuid,
+    pub message: String,
+}
+
+/// Data structure for changing the call state.
+#[derive(Serialize, Deserialize, Debug)]
+struct ChangeCallStateData {
```
rosswilson2306: This is a pernickety, semantic thing but I think `ChangeCallStatusData` makes more sense here as it only changes the status part of the state. Each one of the websocket operations will change the `state` in some way.

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// - Failed to acquire lock on the call state
+    /// - Broadcasting the state fails
+    pub async fn set_agenda_item(
+        &self,
+        event_id: &Uuid,
+        data: &serde_json::Value,
+        _connection: &WebSocketConnection,
+        state: &Arc<ComhairleState>,
+    ) -> Result<(), VideoCallWSError> {
+        let agenda_update: SetAgendaItemData = serde_json::from_value(data.clone())?;
```
rosswilson2306, line 259: @stuartlynn Why not just make `data: SetAgendaItemData`?

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+    ///
+    /// Returns an error if JSON deserialization fails or lock acquisition fails.
+    pub async fn broadcast_message(
+        &self,
+        event_id: &Uuid,
+        data: &serde_json::Value,
+        connection: &WebSocketConnection,
+        state: &Arc<ComhairleState>,
+    ) -> Result<(), VideoCallWSError> {
+        let broadcast_data: BroadcastMessageData = serde_json::from_value(data.clone())?;
```
rosswilson2306, line 297: Same question here.

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+        connection: &WebSocketConnection,
+        state: &Arc<ComhairleState>,
+    ) -> Result<(), VideoCallWSError> {
+        let leave_data: UserLeaveData = serde_json::from_value(data.clone())?;
+
+        let user_id = connection.user.id;
+
+        // Remove the participant from the video call
+        self.with_video_call_state_mut(&leave_data.event_id, |call| {
+            call.participants.remove(&user_id);
```
rosswilson2306, line 471: @stuartlynn We don't need to also remove their id from the breakout room lists?

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// Only users with role "moderator" or "facilitator" can change call state.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - JSON deserialization fails
+    /// - User is not authorized (`UnauthorizedStateChange`)
+    /// - Failed to acquire lock on the call state
+    /// - Broadcasting fails
+    pub async fn change_call_state(
```
rosswilson2306: Same comment as before, I think this should be `change_call_status` as all the previous methods also change the `state` in some way.

api/src/websockets/handlers/video_call.rs
```diff
@@ -0,0 +1,733 @@
...
+
+        let user_id = connection.user.id;
+        let max_users = assignment_data.max_users_per_room;
+
+        // Check authorization and assign breakout rooms
+        let authorized = self
+            .with_video_call_state_mut(&assignment_data.event_id, |call| {
+                // Check if the user is authorized (moderator or facilitator)
+                if let Some(participant) = call.participants.get(&user_id) {
+                    if participant.role == "moderator" || participant.role == "facilitator" {
```
rosswilson2306, line 588: We should probably make this an enum at some point.

api/src/websockets/handlers/video_call.rs
rosswilson2306: I can see in the code below that a vector works quite nicely so feel free to ignore this.

api/src/websockets/handlers/video_call.rs
rosswilson2306: Ignore this, I can see why in the implementation of `handle_message`

ui/packages/comhairle/src/lib/services/videoCallService.svelte.ts
```diff
@@ -0,0 +1,180 @@
+import { ws } from '$lib/api/websockets.svelte';
+
+// Types matching backend structures
+export type VideoCallStatus = 'Waiting' | 'InProgress' | 'Ended';
```
rosswilson2306, line 4: It might be nice in the future to see if we can generate these the same as the openapi types so they're always in sync with the backend. For another time though.

ui/packages/comhairle/src/lib/services/videoCallService.svelte.ts
```diff
@@ -0,0 +1,180 @@
...
+	leaveCall(eventId: string) {
+		ws.sendCustom('video_call:user_left', {
+			event_id: eventId
+		});
+		// Clear local state when leaving
+		this._currentCallState = null;
+	}
+
+	// Change the call state (moderator/facilitator only)
+	changeCallState(eventId: string, status: VideoCallStatus) {
```
rosswilson2306, line 108: Same as before, maybe `changeCallStatus` is clearer? Happy to be disagreed with though.

rosswilson2306 (commented):
Left some comments / questions but quite a few can probably be ignored. The only one to definitely check is removing a user. I think this should also remove them from the breakout rooms?

api/src/websockets/handlers/notifications.rs
stuartlynn: Might as well do that migration here. Trying to get better at using local error structs

api/src/websockets/handlers/video_call.rs
stuartlynn: So perhaps the variable name wasn't quite right here. These are breakout room assignments rather than people who are currently in a breakout room. We want I think to handle the case where a user's connection goes down, they log back in again and they should be directed to the same breakout room. 

So I dont think we want to remove them from the breakout rooms assignments at this point

ui/packages/comhairle/src/lib/services/videoCallService.svelte.ts
stuartlynn: Yeah totally. Not sure how to do that though. We could go the RCP route potentially with messages being passed over websockets.

api/src/websockets/handlers/video_call.rs
stuartlynn: Yeah. I think that in the DB it's just stored as a string but we should change that in  a separate PR. Opening an issue here:  https://github.com/crownshy/comhairle/issues/260

ui/packages/comhairle/src/lib/services/videoCallService.svelte.ts
stuartlynn: Opened an issue to look into this further https://github.com/crownshy/comhairle/issues/261

api/src/websockets/handlers/video_call.rs
stuartlynn: I think your right. Should be changed.

api/src/websockets/handlers/video_call.rs
stuartlynn: I removed the jtisi one. I couldn't remember if we had a one to one correspondence

api/src/websockets/handlers/notifications.rs
stuartlynn: Migrated it to use the local error

PR #274, opened 2026-04-30
257/event invite links

rosswilson2306:
Actions:

- new OTP user type
- migration to add optional `event_id` to invite table
- model methods and routes for event invites
- admin ui for event invites
- route handler for auto-registering a user to an event via an invite, creating a new OTP user if none exist with target email
- event invite flow on FE

Motivation:

- auto-register / signup users to events


rosswilson2306:
https://github.com/crownshy/comhairle/pull/274

stuartlynn (approved):

PR #287, opened 2026-05-07
try to remove watermark (unsuccessfully lol)

dtodorowa:
SHOW_JITSI_WATERMARK + SHOW_WATERMARK_FOR_GUESTS already there, watermark still visible :/ 
image

server-side interface_config.js probably overrides client?

Two paths to try?
1. SSH into Jitsi host, edit /etc/jitsi/meet/<domain>-config.js or /usr/share/jitsi-meet/interface_config.js, set same flags false, reload nginx.
2. Check Jitsi server got interfaceConfigOverwrite whitelist enabled. Apparently new Jitsi versions ignore some client overrides unless server allow.

### DIFF
--- a/ui/packages/comhairle/src/lib/components/JitsiMeet/JitsiMeet.svelte
+++ b/ui/packages/comhairle/src/lib/components/JitsiMeet/JitsiMeet.svelte
@@ -121,6 +121,11 @@
 			interfaceConfigOverwrite: {
 				SHOW_JITSI_WATERMARK: false,
 				SHOW_WATERMARK_FOR_GUESTS: false,
+				SHOW_BRAND_WATERMARK: false,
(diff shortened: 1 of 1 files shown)


rosswilson2306:
@dtodorowa 
> SSH into Jitsi host, edit /etc/jitsi/meet/-config.js or /usr/share/jitsi-meet/interface_config.js, set same flags false, reload nginx.

We shouldn't have to do this. I'm pretty sure you can configure the `interface_config.js` settings in the helm_charts. See https://github.com/crownshy/helm_charts/blob/3b190fd704988462eb80d9dabee3729004123ffd/comhairle_jitsi/values/values.yml?plain=1#L119

dtodorowa:
closing but we should still do at some point

PR #291, opened 2026-05-07
275/media uploads

rosswilson2306:
Actions:

- migration to add media table
- media model and methods
- media router and endpoints using bulk_storage_service
- refactor bulk_storage_service to be optional and bucket name to come from config

Motiviation:

- allow image and other media uploads

Still to do:

- refactor conversations to use references to media table instead of image_url
- admin ui for uploading images for conversations


rosswilson2306:
@stuartlynn 

Sorry for the length of this! I could do with your thoughts on how to handle the conversation images. 

The first thing to note is that I've not stored the constructed urls on the media table as this would contain references to aws / s3. I thought if we ever had to change to a different provider we could probably keep the same bucket / store name and file structure so only storing:
```
store_name: String,
storage_key: String,
```
without the `.s3.amazonaws.com` subdomain gave us that flexibility. So the urls will need to be constructed in the rust code. If you disagree with this I'm happy to change it (it might make the subsequent issues a little easier to handle).

With the new `media` table, I think we would replace `image_url` with `image` on the conversation table, eg
```sql
ALTER TABLE conversation
ADD COLUMN image uuid REFERENCES media(id)
DROP COLUMN image_url;
```

However, we'll want our read endpoints to return the constructed image urls rather than an id. As we already have `Conversation`, `LocalizedConversation`, `ConversationDto` and `LocalizedConversationDto` I don't really want to create a new class of types for models with constructed media urls instead of raw ids. 

I had thought the quickest solution would be to just do this and create the new types just for the conversation, something like:
```
pub struct LocalizedConversationExt {
  #[serde(flatten)]
  conversation: LocalizedConversation,
  image_url: Option<String>
}
```

Two things that might complicate this are:
1. the media record doesn't contain the constructed url so we would need a way of passing in the `.s3.amazonaws.com` subdomain to the join in the query.
2. We might need to order this before the `query_to_localization` in which case it would be `conversation: Conversation` instead of `conversation: LocalizedConversation`  (I might wrong about this so might not be relevant)

This doesn't feel like a clean, future-proof solution anyway.

At the moment, I think it is only the conversations that have images but in case we add media to other models I thought it might be good to extract away the join functionality similar to what you did with the `Translatable` macro: 
- create a `MediaId` type that is essentially a wrapper around a `Uuid`
- have a macro that looks for fields of type `MediaId` or `Option<MediaId>` and handles the joins automatically and constructs the s3 url for that field.

Like I mentioned, I would prefer not to complicate things further with an additional class of types, eg.
```
Conversation
LocalizedConversation
MediaConversation
LocalizedMediaConversation
```
and the associated dto types and from conversions for each of those. So rather than an additional macro maybe it would be better to add this logic into the Translatable macro (and maybe rename it) so that:
```rust
pub struct Conversation {
  id: Uuid,
  title: Uuid,
  image: Option<Uuid>,
  ...
}
```
becomes
```rust
pub struct LocalizedConversation {
  id: Uuid,
  title: String,
  image: Option<String> // constructed url
}
```

One issue we'd still have is that in order to construct the image url from:
```
store_name: comhairle-media
storage_key: images/test.jpg
```
We still need to hard code `.s3.amazonaws.com` somewhere in the macro, whether we do it as one macro or two different macros. We could maybe have a method on the bulk_storage_service to construct the urls from the `store_name` and `storage_key` and implement it for s3 but I don't know how we would access this from a macro...

Or we just store the url in the media table when a new file is uploaded and deal with any provider migration issues if and when that happens. 


It feels to me like storing the s3 urls on the media table simplifies things, even if it creates a little vendor lock-in, and refactoring / renaming the Translatable macro to handle the media joins keeps the number of types / dtos simpler as well. 

I'll see if I can get that working on a new branch and make a separate PR to this one and see what you think.

rosswilson2306:
@stuartlynn Regarding the last comment, I've had a crack at my proposed solution, ie store the s3 url in the media table and refactor the `Translatable` macro to create joins and resolve fields such as `image -> media(id)` to `image -> media(url)`.

https://github.com/crownshy/comhairle/pull/299/changes

It seems to work and hopefully is a decent solution for any other models that require media fields in the future.

I haven't renamed the `Translatable` macro yet. The macro is starting to get quite complicated but maybe that's okay if it keeps the rest of the code base simpler? 

I'm curious to know what you think.

stuartlynn (approved):

PR #328, opened 2026-05-13
327/Thinking Space UI

### LINKED ISSUE #327: Build a Prototype in Comhairle
## Summary

Build a working demo within Comhairle to validate the interaction pattern. Some backend capabilities are most likely going to be mocked or simplified for this phase.

- Admin configures a set of main questions and a global follow-up count.
- Participants answer one question at a time, pick from AI-generated follow-up questions, and review extracted claims at the end before submission.

Key problems this addresses:

- "Chatbot feel" creates wrong expectations (users expect real-time dialogue)
- Real-time claim extraction in the sidebar adds cognitive load mid-flow
- Open-ended first question causes blank-page anxiety
- No clear sense of progress or completion

---

### Admin (Configuration)

1. Admin adds/removes/reorders main questions
2. Admin sets a **global follow-up count** (e.g. 2). This applies to all questions equally. Setting 0 means no follow-ups.

### Participant (Public-facing)

1. **Welcome page:** topic intro, expectations (question count, estimated time, privacy note)
2. **Question page:** (one per main question):
   - Main question displayed prominently at top
   - Participant writes their answer and submits
   - Their answer locks in and displays on the page
   - If follow-ups > 0: a set of follow-up question options appears below (4–5 options)
   - Participant picks one (or clicks "Pick one for me")
   - A text area appears for that follow-up; participant answers
   - If more follow-ups required: the picker appears again with remaining options
   - Follow-ups and answers stack on the page (scroll down to current)
   - When all required follow-ups are answered: auto-advance to next question
3. **Review page:** all extracted claims displayed as cards
   - Each claim shows the source question it came from
   - Participant must **approve**, **edit**, or **remove** every claim
   - All claims must be resolved (approved or removed) before the submit button activates
   - Submit → confirmation screen

---

## Acceptance Criteria

- [ ] Admin can add, remove, and reorder main questions
- [ ] Admin can set a global follow-up count (0–5)
- [ ] Participant sees welcome page with topic, question count, and time estimate
- [ ] Participant answers main question → answer locks in and displays
- [ ] If follow-ups > 0: participant sees 4–5 follow-up options to choose from
- [ ] Participant can click "Pick one for me" to auto-select
- [ ] Label shows how many follow-ups remain
- [ ] Answered follow-ups stack on the page with auto-scroll to bottom
- [ ] After all follow-ups answered: auto-advance to next question
- [ ] After all questions: "Review my views" button appears
- [ ] Review page shows all extracted claims with approve/edit/remove controls
- [ ] Submit button only activates when all claims are resolved
- [ ] Confirmation screen shown after submission

---

## Open Questions

1. **Should auto-advance have a brief pause/countdown?** Or is immediate transition (with fade) sufficient? Needs user testing.
2. **Should participants be able to go back to previous questions?** Current design is forward-only. Adding back-navigation increases complexity.
3. **How many follow-up options should we show?** Wireframes show 4–5. Fewer reduces choice overload; more gives variety.
4. **RAGFlow response format...**

dtodorowa:
description in progress

https://github.com/user-attachments/assets/86767c8f-318c-49ce-9b9e-a41cd914a4bd

https://github.com/user-attachments/assets/a71da344-a427-43e7-a2e3-77191c002018


api/src/tools/thinking_space.rs
```diff
@@ -175,6 +202,80 @@ impl From<ThinkingSpaceAnswer> for ThinkingSpaceAnswerDto {
...
+    )
+    .await?;
+
+    let params = AgentConversationRequest {
+        question: payload.history.clone(),
+        topic: Some(tool_config.topic.clone()),
+        history: Some(payload.history),
+        starting_question: Some(payload.starting_question),
+        question_intent: Some(payload.question_intent),
+    };
```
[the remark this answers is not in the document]
rosswilson2306: @dtodorowa Sort of related to the copilot: Stuart and I briefly chatted about whether we needed both the `history` variable if we use the `question` param. I wasn't sure and just wanted to get something working the way we chatted about it (ie, using the `history` variable).

Are you using the question param instead on the frontend? Maybe this is more to do with how the agent is configured, but it might be better to pass the history in to `question` as this gets tracked by ragflow in the `session` history. So we'd have access to the history afterwards if we ever needed it.

This is probably something to discuss separately once this work is merged though, just wanted to bring it up.

ui/packages/comhairle/src/lib/tools/thinking_space/converse.ts
```diff
@@ -0,0 +1,137 @@
...
+ * far, it emits a JSON object `{ questions: [{ type, question }] }`.
+ *
+ * The generated api-client cannot consume streaming responses, so this uses a
+ * raw fetch and reads the stream incrementally.
+ */
+
+export interface FollowUpQuestion {
+	type: string;
+	question: string;
+}
```
rosswilson2306, line 14: @dtodorowa If the streamed response is no good for how this tool is structured then we could just disable it so that we get a normal json response. Again, maybe lets not worry about that now, we can get this merged and come back to it, but there is an option to disable the streaming in ragflow. I think I set it to always stream as I thought we would be using these mainly for chat like tools where streaming is better UX.

ui/packages/comhairle/src/lib/tools/thinking_space/Overview.svelte
```diff
@@ -0,0 +1,146 @@
...
+		draft = '';
+	}
+
+	async function saveEdit(id: string) {
+		const value = draft.trim();
+		if (!value || saving) return;
+		saving = true;
+		try {
+			await apiClient.UpdateThinkingSpaceAnswer(
+				{ answer: value },
```
rosswilson2306, line 47: @dtodorowa Is there somewhere else that handles setting the answer `status` to either `approved` or `declined`?

ui/packages/comhairle/src/lib/tools/thinking_space/QuestionFlow.svelte
```diff
@@ -0,0 +1,511 @@
...
+		initialAnswers = [],
+		onComplete
+	}: Props = $props();
+
+	type Phase = 'main' | 'picking' | 'answering';
+
+	type LocalQuestionState = {
+		mainAnswer: string;
+		mainSubmitted: boolean;
+		mainAnswerId: string | null;
```
rosswilson2306: @dtodorowa If I'm thinking about this correct, main is the admin configured question? If so, can we call this `root` instead of `main` so that it matches the language on the backend?

ui/packages/comhairle/src/lib/tools/thinking_space/QuestionFlow.svelte
```diff
@@ -0,0 +1,511 @@
...
+	let currentState = $derived(states[currentQIdx]);
+	let currentQuestion = $derived(questions[currentQIdx]);
+	let followUpsDone = $derived(currentState.followUps.length);
+	let followUpsRemaining = $derived(Math.max(0, followUpCount - followUpsDone));
+	let isLastQuestion = $derived(currentQIdx === questions.length - 1);
+	// Minimum follow-ups reached for the current question — Continue button
+	// is revealed but the user may keep answering more follow-ups.
+	let minReached = $derived(currentState.mainSubmitted && followUpsDone >= followUpCount);
+
+	let totalSteps = $derived(questions.length * (1 + followUpCount));
```
rosswilson2306: This isn't really an issue for this PR, but this is a lot of state. I feel like we have a few of these components that have complex functionality and a ton of state variables. 

I don't know enough about svelte, but if this was react I'd probably be reaching for `useReducer` / `useContext` or redux when it gets to this stage. Maybe at some point, we should look into design patterns and how best to handle complex state requirements in svelte for things like this

ui/packages/comhairle/src/lib/tools/thinking_space/QuestionFlow.svelte
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	async function loadPicker(qIdx: number) {
+		states[qIdx] = { ...states[qIdx], pickerLoading: true, pickerError: false };
+		try {
+			const followUps = await fetchFollowUps({
+				workflowStepId,
+				startingQuestion: questions[qIdx].text,
+				// No dedicated intent field yet — the question text is the
+				// best proxy until the config schema gains one.
```
rosswilson2306: @dtodorowa Do we not have `question_intent` on the backend? Is this missing somewhere?

ui/packages/comhairle/src/lib/tools/thinking_space/QuestionFlow.svelte
```diff
@@ -0,0 +1,511 @@
...
+		try {
+			const saved = await apiClient.CreateThinkingSpaceAnswer({
+				workflow_step_id: workflowStepId,
+				question: currentState.currentPick,
+				answer: value,
+				is_follow_up: true,
+				root_question_id: currentState.mainAnswerId,
+				other_questions: currentState.picker
+			});
+			const fu: FollowUpAnswer = {
```
rosswilson2306: I know this one is kind of obvious, but can we try and avoid using abbreviations for variable names? It's just so much easier to read at a glance when things aren't abbreviated. I feel like LLMs generate code like this all the time but it's much easier for people reasoning about if things have names instead of abbreviations. Unless it's stupidly long!

rosswilson2306: Maybe I'm just being picky. It's a bit of a pet peeve of mine I guess.

rosswilson2306 (commented):
Left a couple questions and a few comments. Some of the comments aren't PR-blockers though.

ui/packages/comhairle/src/lib/tools/thinking_space/QuestionFlow.svelte
dtodorowa: nahh this is a good point! I will add it to my prompt when i mess with LLMs
also... "fu" lhahahah.. awful

ui/packages/comhairle/src/lib/tools/thinking_space/converse.ts
dtodorowa: yeh, definitely worth trying I think

ui/packages/comhairle/src/lib/tools/thinking_space/Overview.svelte
dtodorowa: this is something I think we need to align on, because I'm not 100% sure we even need approve/declined. 

My thinking is: those are no longer AI generated claims, just answers and they can be edited. Maybe no point of approving/declining?

Backend has a status field that we might not need? Idk, should probably confirm with Stuart.
When I was building this, I based it on an assumption based on a memory of a conversation we had on a call. it was in my notebook.

Could also be a misunderstanding on my end, just gotta get more clarity on this

api/src/tools/thinking_space.rs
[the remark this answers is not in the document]
dtodorowa: From the frontend side I'm only sending history + starting_question + question_intent, no separate question param.

Agreed this is worth a follow-up sync, if ragflow's session history is the place we want the transcript stored, passing history into question (and dropping the separate history param) sounds cleaner. Happy to chat about this anytime

rosswilson2306 (approved):

PR #345, opened 2026-05-18
UI prioritisation tool v1

dtodorowa:
The prioritization tool store `prioritization/store.svelte.ts` currently splits its state between the backend and `localStorage`. Proposals and submitted responses are fully server-owned, but several other pieces of poll state live only in the browser because the backend has no place to put them yet.

I tried to document those gaps below so we can delete the `localStorage` fallback for everything except in-progress participant drafts.

## What's server-backed today

- **Proposals** `POST/GET/PUT/DELETE /tools/prioritization/proposals[...]`
- **Responses** `POST/GET /tools/prioritization/proposals/{id}/responses`

## What's in `localStorage`

All keyed by workflow step id under `comhairle:prioritization:<stepId>`:

- **Poll title and description** the workflow step has no field for either; the UI lets facilitators set them but they only persist in the editor's browser.
- **Tool config** (`randomizeOrder` + `questions[]`)  `workflow_step.tool_config` exists in the DB, but there is no update endpoint, so the questions a facilitator designs never reach the server. Participants on a different device would see an empty question list.
- **Report pages** (`report.pages[]` + `publishedAt`)  no backend representation at all. Published reports are not actually published; they're just marked locally.

## What we need in the backend?

1. **Tool config update endpoint**
   - `PUT /workflow_steps/{id}/tool_config` (or extend an existing workflow-step update path) accepting the prioritization `ToolConfig` shape (`randomize_order: bool`, `questions: Question[]`).
   - Persist into the existing `workflow_step.tool_config` JSONB column.
   - Server should be the source of truth for question ids so re-edits don't collide.

2. **Poll title/description**
   - Decide whether title/description belong on the workflow step itself (likely yes, they describe the step) or inside `tool_config`.
   - If on the step: extend the workflow-step update endpoint to accept `title` and `description`.
   - If inside `tool_config`: fold them into the schema from (1).

3. **Report storage + publish**
   - New table or JSON column for `report.pages[]` per workflow step.
   - Endpoints: `GET /tools/prioritization/{stepId}/report`, `PUT /tools/prioritization/{stepId}/report`, `POST /tools/prioritization/{stepId}/report/publish` (sets `published_at`).
   - Decide whether unpublished reports are visible to non-facilitators (probably no).

4. **Text-answer responses (separate but related)**
   - `submitDraft` currently drops text-question answers because the `Response` model only carries a numeric `value`. If we want text questions to round-trip, extend the response schema.

## Frontend cleanup once backend lands

Delete from `store.svelte.ts`:

- `migratePoll`, `migrateReport`, `pollKey`, `loadJSON`, `saveJSON` (for poll), all `this.persist()` calls.
- Replace `setTitle` / `setDescription` / `setRandomizeOrder` / `addQuestion` / `updateQuestion` / `addReportPage` / `publishReport` etc. with debounced server writes (mirror the `updateProposal` pattern already in the file).
- Keep `loadDraft` / `saveDraft` / `setDraftAnswer` / participant id helpers as-is.

Net effect: ~80 lines of migration/serialization plumbing removed and a single source of truth for everything except drafts.

## Out of scope

- Proposal ordering persistence? 

image
image
image
image
image
image
image
image


dtodorowa:
Closing because it got messy. Creating a new PR soon

PR #360, opened 2026-05-21
Prioritization tool UI - connect backend

dtodorowa:
Add prioritization tool manage and live ui

### Deferred

- Text-type question responses?? This one I am not so sure about, can I actually store a text-type response and how
- Report/shareback pages (no backend table yet)
- Reordering proposals (no `order` column in data model)

image
image
image
image
image


api/src/tools/prioritization.rs
```diff
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use aide::axum::{
-    routing::{get_with, post_with},
+    routing::{delete_with, get_with, post_with},
```
dtodorowa, line 4: think this is leftover from my dev only reset. will rmove

api/src/models/proposal.rs
```diff
@@ -103,6 +107,43 @@ pub async fn list(
...
+pub async fn list_with_translations(
+    db: &PgPool,
+    workflow_step_id: &Uuid,
+    locale: &str,
+) -> Result<Vec<ProposalWithTranslations>, ComhairleError> {
+    let proposals = list_raw(db, workflow_step_id).await?;
+    let mut out = Vec::with_capacity(proposals.len());
+    for proposal in proposals {
+        out.push(ProposalWithTranslations::from_original(db, proposal, locale).await?);
+    }
```
rosswilson2306, line 137: @dtodorowa To be honest, I think this is fine. We're already actually doing this for the workflow steps https://github.com/crownshy/comhairle/blob/be82c9b8f7283f91ba866ec2bcf93fc6c79a629b/api/src/models/workflow_step.rs?plain=1#L454-L458.

The issue is that the `Translatable` macro where we get the `from_original` was built with `get_by_id` functionality in mind. As in, get me a single conversation and then we can fetch the translations for that single instance. Rather than list a bunch of elements with the translations attached for each. 

We have a few use cases for this so maybe we should have a macro method that takes the original list query and attaches the translations for each objects in the results. That will be a fair bit of work though, the macro stuff is pretty abstract and fries my brain any time I have to make a change to it. I'll make a ticket to look into that. For now I think this is fine and consistent with what we have elsewhere.

The only thing I would say is, that instead of having the methods:

- `list` returns `Vec<LocalizedProposal>`
- `list_raw` returns `Vec<Proposal>`

I would change it to:

- `list` returns `Vec<Proposal>`
- `list_localized` returns `Vec<LocalizedProposal>`

so that it's consistent with our other models.

This is my fault. I didn't think through the translations properly and thought that we wouldn't have any cases where we would want to list out the non-localised proposals. I forgot about the fact we'd need to return them with the translations and we have to do that off `Proposal` and not `LocalizedProposal`. Sorry, my bad!

api/src/tools/prioritization.rs
```diff
@@ -50,8 +50,32 @@ pub struct Question {
...
+        #[serde(default = "default_max_value")]
+        max_value: f64,
+        #[serde(default)]
+        min_label: String,
+        #[serde(default)]
+        max_label: String,
+        // NOTE: I added min_label and max_label because I think we need starting
+        // and ending labels for the continuous scale UI. If this is wrong or I
+        // misunderstood the requirements, let me know.
+    },
```
rosswilson2306, line 67: @dtodorowa I meant to bring this up with you, when I was building this part out the config we'd scoped out didn't cover what would be required for this variant, so this seems sensible to me.

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -0,0 +1,277 @@
...
+	function shuffle<T>(arr: T[]): T[] {
+		const out = [...arr];
+		for (let i = out.length - 1; i > 0; i--) {
+			const j = Math.floor(Math.random() * (i + 1));
+			[out[i], out[j]] = [out[j], out[i]];
+		}
+		return out;
+	}
+
+	async function bootstrap() {
```
rosswilson2306: @dtodorowa What does this function actually do? I'm not sure `bootstrap` is that clear a name.

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -0,0 +1,277 @@
...
+					submitted.add(proposal.id);
+					restoredAnswers[proposal.id] = Object.fromEntries(
+						mine.responses.map((r) => [r.questionId, r.value])
+					);
+				}
+			});
+			submittedIds = submitted;
+
+			/** Layer drafts from localStorage on top, but never overwrite already-submitted answers. */
+			const draft = api.loadDraft(stepId, participantId);
```
rosswilson2306: @dtodorowa Is localStorage required here because there is missing functionality on the backend?

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -0,0 +1,277 @@
...
+					(q) => typeof currentAnswers[q.id] === 'number' && currentAnswers[q.id] !== null
+				)
+			: false
+	);
+
+	let allDone = $derived(proposals.length > 0 && proposals.every((p) => submittedIds.has(p.id)));
+
+	/** Persist drafts (debounced via microtask batching — the value only settles after Svelte's reactive update, so any quick succession of changes collapses into one write). */
+	let saveTimer: ReturnType<typeof setTimeout> | undefined;
+	function saveDraftSoon() {
```
rosswilson2306: @dtodorowa I might be misunderstanding what is happening here, but isn't it simpler to just pass the `answers`, `participantId` and `stepId` as arguments to `saveDraft` rather than reading them from the component scope. It has the data passed in to it reliably instead of having to wait for svelte updates to occur?

I might be missing something, but this seems like a code smell. If you're having to do a work around like this to wait for svelte's updates to occur before submitting a request then it seems like something isn't right.

ui/packages/comhairle/src/lib/tools/prioritization/types.ts
```diff
@@ -0,0 +1,108 @@
...
+export type Locale = string;
+
+export type TextFormat = 'plain' | 'rich' | 'markdown';
+
+export type TextContent = {
+	id: string;
+	primaryLocale: Locale;
+	format: TextFormat;
+};
+
```
rosswilson2306, line 21: @dtodorowa Are a lot of these types not just the same as the ones generated by the api client?

rosswilson2306 (commented):
The translations stuff looks good. I left a comment about method naming (which is actually my fault) and some other small questions.

api/src/models/proposal.rs
rosswilson2306: @dtodorowa I would still do this part:
> The only thing I would say is, that instead of having the methods:
> 
> list returns Vec<LocalizedProposal>
> list_raw returns Vec<Proposal>
> I would change it to:
> 
> list returns Vec<Proposal>
> list_localized returns Vec<LocalizedProposal>
> so that it's consistent with our other models.

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
dtodorowa: removed all the draft stuff because this was experimental when we relied on local storage. not needed anymore, so will resolve this.

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
dtodorowa: good point. renamed to loadProposalsAndProgress

ui/packages/comhairle/src/lib/tools/prioritization/types.ts
dtodorowa: yehh .. seems to be leftover stuff i forgot about :/ sorry 
on it

api/src/models/proposal_response.rs
```diff
@@ -68,7 +68,28 @@ impl<'r> Decode<'r, Postgres> for QuestionResponses {
 #[derive(PartialEq, Deserialize, Serialize, JsonSchema, Debug, Clone)]
 pub struct Response {
     pub question_id: Uuid,
-    pub value: f64,
+    pub value: ResponseValue,
+}
+
```
dtodorowa, line 73: confirm if we are cool with this

ui/packages/comhairle/src/lib/tools/prioritization/components/QuestionEditorDialog.svelte
```diff
@@ -0,0 +1,329 @@
...
+		{ label: 'Strongly disagree', value: -2 },
+		{ label: 'Disagree', value: -1 },
+		{ label: 'Neutral', value: 0 },
+		{ label: 'Agree', value: 1 },
+		{ label: 'Strongly agree', value: 2 }
+	];
+
+	function newQuestion(): Question {
+		return {
+			id: crypto.randomUUID(),
```
rosswilson2306: @dtodorowa I don't think you need this do you? The ids for the questions are generated on the backend? Or am I thinking of something else?

ui/packages/comhairle/src/lib/tools/prioritization/prioritizationApi.ts
```diff
@@ -0,0 +1,237 @@
...
+}): Promise<void> {
+	const payload = {
+		type: 'prioritization' as const,
+		questions: opts.toolConfig.questions.map(denormaliseQuestion),
+		randomize_order: opts.toolConfig.randomizeOrder
+	};
+	await apiClient.UpdateConversationWorkflowStep(
+		(opts.isLive
+			? { tool_config: payload }
+			: { preview_tool_config: payload }) as unknown as Parameters<
```
rosswilson2306: Christ, typescript is horrible sometimes.

ui/packages/comhairle/src/lib/tools/prioritization/prioritizationApi.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/* ---------- Responses ---------- */
+
+export async function submitResponse(
+	proposalId: string,
+	responses: QuestionResponse[]
+): Promise<void> {
+	/** Backend `value` is an untagged enum accepting number | string, but the
+	 * generated zod schema still types it as number. Cast until the api-client
+	 * is regenerated. */
```
rosswilson2306: @dtodorowa 
> Cast until the api-client is regenerated

Why hasn't this happened automatically? Are you just `just api-watch` to regenerate the api client when the api changes? (You might need to re-start the svelte dev server for the changes to the api client to be picked up).

ui/packages/comhairle/src/lib/tools/prioritization/prioritizationApi.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+export async function listResponses(proposalId: string): Promise<ProposalResponse[]> {
+	const dtos = await apiClient.ListProposalResponses({ params: { proposal_id: proposalId } });
+	return dtos.map((d) => ({
+		id: d.id,
+		proposalId: d.proposalId,
+		userId: d.userId,
+		/** Backend `value` is number | string at runtime; the generated zod schema
+		 * narrows to number. Widen here until the api-client is regenerated. */
```
rosswilson2306: Again, this should happen automatically.

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationManage.svelte
```diff
@@ -0,0 +1,423 @@
...
+
+	let editorOpen = $state(false);
+	let editingProposal = $state<Proposal | null>(null);
+	let deletingProposal = $state<Proposal | null>(null);
+	let deleting = $state(false);
+
+	let questionEditorOpen = $state(false);
+	let editingQuestion = $state<Question | null>(null);
+	let deletingQuestion = $state<Question | null>(null);
+	let deletingQuestionInFlight = $state(false);
```
rosswilson2306, line 60: @dtodorowa Is there a reason we can't just have `selectedProposal` | `selectedQuestion` instead of having separate states for different actions?

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -0,0 +1,292 @@
...
+	} = $props();
+
+	const stepId = $derived(workflowStep.id);
+	const toolConfig = $derived(api.resolveToolConfig(workflowStep, conversation.isLive ?? false));
+
+	let proposals = $state<LocalizedProposal[]>([]);
+	let answers = $state<Record<string, Record<string, number | string>>>({}); // proposalId → questionId → value
+	let submittedIds = $state<Set<string>>(new Set());
+	let loadState = $state<'loading' | 'ready' | 'error'>('loading');
+	let loadError = $state<string | null>(null);
```
rosswilson2306: This is when I wish javascript had enums like Rust, so that the state could be something like `Error(message)`. Would be so much nicer!

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -0,0 +1,292 @@
...
+			currentIndex += 1;
+		}
+	}
+
+	function goBack() {
+		if (currentIndex > 0) currentIndex -= 1;
+	}
+
+	function finish() {
+		onDone?.();
```
rosswilson2306: @dtodorowa So if `onDone` is undefined then this function does nothing? Should `onDone` be required so we always do something?

rosswilson2306 (commented):
I left some questions. Also, you have some components under `tools/prioritization` and then others under `tools/prioritization/components`. Is this intentional? It feels like all the components should organised together but are they split because the serve different purposes or something?

ui/packages/comhairle/src/lib/tools/prioritization/prioritizationApi.ts
dtodorowa: yup.. that was the case :/

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationManage.svelte
dtodorowa: good point. think i resolved this, but i mightve misunderstood

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
dtodorowa: I looked into this. Apparently TS discriminated unions can get us close to that. I'll fold loadError into the error variant: { kind: 'error', message }

the research:

```ts
type LoadState =
    | { kind: 'loading' }
    | { kind: 'ready' }
    | { kind: 'error'; message: string };

let loadState = $state<LoadState>({ kind: 'loading' });
```

```ts
if (loadState.kind === 'error') {
    // loadState.message is typed as string here
} 
```

```ts
loadState = { kind: 'loading' };
loadState = { kind: 'ready' };
loadState = { kind: 'error', message: 'Failed to load proposals' };
```

stuartlynn (approved):

PR #393, opened 2026-05-26
388 ux improvements

### LINKED ISSUE #388: UX improvements
The flow overall is good! 

There are only a few things if I am being really picky:
https://www.loom.com/share/b4e033ed0ce745a689ca107f0f955a17

1. when privacy policy modal displays, there is no way for user to cancel. suggest to let use click outside of the modal to cancel display. 
2. the continue button is always on the left side of the page, it looks a bit odd. Suggest to bring the button back to center? 
3. (this is more of a question) in the thank you page, where does their feedback go if they actually click on "Give feedback"? 
4. (the following are less relevant to conversation itself but the whole site experienece when user explore the site a bit more) when going to setting, in the notification setting part, it still shows the conversation id instead of conversation name - suggest to display conversation name for better explainability 
In the footer, i think crownshy should be one word "CrownShy"? instead of "Crown Shy"..?

dtodorowa:
Small UX fixes from feedback round

Here were the requests: https://github.com/crownshy/comhairle/issues/388

- **Privacy policy modal**: was trapping users. only the Accept button closed it. Removed the `escapeKeydownBehavior="ignore"` / `interactOutsideBehavior="ignore"` overrides so ESC and outside-click now dismiss. No consent fires on dismiss; user just returns to the conversation landing page.
- **Learn tool Continue / Next buttons**: move back to middle
- **Footer copyright**: "Crown Shy" → "CrownShy" 
- **Settings → Notifications**: was showing the conversation UUID instead of the name. Settings loader now resolves each title via `GetConversation` in parallel and the page renders the real name.

> ⚠️ Note on the settings fix: this is dedinitely not the best approach. Wanted to have something now, but I think we need to do this properly?

### DIFF
--- a/ui/packages/comhairle/src/routes/(public)/settings/+page.ts
+++ b/ui/packages/comhairle/src/routes/(public)/settings/+page.ts
@@ -11,7 +11,24 @@ export const load: PageLoad = async ({ parent }) => {
 		const participation = await api.GetConversationsUserIsParticipatingIn();
 		const conversation_settings = await api.GetAllUserConversationPreferences();
 
-		return { participation, conversation_settings, user };
+		const localizedParticipation = await Promise.all(
+			participation.map(async (c) => {
+				try {
+					const localized = await api.GetConversation({
+						params: { conversation_id: c.id }
+					});
+					return { id: c.id, title: localized.title };
+				} catch {
+					return { id: c.id, title: c.id };
+				}
+			})
+		);
+
+		return {
+			participation: localizedParticipation,
+			conversation_settings,
+			user
+		};
 	} catch (e) {
 		return { error: e };
 	}
--- a/ui/packages/comhairle/src/routes/(public)/settings/+page.ts
+++ b/ui/packages/comhairle/src/routes/(public)/settings/+page.ts
@@ -11,24 +11,7 @@ export const load: PageLoad = async ({ parent }) => {
 		const participation = await api.GetConversationsUserIsParticipatingIn();
 		const conversation_settings = await api.GetAllUserConversationPreferences();
 
-		const localizedParticipation = await Promise.all(
-			participation.map(async (c) => {
-				try {
-					const localized = await api.GetConversation({
-						params: { conversation_id: c.id }
-					});
-					return { id: c.id, title: localized.title };
-				} catch {
-					return { id: c.id, title: c.id };
-				}
-			})
-		);
-
-		return {
-			participation: localizedParticipation,
-			conversation_settings,
-			user
-		};
+		return { participation, conversation_settings, user };
 	} catch (e) {
 		return { error: e };
 	}
--- a/api/src/models/conversation.rs
+++ b/api/src/models/conversation.rs
@@ -547,8 +547,9 @@ pub async fn update(
 pub async fn list_for_user_participation(
     db: &PgPool,
     user_id: &Uuid,
-) -> Result<Vec<Conversation>, ComhairleError> {
-    let (sql, values) = Query::select()
+    locale: &str,
+) -> Result<Vec<LocalizedConversation>, ComhairleError> {
+    let query = Query::select()
         .from(ConversationIden::Table)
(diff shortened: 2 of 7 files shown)


ui/packages/comhairle/src/routes/(public)/settings/+page.ts
```diff
@@ -11,7 +11,24 @@ export const load: PageLoad = async ({ parent }) => {
 		const participation = await api.GetConversationsUserIsParticipatingIn();
 		const conversation_settings = await api.GetAllUserConversationPreferences();
 
-		return { participation, conversation_settings, user };
+		const localizedParticipation = await Promise.all(
+			participation.map(async (c) => {
+				try {
+					const localized = await api.GetConversation({
```
rosswilson2306: @dtodorowa I think it makes more sense to do this on the backend. We can change `GetConversationsUserIsParticipatingIn()` to return `Vec<LocalizedConverationDto>` instead of `Vec<ConversationDto>` and that way we don't make one request to get the list and then potentially dozens of additional `GetConversation` requests just to get the localized versions.

I don't see any situation where you'd want `GetConversationsUserIsParticipatingIn` to return the raw conversations so I think it would make sense to make this change.

I think all you'd need to do is change this model method https://github.com/crownshy/comhairle/blob/be82c9b8f7283f91ba866ec2bcf93fc6c79a629b/api/src/models/conversation.rs?plain=1#L546-L587 to something like:

```rust
#[instrument(err(Debug))]
pub async fn list_for_user_participation(
    db: &PgPool,
    user_id: &Uuid,
    locale: &str,
) -> Result<Vec<LocalizedConversation>, ComhairleError> {
    let query = Query::select()
        .from(ConversationIden::Table)
        .columns(DEFAULT_COLUMNS.map(|col| (ConversationIden::Table, col)))
        .join(
            sea_query::JoinType::InnerJoin,
            WorkflowIden::Table,
            Expr::col((WorkflowIden::Table, WorkflowIden::ConversationId))
                .equals((ConversationIden::Table, ConversationIden::Id)),
        )
        .join(
            sea_query::JoinType::InnerJoin,
            UserParticipationIden::Table,
            Expr::col((
                UserParticipationIden::Table,
                UserParticipationIden::WorkflowId,
            ))
            .equals((WorkflowIden::Table, WorkflowIden::Id)),
        )
        .and_where(
            Expr::col((UserParticipationIden::Table, UserParticipationIden::UserId))
                .eq(user_id.to_owned()),
        )
        .to_owned();

    let (sql, values) = LocalizedConversation::query_to_localisation(query, locale).build_sqlx(PostgresQueryBuilder);

    let conversations = sqlx::query_as_with::<_, Conversation, _>(&sql, values)
        .fetch_all(db)
        .await?;

    Ok(conversations)
}
```

Change this route https://github.com/crownshy/comhairle/blob/be82c9b8f7283f91ba866ec2bcf93fc6c79a629b/api/src/routes/user.rs?plain=1#L73-L83 to:
```rust
pub async fn get_conversations_user_participating_in(
    RequiredUser(user): RequiredUser,
    LocaleExtractor(locale): LocaleExtractor,
) -> Result<(StatusCode, Json<Vec<LocalizedConversationDto>>), ComhairleError> {
    let conversations = models::conversation::list_for_user_participation(&state.db, &user.id, &locale)
        .await?
        .into_iter()
        .map(Into::into)
        .collect();

    Ok((StatusCode::OK, Json(conversations)))
}
```

Then change the response type on the open-api doc generator https://github.com/crownshy/comhairle/blob/be82c9b8f7283f91ba866ec2bcf93fc6c79a629b/api/src/routes/user.rs?plain=1#L143 to `Vec<LocalizedConversationDto>`. 

It doesn't look like that route has any tests so I don't think it would break any tests and it looks like the only place that route is called is from the page you have edited, so should be pretty safe to change.

rosswilson2306 (commented):
Left a comment about something I think should be changed on the backend. Let me know if you want to pair up on it.

rosswilson2306 (approved):

PR #404, opened 2026-05-27
Add quickstart docs and make it easier to onboard

dtodorowa:
I think this is the simplest way of onboarding 

```bash
cp .env.example .env   # set up env vars
nix develop            # get all deps
just all               # db, backend and frontend (in the nix env)
just seed              # add admin user (and maybe some conversations)
```

This should get everything running super quickly, I think. Example (i'd already ran `nix develop` so everything was cached, otherwise it's like 2gb download wait):

> I switch between tabs using ctrl-b + tab index (0 pg, 1 api, 2 ui, 3 free terminal)

https://github.com/user-attachments/assets/4d1c8e5d-5d3d-463b-b46c-7608729adf3a

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,9 @@ ADMIN_USERS=admin@crown-shy.com
 # TRANSCRIPTION_SERVICE__AMAZON_TRANSCRIBE_API_KEY=
 
 # ---- Worker service (background jobs via Redis) ----
-# WORKER_SERVICE__REDIS_URL=redis://localhost:6379
+# Started by `just start-redis` (and `just all`) on port 63793.
+# Without this set, the API hangs ~10s on startup waiting for Redis before
+# giving up and running without the worker service.
+WORKER_SERVICE__REDIS_URL=redis://127.0.0.1:63793
 
 # TODO: Add the rest
--- a/.env.example
+++ b/.env.example
@@ -11,7 +11,7 @@
 DATABASE_URL=postgres://comhairle:comhairle@localhost:5434/comhairle
 RUST_LOG=debug
 RESOURCE_BUCKET=comhairle_resources
-ADMIN_USERS=admin@example.com
+ADMIN_USERS=admin@crown-shy.com
 # ENABLE_RATE_LIMITING=false
 
 # ---- Mailer (SMTP) ----
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,47 @@
+# Comhairle local environment.
+#
+# Copy this file to `.env` and fill in values for the services you need.
+# `.env` is gitignored.
+#
+# Naming: the `config` crate maps double-underscores (`__`) to nested
+# fields, so `MAILER__HOST` -> `mailer.host`. Comma-separated lists are
+# parsed for `ADMIN_USERS` and `WHITELISTED_DOMAINS`.
+
+# ---- Core (required for local dev) ----
+DATABASE_URL=postgres://comhairle:comhairle@localhost:5434/comhairle
+RUST_LOG=debug
+RESOURCE_BUCKET=comhairle_resources
+ADMIN_USERS=admin@example.com
+# ENABLE_RATE_LIMITING=false
+
+# ---- Mailer (SMTP) ----
+# MAILER__HOST=
+# MAILER__USER=
+# MAILER__PASSWORD=
+# MAILER__FROM_EMAIL=invites@comhairle.scot
+
+# ---- Translator (Google Translate) ----
+# TRANSLATOR__TYPE=google
+# TRANSLATOR__API_KEY=
+
+# ---- Bot service (RAG) ----
+# BOT_SERVICE_HOST=
+# BOT_SERVICE_API_KEY=
+# DEFAULT_KNOWLEDGE_BASE_ID=
+# ELICITATION_BOT_AGENT_ID=
+
+# ---- Video call service (JWT for Jitsi/8x8 etc.) ----
+# VIDEO_CALL_SERVICE__JWT_APP_ID=
+# VIDEO_CALL_SERVICE__JWT_APP_SECRET=
+# VIDEO_CALL_SERVICE__JWT_SUB=
+# VIDEO_CALL_SERVICE__JWT_ACCEPTED_ISSUERS=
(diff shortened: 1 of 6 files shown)


scripts/seed-minimal.sh
```diff
@@ -0,0 +1,46 @@
...
+    --data "$(jq -nc \
+      --arg u "${ADMIN_USERNAME}-$RANDOM" \
+      --arg e "$ADMIN_EMAIL" \
+      --arg p "$ADMIN_PASSWORD" \
+      '{username:$u,email:$e,password:$p}')" >/dev/null
+  AUTH_COOKIE=$(login || true)
+fi
+
+[ -n "$AUTH_COOKIE" ] || { echo "login failed"; exit 1; }
+echo "✅ admin ready: $ADMIN_EMAIL"
```
dtodorowa, line 46: could also add some conversations here?

.env.example
```diff
@@ -0,0 +1,47 @@
...
+
+# ---- Translator (Google Translate) ----
+# TRANSLATOR__TYPE=google
+# TRANSLATOR__API_KEY=
+
+# ---- Bot service (RAG) ----
+# BOT_SERVICE_HOST=
+# BOT_SERVICE_API_KEY=
+# DEFAULT_KNOWLEDGE_BASE_ID=
+# ELICITATION_BOT_AGENT_ID=
```
rosswilson2306, line 31: @dtodorowa This is going to change pretty soon. I currently have this and the updated version:
```sh
BOT_SERVICE__HOST="https://rag.comhairle.scot"
BOT_SERVICE__API_KEY="..."
BOT_SERVICE__DEFAULT_KNOWLEDGE_BASE_ID="..."
BOT_SERVICE__ELICITATION_BOT_AGENT_ID="..."
BOT_SERVICE__THINKING_SPACE_AGENT_ID="..."
```
in my env. Probably worth adding both to make it smoother when the thinking space work is merged, then we can remove the old ones?

flake.nix
```diff
@@ -1,69 +1,104 @@
...
+            # ---- Native build deps ----
+            openssl
+            pkg-config
+            cmake
+            clang
+            libiconv
+
+            # ---- Database / infra ----
+            postgresql_16   # provides psql client
+            redis           # provides redis-cli; runtime redis can also come from docker
```
rosswilson2306, line 58: @dtodorowa I don't know much about nix and flake files, does this ensure you have dependencies installed? If so, can we add docker to the list of dependencies? 

It's a bit of a given they'll already have docker installed but just in case someone starts on a new laptop like I did, a lot is dependent on having docker and won't work without it. Plus if we're ensuring dependencies it should cover everything.

justfile
```diff
@@ -28,3 +28,15 @@ watch-api-spec:
...
+seed:
+    ./scripts/seed-minimal.sh
+
+# Run DB, API and UI in separate tmux windows
+all:
+    tmux new-session -d -s comhairle -n postgres "just pg" \; \
+        new-window -n api "just api-dev" \; \
+        new-window -n ui "cd ui/packages && pnpm comhairle" \; \
+        new-window -n seed "echo 'wait for api then: just seed'; exec $SHELL" \; \
+        attach-session -t comhairle
```
rosswilson2306, line 51: @dtodorowa Something else that is probably a good idea to add a just command for is the redis instance that the worker service requires. Without it the server hangs for 10 seconds while trying to find a connection, before timing out and starting up without the worker service. If someone isn't familiar with the Rust logs, that might not be clear and it might be confusing why it looks like they have started the server but incoming requests are failing for the first 10 seconds.

This is the command I used to set up redis in a docker container:
```sh
docker run -d \
--name apalis-redis \
-p 63793:6379 \
-v apalis-redis-data:/data \
redis:7 redis-server --save 60 1
```
and then after that it would be `docker start apalis-redis`. So I think you could do something like this:
```just
start-redis:
    docker start apalis-redis 2>/dev/null || docker run -d \
        --name apalis-redis \
        -p 63793:6379 \
        -v apalis-redis-data:/data \
        redis:7 redis-server --save 60 1
```
to try to run the container if it exists, and if it exits with an error code run the fallback to create the container.

scripts/seed-minimal.sh
```diff
@@ -0,0 +1,46 @@
...
+BACKEND_URL="${API_URL:-http://localhost:3000}"
+ADMIN_EMAIL="${ADMIN_EMAIL:-admin@crown-shy.com}"
+ADMIN_PASSWORD="${ADMIN_PASSWORD:-adminPassword123!}"
+ADMIN_USERNAME="${ADMIN_USERNAME:-devadmin}"
+
+command -v curl >/dev/null || { echo "missing curl"; exit 1; }
+command -v jq   >/dev/null || { echo "missing jq";   exit 1; }
+
+login() {
+  curl -s -i -X POST "$BACKEND_URL/auth/login" \
```
rosswilson2306, line 23: @dtodorowa Is this not dependent on the `admin@crown-shy.com` user already existing in the local db? It feels like we should be hitting `/auth/signup` instead if this is to run for the first time?

rosswilson2306 (commented):
Left a couple of comments / questions.

justfile
dtodorowa: ohh yes yes me likeyyy

flake.nix
dtodorowa: If I understand stuff correctly, Docker needs a running daemon with root and kernel-level features, so it has to be installed at the system level. Don't think a flake can spin that up.

I added Docker to the Prereqs section in the quick start doc tho, hope that's clear enough. But a really good point here

scripts/seed-minimal.sh
dtodorowa: yehh this makes complete sense, I think I added this because originally I wanted to always run this on start, and wanted to avoid trying to create the admin account over and over again. 

now seed_minmal is optional, so i think it makes sense to just keep it as signup only

dtodorowa: this was my thinking
```
login as admin
1. no session, signing up
OR
2. login
```

flake.nix
rosswilson2306: Yeah, that's fair enough. Docker is a bit of a faff to install on linux so maybe nix can't just bypass all that.

rosswilson2306 (approved):

PR #408, opened 2026-05-28
407 api key model

rosswilson2306:
Actions:

+ migration to create api_key table
+ api_key model with generate functionality
+ api_key router with single create route which generates key

Motivation:

+ api key auth system

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -121,6 +121,9 @@ urlencoding = "2.1.3"
 css-inline = "0.20.2"
 chrono-tz = "0.10.4"
 icalendar = "0.17.10"
+sha2 = "0.11.0"
+hex = "0.4.3"
+base64 = "0.22.1"
 
 [dev-dependencies]
 fake = { version = "4", features = ["derive"] }
--- a/api/src/models/api_key.rs
+++ b/api/src/models/api_key.rs
@@ -1,0 +1,123 @@
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
+use chrono::{DateTime, Utc};
+use rand::RngCore;
+use schemars::JsonSchema;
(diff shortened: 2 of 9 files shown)


api/src/models/api_key.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub id: Uuid,
+    pub hash: String,
+    pub user_id: Uuid,
+    pub name: String,
+    pub prefix: String,
+    pub revoked_at: Option<DateTime<Utc>>,
+    pub expired_at: Option<DateTime<Utc>>,
+    pub created_at: DateTime<Utc>,
+    pub updated_at: DateTime<Utc>,
+}
```
stuartlynn, line 26: Might be worth creating a method on the key that returns if it's valid or not. Just for convenience. 

Also what are we using the prefix for?

api/src/models/api_key.rs
```diff
@@ -0,0 +1,123 @@
...
+
+fn hash_api_key(raw_key: &str) -> String {
+    let mut hasher = Sha256::new();
```
stuartlynn, line 29: We are using argon I think for the password hashing for user accounts. Any reason to not use it here as well?

api/Cargo.toml
```diff
@@ -121,6 +121,9 @@ urlencoding = "2.1.3"
 css-inline = "0.20.2"
 chrono-tz = "0.10.4"
 icalendar = "0.17.10"
+sha2 = "0.11.0"
+hex = "0.4.3"
+base64 = "0.22.1"
```
stuartlynn, line 126: Just a similar comment as before about argon. Might lets us have a few less dependencies

stuartlynn (commented):
Looks good overall just a few questions

stuartlynn (approved):

PR #412, opened 2026-05-29
407 api key extraction

rosswilson2306:
Actions:

+ method to get user id from api key
+ prevent non-admin users from generating api keys
+ refactor RequiredUser and RequiredAdminUser extractors to check user first from api key bearer token with fallback to session cookie

Motivation:

+ allow developers and external apps to authenticate requests with api keys instead of session cookies

### DIFF
--- a/api/src/error.rs
+++ b/api/src/error.rs
@@ -199,6 +199,9 @@ pub enum ComhairleError {
     #[error("Requires Auth User")]
     RequiresAuthUser,
 
+    #[error("Invalid api key")]
+    InvalidApiKey,
+
     #[error("Only the owner of the conversation can perform this action")]
     UserIsNotConversationOwner,
 
@@ -339,6 +342,7 @@ impl IntoResponse for ComhairleError {
             | ComhairleError::NoUserFoundForId(_) => StatusCode::NOT_FOUND,
             ComhairleError::UserRequired
             | ComhairleError::WrongPassword
+            | ComhairleError::InvalidApiKey
             | ComhairleError::RequiresAuthUser
             | ComhairleError::InviteDoesNotMatchUser
             | ComhairleError::UserIsNotConversationOwner
--- a/api/src/models/api_key.rs
+++ b/api/src/models/api_key.rs
@@ -2,11 +2,11 @@ use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use chrono::{DateTime, Utc};
 use rand::RngCore;
 use schemars::JsonSchema;
-use sea_query::{enum_def, PostgresQueryBuilder, Query};
+use sea_query::{enum_def, Expr, PostgresQueryBuilder, Query};
 use sea_query_binder::SqlxBinder;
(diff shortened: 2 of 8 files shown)


api/src/models/api_key.rs
```diff
@@ -78,10 +81,37 @@ pub async fn create(db: &PgPool, create_key: CreateRequest) -> Result<String, Co
...
+        .and_where(Expr::col(ApiKeyIden::Hash).eq(hash))
+        .and_where(Expr::col(ApiKeyIden::RevokedAt).is_null())
+        .and_where(
+            Expr::col(ApiKeyIden::ExpiredAt)
+                .is_null()
+                .or(Expr::col(ApiKeyIden::ExpiredAt).gt(Expr::current_timestamp())),
+        )
+        .build_sqlx(PostgresQueryBuilder);
+
+    let user_id = query_scalar_with::<_, Uuid, _>(&sql, values)
```
rosswilson2306, line 99: @stuartlynn I wrote this, only extracting the column values I needed, rather than the entire row so that I didn't have the actual api key row / hash in memory so it couldn't be accidentally exposed somehow. But I've started thinking this is maybe overkill. What do you think?

api/src/error.rs
```diff
@@ -339,6 +342,7 @@ impl IntoResponse for ComhairleError {
             | ComhairleError::NoUserFoundForId(_) => StatusCode::NOT_FOUND,
             ComhairleError::UserRequired
             | ComhairleError::WrongPassword
+            | ComhairleError::InvalidApiKey
```
stuartlynn, line 345: I wonder if we should be more specific here. InvalidApiKey  + abother APIKeyExpired etc

api/src/routes/api_keys.rs
stuartlynn, line 1: We probably want a list route to return a list of API keys. Usually this will show the list of active api keys for the user with created_at, expires_at and the API key but with all but the last  4 characters x'ed out.

api/src/models/api_key.rs
```diff
@@ -42,14 +42,17 @@ fn generate_api_key(prefix: &str) -> (String, String) {
...
+    pub name: String,
+    pub prefix: String,
 }
 
 // No tracing to avoid keys being exposed in logs
-pub async fn create(db: &PgPool, create_key: CreateRequest) -> Result<String, ComhairleError> {
+pub async fn create(
+    db: &PgPool,
+    user_id: Uuid,
+    create_key: CreateApiKeyRequest,
```
stuartlynn, line 54: I wonder if it would be good to add a name for each api key. A way to quickly indicate what it's being used for (civicos, some random other thing, etc)

stuartlynn (approved):

api/src/models/api_key.rs
rosswilson2306: @stuartlynn That's one of the columns and a required param on `CreateApiKeyRequest`:
```
|  name   |
+---------+
| dev_key |
```

PR #419, opened 2026-05-29
191/event registration auth

rosswilson2306:
Actions:

- move event confirmation out of invite acceptance and into attendance creation route
- refactor otp creation to optionally take username
- refactor event attendance to optionally take email param and create attendance for associated user with fallback to logged in user

Motivation:

- allow for event registration flows that don't involve invites

### DIFF
--- a/api/src/routes/event_attendances.rs
+++ b/api/src/routes/event_attendances.rs
@@ -66,9 +66,10 @@ pub async fn get(
     Ok((StatusCode::OK, Json(event_attendance)))
 }
 
-#[derive(Serialize, Deserialize, JsonSchema, Debug)]
+#[derive(Serialize, Deserialize, JsonSchema, Debug, Default)]
 pub struct CreateEventAttendanceRequest {
     role: String,
+    user_email: Option<String>,
 }
 
 #[instrument(err(Debug), skip(state))]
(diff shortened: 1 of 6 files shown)


api/src/routes/event_attendances.rs
```diff
@@ -78,6 +80,13 @@ pub async fn create(
     RequiredUser(user): RequiredUser,
     Json(payload): Json<CreateEventAttendanceRequest>,
 ) -> Result<(StatusCode, Json<EventAttendanceDto>), ComhairleError> {
+    // Search for user by email if passed, otherwise create attendence from
+    // logged in user (session cookie)
+    let user = match payload.user_email {
+        Some(email) => users::get_user_by_email(&email, &state.db).await?,
+        None => user,
+    };
```
stuartlynn, line 97: We only want to allow this branch when the user is the owner of the conversation right. Right now this allows any user to signup any other user to the event

stuartlynn: Where the user is either the owner by virtue of being logged in as such or using an API key associated with that user

stuartlynn (changes requested):

api/src/routes/event_attendances.rs
rosswilson2306: Totally forgot about this part. Good catch.

stuartlynn (approved):

PR #426, opened 2026-06-03
211/in person events

rosswilson2306:
Actions:

- migration to add JSONB location column to event table
- model updates for events to allow adding / updating location data
- admin ui for adding updating event location

Motiviation: 

- required for in-person events

### DIFF
--- a/api/src/models/event.rs
+++ b/api/src/models/event.rs
@@ -8,7 +8,7 @@ use sea_query_binder::SqlxBinder;
 use serde::{Deserialize, Serialize};
 use sqlx::{
     prelude::{FromRow, Type},
-    query_as_with, Decode, Encode, PgPool, Postgres,
+    Decode, Encode, PgPool, Postgres,
 };
 use sqlx_postgres::{PgArgumentBuffer, PgValueRef};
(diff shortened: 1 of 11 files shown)


api/src/models/event.rs
```diff
@@ -78,12 +82,53 @@ pub struct Event {
...
     pub created_at: DateTime<Utc>,
     #[partially(omit)]
     pub updated_at: DateTime<Utc>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema, PartialEq)]
+pub struct EventLocation {
+    venue: String,
+    address: String,
+}
```
stuartlynn, line 106: I wonder if we need a bit more structure here for the address. Something like 

```
{
 "venue_name": "Some Venue",
  "line1": "123 Main Street",
  "line2": "Apt 4B",
  "line3": null,
  "city": "Edinburgh",
  "state_province": "Scotland",
  "postal_code": "EH1 1YZ",
  "country_code": "GB"
}

stuartlynn (changes requested):

stuartlynn (approved):

PR #463, opened 2026-06-11
On return take compleated users to the first returnable step

stuartlynn:
Motivation. 

Some users where confused as they return to a conversation once completed and they are taken to a thank you page. Instead if there are any returnable steps we should send them to the first one.


dtodorowa (approved):

georgehulme (approved):
LGTM

PR #465, opened 2026-06-12
456 webhook report deserialization

rosswilson2306:
Actions:

+ extract request body as string for webhook url and use string in webhook signature verification

Motivation:

+ extracting to serde_json::Value changes order of key / value pairs breaking hash verification


stuartlynn (dismissed):
This looks good!

PR #467, opened 2026-06-16
Aux polis data table

stuartlynn:
This PR creates a new table and endpoints to save additional data associated with a polis comment. 

It  enables this on the custom polis front end by posting both to polis and the comhairle api when a new comment is added. 

It  also includes a sync endpoint that will synchronize the data in the a polis database with the new table so we can backport previous conversations (mostly for bloom)

### DIFF
--- a/api/src/models/polis_statement_aux.rs
+++ b/api/src/models/polis_statement_aux.rs
@@ -1,0 +1,386 @@
+use chrono::{DateTime, Utc};
+use schemars::JsonSchema;
+use sea_query::{enum_def, Expr, PostgresQueryBuilder, Query, SelectStatement, SimpleExpr};
+use sea_query_binder::SqlxBinder;
+use serde::{Deserialize, Serialize};
+use sqlx::{prelude::FromRow, query_as_with, PgPool};
+use tracing::instrument;
+use uuid::Uuid;
+
+use crate::error::ComhairleError;
+
+#[derive(Debug, Deserialize, Serialize, PartialEq, sqlx::Type, Clone, JsonSchema)]
+#[sqlx(type_name = "TEXT")]
+#[serde(rename_all = "snake_case")]
+pub enum ModerationStatus {
+    #[sqlx(rename = "accepted")]
+    Accepted,
+    #[sqlx(rename = "rejected")]
+    Rejected,
+    #[sqlx(rename = "pending")]
+    Pending,
+}
+
+impl Default for ModerationStatus {
+    fn default() -> Self {
+        Self::Pending
+    }
+}
+
+impl std::fmt::Display for ModerationStatus {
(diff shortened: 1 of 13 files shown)


api/src/models/polis_statement_aux.rs
```diff
@@ -0,0 +1,449 @@
...
+
+    Ok(aux)
+}
+
+#[instrument(err(Debug))]
+pub async fn list_filtered(
+    db: &PgPool,
+    workflow_step_id: Option<Uuid>,
+    polis_conversation_id: Option<String>,
+) -> Result<Vec<PolisStatementAux>, ComhairleError> {
```
rosswilson2306, line 327: @stuartlynn Are both `list` and `list_filtered` needed? It looks like the `filter_options` for `list` covers this? Is this just a convenience method for the most common use case?

api/src/tools/polis.rs
```diff
@@ -227,6 +306,152 @@ async fn get_report_data(
...
+        .login(&WikiPollLogin {
+            email: config.admin_user.clone(),
+            password: config.admin_password.clone(),
+        })
+        .await?;
+    info!("LOGGED IN");
+
+    let xid_rows = client.get_xids(&config.poll_id, &auth_cookies).await?;
+
+    info!("XID ROWS: {xid_rows:#?}");
```
rosswilson2306: Are these just debug logs or are they useful for logging do you think?

api/src/wiki_poll_service/polis_service.rs
```diff
@@ -549,6 +554,31 @@ impl WikiPollService for PolisClient {
...
+        &self,
+        poll_id: &str,
+        auth_cookies: &str,
+    ) -> Result<Vec<WikiPollXid>, WikiPollServiceError> {
+        let url = format!(
+            "https://{}/api/v3/xids?conversation_id={poll_id}",
+            self.base_url
+        );
+
+        info!("attempting to get xids at {url}");
```
rosswilson2306, line 567: Same question here, there are some debug logs but if these are useful for logging then ignore me.

rosswilson2306 (dismissed):
Left a couple of questions but I've approved in case they can be ignored.

api/src/models/polis_statement_aux.rs
stuartlynn: Just simplified this. We might need to tweak it a bit in future to allow for grabbing statements filtered by user  / created at time etc but I think this is probably good

api/src/tools/polis.rs
stuartlynn: Removed

api/src/wiki_poll_service/polis_service.rs
stuartlynn: removed

PR #473, opened 2026-06-17
bump version of api-client package to match newly published version

rosswilson2306:


ui/packages/api-client/src/api.ts
```diff
@@ -4106,11 +4106,6 @@ Use a raw HTTP request and process the response body incrementally.
         type: "Query",
         schema: is_complete,
       },
-      {
```
alpinegeorge, line 4109: Did you mean to include this?

rosswilson2306: @alpinegeorge Yeah. That file is generated of the current state of the api OpenAPI spec, and is used to generate the api-client. I was publishing the api-client off a fresh pull of staging so there must have been a change to the api that got merged without this file getting regenerated. 

This change should put it properly in sync with the current staging api, which is what we needed for the new api-client version.

Should have made that clearer in the PR.

alpinegeorge: Sounds like another thing we should include in a commit hook!
Ideally our API shouldn't be committed in a stale state.

alpinegeorge (approved):
LGTM

PR #475, opened 2026-06-17
430/configurable emails

rosswilson2306:
Actions:

- new table and data model for email_template_configs, representing custom configurations for comhairle emails
- schemas to act as a source of truth for email template content, including fallback content
- refactor associated mailer methods to use email_template_configs if available and fallback content from schemas
- admin UI for creating, editing and previewing custom emails

Motiviation:

- allow client configuration of emails and remove hardcoded content

TODO:

- find a solution that will work with scheduled emails
- other mailer methods for other email templates

### DIFF
--- a/api/src/models/email_template_config.rs
+++ b/api/src/models/email_template_config.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use chrono::{DateTime, Utc};
 use schemars::JsonSchema;
 use sea_query::{enum_def, Expr, PostgresQueryBuilder, Query, SelectStatement, SimpleExpr};
@@ -12,7 +14,10 @@ use sqlx_postgres::{PgArgumentBuffer, PgTypeInfo, PgValueRef};
 use tracing::instrument;
 use uuid::Uuid;
 
-use crate::{error::ComhairleError, models::users};
+use crate::{
+    error::ComhairleError,
+    models::{users, SqlxResultExt},
+};
 
 /// A client-configured email template, persisted to the `email_template_config` table.
 ///
@@ -52,8 +57,8 @@ pub enum EmailTemplateSlots {
     EventRegistrationConfirmation(DefaultEmailSlots),
 }
 
-const TYPE_CONVERSATION_INVITE: &str = "conversation_invite";
-const TYPE_EVENT_REGISTRATION_CONFIRMATION: &str = "event_registration_confirmation";
+pub const TYPE_CONVERSATION_INVITE: &str = "conversation_invite";
+pub const TYPE_EVENT_REGISTRATION_CONFIRMATION: &str = "event_registration_confirmation";
 
 impl EmailTemplateSlots {
     /// Returns the schema for every email template type.
@@ -95,6 +100,29 @@ impl EmailTemplateSlots {
             EmailTemplateSlots::EventRegistrationConfirmation(_) => DEFAULT_SLOTS_SCHEMA,
         }
     }
+
+    /// Converts the email template slots into a `HashMap` of key/value pairs
+    /// suitable for use as a minijinja template context.
+    ///
+    /// The returned map contains the slot field names (e.g. `heading`, `intro`,
+    /// `body`, `footer`) and their corresponding values, which are used to
+    /// populate the variables referenced in the email's minijinja template.
+    ///
+    /// Because `minijinja::context!` supports spreading a `HashMap` with `..`,
+    /// the map returned here can be combined with additional ad-hoc context
+    /// variables that aren't stored as part of the email config (e.g. dynamic
+    /// links or IDs generated at send time). For example:
+    ///
+    /// ```ignore
+    /// let base = email_config.slots.to_mailer_map();
+    /// let context = minijinja::context! { invite_link => "foo@bar.com", ..base };
+    /// ```
+    pub fn to_mailer_map(&self) -> HashMap<&str, String> {
+        match self {
+            EmailTemplateSlots::ConversationInvite(slots) => slots.to_mailer_map(),
+            EmailTemplateSlots::EventRegistrationConfirmation(slots) => slots.to_mailer_map(),
+        }
+    }
 }
 
 impl std::fmt::Display for EmailTemplateSlots {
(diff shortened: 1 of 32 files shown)


api/src/models/email_template_config.rs
```diff
@@ -0,0 +1,814 @@
...
+    ///
+    /// Each entry describes a variant of [`EmailTemplateSlots`], pairing the
+    /// variant's string identifier with the [`SlotDefinition`]s that define its
+    /// configurable slots. This is intended to be served to the frontend so it
+    /// can render the correct form fields when a user selects an email type to
+    /// configure.
+    ///
+    /// # Adding a new email type
+    ///
+    /// When a new variant is added to [`EmailTemplateSlots`], a corresponding
```
georgehulme: There seems to be a fair amount of repeated code due to the email template slots sharing a single enum type.
Could we not enforce this by introducing a new trait? e.g.
```rust
trait EmailTemplateSlot: EmailTemplateSchema + Serialize + Deserialize + JsonSchema {
  fn schema(&self) -> EmailTypeSchema;
  fn email_template() ->&'static str;
  fn template_variables() -> &'static [&'static str]
  fn type_str() -> &'static str
  fn mailer_slots_map(&self) -> HashMap<String, String>
  fn preview_variables_map() -> HashMap<String, String>
  fn schema_slots() -> &'static [SlotSchemaDefinition]
}
```
and update the database row to: `pub slots: Slot` where `Slot: EmailTemplateSlot`.
Then each slot type corresponds 1:1 with an enforced rust type and we could easily extend it with `fn add_schema_slot(&mut self, SlotSchemaDefinition)` and `fn render(&self) -> Html` methods if we wanted to add user-made templates in future.

api/src/models/email_template_config.rs
```diff
@@ -0,0 +1,814 @@
...
+                    "event_name".to_string(),
+                    "Prioritising accessibility in websites".to_string(),
+                ),
+                ("event_time".to_string(), "24 May, 2026".to_string()),
+                ("organization_name".to_string(), "CrownShy".to_string()),
+            ]),
+        }
+    }
+
+    pub fn fallback_slots_map(&self) -> HashMap<String, String> {
```
rosswilson2306: Should add some documentation comments here.

api/src/routes/email_template_configs.rs
```diff
@@ -0,0 +1,230 @@
...
+                op.id("PreviewEmailTemplateConfig")
+                    .summary("Preview email template config")
+                    .description("Preview appearance of custom email before sending")
+                    .security_requirement("JWT")
+                    .tag("EmailTemplateConfig")
+                    .response::<200, Json<PreviewEmailTemplateConfigResponse>>()
+            }),
+        )
+        .with_state(state)
+}
```
rosswilson2306, line 237: Should add some tests for the router

api/src/models/email_template_config.rs
rosswilson2306: @georgehulme With the "repeated code", I take it you are talking about the match block arms in each method? These are going to grow as we add more email templates to this system so it's definitely a fair assessment.

I like the idea of using a trait but I'm struggling a little to understand how it would serialize into the database column. For JSONB colums we need to implement quite a few `sqlx` traits on the type / enum so that it can be serialized in the db. Maybe this is possible with a type that implements a trait, I've just never done that before so I'm not sure how complicated it is, I'd worry we'd need to implement the same `sqlx` traits for every type instead of just the one enum. I'll look into it though. Maybe it's simpler than I'm thinking.

Also, I might be missing something, but for most of the enum methods the match block arms are different (if only slightly), the only one where the arms are the exact same is `mailer_slots_map` and I could probably rewrite this to be cleaner. I feel like with a trait we wouldn't be able to rely on default implementations for most methods so every `Slot` type would need its own full implementation of the trait, which is kind of getting to the same place with repeated code. Eg for the `type_str` method we'd have something like:
```rust
impl EmailTemplateSlot for ConversationInvite {
    fn type_str() -> &'static str {
        TYPE_CONVERSATION_INVITE
    }
}

impl EmailTemplateSlot for EventRegistrationInvite {
    fn type_str() -> &'static str {
        TYPE_EVENT_REGISTRATION_INVITE
    }
}

impl EmailTemplateSlot for EventRegistrationConfirmation {
    fn type_str() -> &'static str {
        TYPE_EVENT_REGISTRATION_CONFIRMATION
    }
}
```
And the same for most of the other methods.

Like I say though, I definitely could be misunderstanding what you're meaning so let me know if am and using a trait would have a bigger benefit than what I'm thinking.

alpinegeorge: We could side-step the whole issue for now by enforcing the size constraint like so:
```rust
use strum_macros::EnumCount;
use strum::EnumCount as _;

#[derive(EnumCount)]
enum Test{
    A,
    B,
    C,
}

fn test() -> [&'static str; Test::COUNT] {
    [
        "a",
        "b",
        "c",
    ]
}
```
The strum and strum_macros crates have a very small dependency graph. so they should have minimal impact on build times.

api/src/models/email_template_config.rs
```diff
@@ -33,189 +33,76 @@ pub struct EmailTemplateConfig {
...
 #[serde(rename_all = "snake_case")]
-enum ContentType {
-    PlainText,
-    RichText,
+pub enum EmailType {
+    #[sqlx(rename = "conversation_invite")]
+    ConversationInvite,
+    #[sqlx(rename = "event_registration_invite")]
+    EventRegistrationInvite,
+    #[sqlx(rename = "event_registration_invite")]
```
alpinegeorge: I think this should be "event_registration_confirmation", right?
Also, if that is the case, then these renames are no-ops since `rename_all = "snake_case"` will handle renaming automatically.

api/src/models/email_template_config.rs
```diff
@@ -33,189 +33,76 @@ pub struct EmailTemplateConfig {
...
-
-/// Converts `self` to [`HashMap`] mapping email template variable keys to their content.
-pub trait MailerContextMap {
-    fn mailer_context_map(&self) -> HashMap<String, String>;
+impl std::fmt::Display for EmailType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let value = match self {
+            EmailType::ConversationInvite => "conversation_invite",
+            EmailType::EventRegistrationInvite => "event_registration_invite",
+            EmailType::EventRegistrationConfirmation => "event_registration_invite",
```
alpinegeorge: "event_registration_invite" maps to multiple email types.

rosswilson2306: Woa! Great spot!

api/src/models/email_template_config.rs
rosswilson2306: The `rename_all` I had was `serde` and the db tests broke without the `sqlx` attributes as well, but looks like I can do:
```rust
#[sqlx(type_name = "TEXT", rename_all = "snake_case")]
#[serde(rename_all = "snake_case")]
```
which is better than me making silly mistakes like the above!

alpinegeorge (dismissed):
LGTM!

PR #478, opened 2026-06-22
Fixes sync bug and adds endpoints for moderation

stuartlynn:
This PR 

1. Fixes a quick bug where only approved statements where being synced from polis
2. Fixes a bug where the moderation state was not being synced properly
3. Adds an endpoint to moderate statements 
4. Adds some checking for permissions on polis operations

### DIFF
--- a/api/src/tools/polis.rs
+++ b/api/src/tools/polis.rs
@@ -515,7 +515,7 @@ async fn moderate_statement_aux(
 
     let updated = models::polis_statement_aux::update(
         &state.db,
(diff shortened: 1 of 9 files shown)


api/src/tools/polis.rs
alpinegeorge: The admin_login handler still exists even though it is now unused.
We also need to make sure that ui/packages/api-client/src/api.ts is updated to match this.

alpinegeorge (changes requested):

PR #487, opened 2026-06-23
486 disable send email button when email is being sent

### LINKED ISSUE #486: Disable send email button on the notify tab when participant emails are being sent
When we send emails to participants it can sometimes take a long time.  We should disable the send email button and show a loading spinner or similar instead while this happens 

Image

https://github.com/crownshy/comhairle/blob/49ed20ea9e49b782dde71d4fd975600652009242/ui/packages/comhairle/src/routes/(admin)/admin/conversations/%5Bconversation_id%5D/notifications/%2Bpage.svelte#L20

https://github.com/crownshy/comhairle/blob/49ed20ea9e49b782dde71d4fd975600652009242/ui/packages/comhairle/src/lib/components/NotificationForm.svelte#L18

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/486

### Testing summary

- Checked that the email button couldn't be clicked multiple times anymore

[Kooha-2026-06-23-17-29-43.webm](https://github.com/user-attachments/assets/b0ee5ae3-d5ab-4a18-bc93-2e9dcf3f8b82)

- Fixed issue where the body of the form wasn't being cleared after the email was sent
_Before:_

[Kooha-2026-06-23-17-33-34.webm](https://github.com/user-attachments/assets/cf82b982-f9fd-4a90-a0b1-b3b7ae3ed978)

_After:_

[Kooha-2026-06-23-17-34-02.webm](https://github.com/user-attachments/assets/6a051128-5e50-4d80-9933-eedc0db9f67e)

### DIFF
--- a/ui/packages/comhairle/src/lib/components/NotificationForm.svelte
+++ b/ui/packages/comhairle/src/lib/components/NotificationForm.svelte
@@ -56,11 +56,17 @@
 			validators: zodClient(notificationFormSchema),
 			taintedMessage: false,
 			validationMethod: 'oninput',
-			onSubmit: sendNotification
+			onSubmit: async (actions) => {
+				submitting = true;
+				await sendNotification(actions);
+				submitting = false;
+			}
 		}
 	);
 
-	const { form, enhance, validateForm, submitting, reset } = notificationForm;
+	const { form, enhance, validateForm, reset } = notificationForm;
+
+	let submitting = $state<boolean>(false);
 
 	let failedRecipients = $state<string[]>([]);
 	let lastSendMessage = $state<string | null>(null);
@@ -241,7 +247,7 @@
 							placeholder={isEmail
 								? 'Enter email subject...'
 								: 'Enter notification title...'}
-							disabled={$submitting}
+							disabled={submitting}
 						/>
 						<Form.FieldErrors />
 					</div>
@@ -260,7 +266,7 @@
 							<RichTextEditor
 								value={$form.content || null}
 								placeholder="Compose your email..."
-								editable={!$submitting}
+								editable={!submitting}
 								onChange={(json) => ($form.content = json)}
 							/>
 							<input type="hidden" {...props} value={$form.content} />
@@ -270,7 +276,7 @@
 								bind:value={$form.content}
 								placeholder="Enter your notification message here..."
 								rows={4}
-								disabled={$submitting}
+								disabled={submitting}
 							/>
 						{/if}
 						<Form.FieldErrors />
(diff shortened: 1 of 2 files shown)


ui/packages/comhairle/src/lib/components/RichTextEditor/RichTextEditor.svelte
```diff
@@ -2,10 +2,6 @@
 	import { onMount, onDestroy } from 'svelte';
 	import { useResizeObserver } from 'runed';
 	import { Editor } from '@tiptap/core';
-	import { Color } from '@tiptap/extension-color';
-	import { ListItem } from '@tiptap/extension-list-item';
-	import { TextStyle } from '@tiptap/extension-text-style';
-	import { Underline } from '@tiptap/extension-underline';
```
ADSmith-0, line 8: Removed these as in my editor it flagged that they weren't used anywhere

ui/packages/comhairle/src/lib/components/RichTextEditor/RichTextEditor.svelte
```diff
@@ -28,8 +24,7 @@
 	};
 
 	let {
-		value = null,
-		placeholder = 'Start typing...',
```
ADSmith-0, line 32: Similarly it looks like this wasn't used anywhere either, I'm not sure if we'd prefer to show it as a default value instead of getting rid of it?

ui/packages/comhairle/src/lib/components/RichTextEditor/RichTextEditor.svelte
```diff
@@ -28,8 +24,7 @@
...
+		value = $bindable(),
```
ADSmith-0, line 27: Changed the value to be bindable so that the value stays in sync

ui/packages/comhairle/src/lib/components/NotificationForm.svelte
```diff
@@ -56,11 +56,17 @@
...
+				await sendNotification(actions);
+				submitting = false;
+			}
 		}
 	);
 
-	const { form, enhance, validateForm, submitting, reset } = notificationForm;
+	const { form, enhance, validateForm, reset } = notificationForm;
+
+	let submitting = $state<boolean>(false);
```
ADSmith-0, line 69: The issue here seemed to be that the `submitting` finishes quite quickly but then the form doesn't reset quick enough afterwards, so you can then submit again

So instead of using the `submitting` value from the `notificationForm` I created a client side one that will only finish once the form has either completed or rejected

dtodorowa:
loooove the recordings hahah this is what i try to do too! Looks good!

dtodorowa (approved):

PR #488, opened 2026-06-24
Delete workflow step (backend)

### LINKED ISSUE #476: Delete workflow step
Deleting a workflow_step requires an implementation of the `ToolImpl` trait's `delete` method https://github.com/crownshy/comhairle/blob/2cb82bd1c3521fbc69c57b52eee2fcd310ee2c93/api/src/tools.rs?plain=1#L70-L77 for each tool, where the implementation will clean up the resources for the associated tool, both in the comhairle DB and with any associated third parties. 

Learning:
- should be as simple as deleting the workflow_step as data is contained within `workflow_step.tool_config` in comhairle DB

Polis:
- delete polis conversation (id stored in workflow_step.tool_config or workflow_step.preview_tool_config)
- delete comments for polis conversation, it doesn't look like there is a native way of deleting comments, only setting them to `!active`, see https://github.com/crownshy/polis/blob/99c3dbb8f5e55cd36a1c637c9d99ded7888c77fa/server/app.ts?plain=1#L944, https://github.com/crownshy/polis/blob/99c3dbb8f5e55cd36a1c637c9d99ded7888c77fa/server/src/server.ts?plain=1#L6592
- we could add a route to the `comhairle` branch of `crownshy-polis` to delete a comment, or mark them as inactive (we will likely be moving to voxit at some point so deleting comments may come out the box with that and marking as inactive may be enough in the meantime)

Heyform:
- delete heyform form
- heyform uses a graphql api layer that we can interact with using the Heyform client `execute_graphql` method with an associated query const, see https://github.com/crownshy/comhairle/blob/2cb82bd1c3521fbc69c57b52eee2fcd310ee2c93/adaptors/heyform-rust-sdk/src/client.rs?plain=1#L217, which corresponds with a execution const in the heyform app https://github.com/crownshy/heyform/blob/83bdee5425dcc51ec5b0db2f0feedb65217078bd/packages/webapp/src/service/form.service.ts?plain=1#L18
- Delete form submissions https://github.com/crownshy/heyform/blob/83bdee5425dcc51ec5b0db2f0feedb65217078bd/packages/webapp/src/service/submission.service.ts?plain=1#L36-L46

Prioritization:
- all comhairle DB data, no third party data involved
- all proposal_evaluation_proposal for a given workflow_step
- all proposal_evalution_proposal_response for each proposal_evaluation_proposal

Thinking space:
- ragflow agent user sessions, a record of these are stored in the comhaire DB in the `bot_service_user_session` table https://github.com/crownshy/comhairle/blob/2cb82bd1c3521fbc69c57b52eee2fcd310ee2c93/api/src/tools/thinking_space.rs?plain=1#L380-L387
- use the `bot_service_session_id` in these records to delete the correct sessions from ragflow https://github.com/crownshy/comhairle/blob/2cb82bd1c3521fbc69c57b52eee2fcd310ee2c93/adaptors/ragflow/src/agent/session.rs?plain=1#L21-L30 then delete the records themselves

Elicitation bot:
- same as Thinking space, delete the agent user sessions using the `bot_service_user_session` table https://github.com/crownshy/comhairle/blob/2cb82bd1c3521fbc69c57b52eee2fcd310ee2c93/api/src/tools/elicitation_bot.rs?plain=1#L207-L214 then delete the table records for that workflow step

Stories:
- I don't believe we are actually using this tool but there doesn't seem to be any third party or additional comhairle DB code in the tool trait implementation so it should be as simple as deleting the workflow_step

alpinegeorge:
Motivations:

- Currently, workflow steps cannot be removed after addition.

Actions:

- Add workflow step deletion with cleanup for related tool resources.
- Added test buttons to comhairle frontend for testing (visibility controlled by testButtons query parameter)

### DIFF
--- a/adaptors/heyform-rust-sdk/src/client.rs
+++ b/adaptors/heyform-rust-sdk/src/client.rs
@@ -264,6 +264,20 @@ impl HeyFormClient {
         Ok(response.get("updateFormSchemas").copied().unwrap_or(false))
     }
 
+    pub async fn delete_poll(&self, form_id: &str) -> Result<bool> {
+        let variables = json!({
+            "input": {
+                "formId": form_id
+            }
+        });
+
+        let response: HashMap<String, bool> = self
+            .execute_graphql(DELETE_FORM_MUTATION, variables, Some("deleteForm"))
+            .await?;
+
+        Ok(response.get("deleteForm").copied().unwrap_or(false))
+    }
+
     /// Set custom CSS on a form
     pub async fn set_custom_css(&self, form_id: &str, custom_css: &str) -> Result<bool> {
         let theme = FormTheme {
--- a/adaptors/heyform-rust-sdk/src/queries.rs
+++ b/adaptors/heyform-rust-sdk/src/queries.rs
@@ -29,6 +29,12 @@ mutation updateForm($input: UpdateFormInput!) {
 }
 "#;
 
+pub const DELETE_FORM_MUTATION: &str = r#"
+mutation deleteForm($input: FormDetailInput!) {
+  deleteForm(input: $input)
+}
+"#;
+
 pub const UPDATE_FORM_SCHEMAS_MUTATION: &str = r#"
 mutation updateFormSchemas($input: UpdateFormSchemasInput!) {
   updateFormSchemas(input: $input)
--- a/adaptors/heyform-rust-sdk/src/types.rs
+++ b/adaptors/heyform-rust-sdk/src/types.rs
@@ -176,6 +176,12 @@ pub struct UpdateFormThemeInput {
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
+pub struct FormDetailInput {
+    pub form_id: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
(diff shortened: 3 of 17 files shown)


dtodorowa:
this is maybe more of a behaviour question. I gave it a quick test - works ok for conversations that are not yet launched

when i (by accident lol) tried to delete a step from a launched convo I got `workflow_step not found`

```bash
curl 'http://localhost:5176/api/conversation/9080acf2-d49a-4966-b427-eca201be2473/workflow/aa467403-9487-4f12-88b0-33f42670cce1/workflow_step/5c8d0ad3-d89c-4708-a2ec-b925a186794d' \
  -X 'DELETE' \
  etc etc
  ```

Guess my question would be: do we want to be able to delete steps in a launched convo? Maybe this is an actual requirement that I missed? Either way would like to align, so I can make sure the frontend reflects this. 

@stuartlynn @alpinegeorge lmk your thoughts

dtodorowa:
Adding another comment bc diss-cord doesnt let me upload larger vids

1. worked with a newly created convo
2. didnt work with an old live convo i had. Preview said `"workflow_step not found"`

prob worth testing on staging

https://github.com/user-attachments/assets/2e695316-a769-49b3-814b-f476b50e456f

dtodorowa (approved):

PR #508, opened 2026-06-26
Couldn't register event as invited user

### LINKED ISSUE #309: Couldn't register to event as invited user (logged in)
When an admin created an event, invited a test user email 
Open another browser that is logged in with test user email (not admin) 
test user receives an email with event link
Copy paste the event link to the browser signed in with test user's account. 
Expecting to see evert registration success but got brought to homepage instead.

alpinegeorge:
Motivations:

 - When attempting to join a conversation that the user has been invited to, the user will occasionally remain on the same page instead of joining the new conversation. The issue resolves on a refresh, but leads to a confusing user experience.

Actions:

 - This looks like it's due to a race condition between `invalidateAll` and `goto` calls in the same scope. Replaced all cases with `goto(..., { invalidateAll: true })` invocations instead.

### DIFF
--- a/ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/+page.svelte
+++ b/ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/+page.svelte
@@ -74,8 +74,7 @@
 			params: { conversation_id: data.conversation.id, workflow_id: firstWorkflow.id }
 		});
 
-		await invalidateAll();
-		goto(firstWorkflowPath);
+		goto(firstWorkflowPath, { invalidateAll: true });
 	}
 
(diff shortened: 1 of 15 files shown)


ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/+page.svelte
```diff
@@ -74,7 +74,7 @@
 			params: { conversation_id: data.conversation.id, workflow_id: firstWorkflow.id }
 		});
 
-		await invalidateAll();
+		invalidateAll();
```
Comment (alpinegeorge) [inline ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/[[preview]]/+page.svelte]: Leftover from testing eslint rule!

dtodorowa (dismissed):

PR #513, opened 2026-06-26
API changes to support individual audio recording uploads

### LINKED ISSUE #490: Individual audio recordings
[Issue #477 - Audio recording upload](https://github.com/crownshy/comhairle/issues/477) introduced a self-serve audio recording upload path to the API.
Currently, audio recordings are all uploaded simultaneously, however it would be better if we allowed for individual recordings to be uploaded, giving user's more flexibility on their upload schedule.

mag:
See issue #490

Reworks audio recordings from a single per-event record into one record per named recording; replaces the batch endpoints with a RESTful per-recording API, and tracks each recording's progress through the transcription → categorization pipeline
  with distinct states.

  Key changes

  - Data model. audio_recording is now one row per recording, each with a name that's unique within its event (UNIQUE(event_id, name)). Drops the old single-row-per-event model and its breakout_room_ids array. S3 layout is
  events/{event_id}/recordings/{recording_id}/….
  - New REST API under /conversation/{id}/events/{id}/audio_recordings: create (returns a presigned upload URL), list, get (with presigned download URLs), …/process, and the categorization …/report webhook — all scoped to a single recording. The old
  event-scoped upload / download / batch-transcriptions / report endpoints are removed.
  - Per-recording status lifecycle. Replaces the event-wide pending/completed/failed with pending → transcript_available → both_available, plus transcript_failure / categorization_failure. Completion is now driven by the report webhook actually
  arriving, not by submission to the categorization service.
  - Pipeline. The Apalis worker processes one recording by id, reads its stored S3 prefix as the single source of truth, and writes the recording's status at each stage boundary.

  Notes

  - Breaking: removes the old endpoints/model and will cause the generation of a new OpenAPI contract, causing the UI to fail to build. So this should not be merged quite yet, as it will break anybody who is changing any part of the API and moving the frontend forward. 
  - A migration alters audio_recording in place and clears existing rows (prototype data only). This PR will delete data when deployed to production. @stuartlynn and @alpinegeorge  please confirm this is OK.
  - Deferred: no reconciliation/timeout for a report webhook that never fires (recording stays transcript_available); no stage-specific retry endpoints yet (the status split sets this up).

### DIFF
--- a/api/migrations/20260626120000_make_audio_recording_per_room.sql
+++ b/api/migrations/20260626120000_make_audio_recording_per_room.sql
@@ -1,0 +1,20 @@
+-- Make audio_recording per-room instead of per-event.
+--
+-- Each row now represents a single named room within an event, and an event may
+-- have many rooms. The previous model kept one row per event with the breakout
+-- rooms stored in a TEXT[] array; that distinction (main vs. breakout) is gone.
+--
+-- The existing data is from a throwaway prototype with no users, so we discard it
+-- up front. That lets us add a NOT NULL `name` column without a backfill.
+DELETE FROM audio_recording;
+
+-- Many rooms per event now, so the event_id can no longer be unique on its own.
+ALTER TABLE audio_recording DROP CONSTRAINT audio_recording_event_id_key;
+
+-- The main-vs-breakout concept is gone; rooms are tracked as their own rows.
+ALTER TABLE audio_recording DROP COLUMN breakout_room_ids;
+
+-- User-supplied room name, unique within an event.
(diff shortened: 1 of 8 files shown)


alpinegeorge (approved):
Looking good Mike. Well done!

alpinegeorge:
@mag Pertaining to deleting data.
There is currently only a single set of recordings we uploaded for bloom both to transcribe the data for them and to test the original audio recording uploads endpoints. Deleting that data from the database should be okay since we could use the same recordings to test the individual upload pathways once in staging.

PR #514, opened 2026-06-29
Bug Fixes: Improve Polis embed UI state handling 

### LINKED ISSUE #171: Polis bug when adding a new opinion after voting on all opinions
How to recreate the bug

- Vote on all statements 
- Add your own 
- Click back to voting 

### Expected behavior

It should show the completed page.

### What actually happens 

It shows a blank voting card.

### LINKED ISSUE #352: "opinion x of y" but x>y
**Issue**
"opinion 5 of 4" seems strange. We shouldn't have 5 opinions out of 4 opinions
Image

**Reproduce issue**
- participant go through all polis statements
- add own opinions
- hit "back voting"

I suspect the total number of opinion should be updated based on new opinions user created?

**Expected outcome** 
In "opinion x of y", x should always be less of equal to y. and y grows when the total number of collective opinions grow.

### LINKED ISSUE #372: ST-Sh-9 Polis Required votes not updated
https://docs.google.com/spreadsheets/d/1zT6pAuUAJRgb3IufwH9rMrydX3fyrEF9fXz2AElnouI/edit?gid=1486323079#gid=1486323079

ST-Sh-9 

Issue: 
Polis  Required votes not updated

How to replicate the issue:
When updating polis' required votes from 10 to 5, refresh, the input field reverted back to 10.

dtodorowa:
Should fix #372 #171 #352 

- Extract opinion counter calculation into separate module with tests
- Fix progress bar calculation to use safe required votes default
- Prevent "Opinion X of Y" counter from exceeding total when pool exhausted
- Add loading states to opinion submission and continue buttons
- Display error message when opinion submission fails
- Anchor opinion counter to prevent jumps when pool size changes mid-session
- Improve skeleton state
----

### Some context, if you have questions:

- `voting` shows the current statement, the "Opinion X of Y" counter, and the Agree / Disagree / Pass + "Add your own" buttons.
- `add-opinion` the participant is writing their own opinion.
- `continue-prompt` reached the required number of votes; offered "keep voting" or "continue to next step".
- `completed` nothing (left) to vote on; shows "you've voted on everything / come back later" + "Add your own".

The counter is only rendered when the **"Show remaining statements"** toggle is on (`show_remaining_statements` in the tool config).

## "Opinion X of Y" counter (`opinionCounter` in `polisCounter.ts`)

`X` = position currently shown (1-based), `Y` = total statements in the pool.
Logic is a pure function, and I wrote tests for it (`polisCounter.test.ts`) to avoid another bug.

| total (Y) | remaining | shows    | why                                                        |
| --------- | --------- | -------- | ---------------------------------------------------------- |
| 4         | 4         | `1 of 4` | nothing voted yet                                          |
| 4         | 1         | `4 of 4` | on the last statement                                      |
| 4         | 0         | `4 of 4` | **all voted — never `5 of 4`** (the original bug)          |
| 5         | 1         | `5 of 5` | a 5th opinion was added; denominator grew                  |
| 0         | 0         | `0 of 0` | empty pool (but screen flips to `completed`, so not shown) |
| 4         | 10        | `1 of 4` | remaining clamped to ≤ total (jittery Polis response)      |
| 4         | -1        | `4 of 4` | negative remaining clamped to 0                            |

Key rules:

- `X` can never exceed `Y`. It is capped so finishing the pool shows "Y of Y".
- `Y` grows when opinions are added. The anchored counts re-sync to the live Polis totals whenever the pool grows (`PolisEmbed` anchor effect), and the pool is refreshed immediately after a participant submits an opinion
  (`fetchNextStatement` in the submit handlers).
- A participant generally can't vote on their **own** submitted opinion, so after adding one you may see e.g. "4 of 5" and then go straight to `completed` -> `Y` grew even though there's nothing new for _you_ to vote on. This is expected.

## Empty / exhausted pool

`poolExhausted` = Polis is ready, not loading, not in error, and has **no current statement**. Covers:

1. A poll with no statements at all (e.g. admin preview of a fresh poll).
2. A poll where the participant has voted on everything available.
3. The moment right after submitting your own (non-votable) opinion empties the queue.

In all three cases the screen must show **`completed`**, not the voting UI. This is enforced by a reactive effect (not just the Polis callback) so that screen changes made elsewhere, e.g. the timed return-to-voting after submitting an opinion (can't strand the user on "Opinion 0 of 0" with live vote buttons). As a belt-and-braces measure the counter and vote buttons are also hidden when there is no current statement.

## Submitting an opinion

- Only show statements **from other participants**. This is enforced by Polis: submitting a statement sends `vote: -1` (an auto-vote on your own comment), so `nextComment?...&not_voted_by_pid=<pid>` excludes it. Your own opinion is never shown back to you to vote on.
    - **PID race:** `not_voted_by_pid` is only sent once `_pid` is known. If `participationInit` fails, the first fetch is unfiltered -> harmless for a new voter (nothing voted yet), but a returning voter could briefly be shown an already-voted statement until a pid is resolved from the next response.
- A failed submission keeps the participant's text and shows an error, rather than flashing a fake "submitted" confirmation and discarding what they wrote.
- Submit buttons are disabled while a submission is in flight (no double-submit).

## Required votes (admin `PolisManage.svelte`)

- The field mirrors the saved value via local state synced with an effect, so it  always reflects what's persisted after a reload / `invalidateAll`.
- Empty, non-numeric, or `< 1` input is **not** saved (empty would wipe the value mid-edit; `< 1` breaks the threshold/progress maths). Input has `min="1"`.
- On the participant side, a missing/`0`/`null` `required_votes` falls back to `10` (`safeRequiredVotes`) so "continue" still unlocks and the progress bar is sane.
- Saves to `tool_config` when the conversation is live, otherwise `preview_tool_config`: test in the same mode you edited in.

## Counted toward the "continue" threshold

- **Pass / Unsure counts as a vote** toward the required-votes threshold (same as Agree/Disagree). Aka, all
  three button types call `incrementVotes`.
- A stray vote with no statement on screen is ignored (`doVote` guards on `polisCurrentStatement`).

## Manual test checklist

Voting (incognito, "Show remaining statements" on, poll with ~4 statements):

- [ ] Fresh start shows `Opinion 1 of N`.
- [ ] Counter advances and never exceeds `N`.
- [ ] Voting the final statement shows `N of N`, then `completed`, never `N+1 of N`.
- [ ] Add an opinion → total `N` grows.
- [ ] Empty poll / after submitting your own opinion → `completed` screen, never "Opinion 0 of 0" with vote buttons.
- [ ] Your own submitted opinion is never shown back to you to vote on.
- [ ] Submit an opinion while offline → text is kept + error shown (no fake "submitted").

Required votes (admin → Design → Polis step → Setup):

- [ ] Change 10 → 5, hard-refresh → still 5.
- [ ] Clear the field, or enter `0` → no garbage persisted.

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/polis/PolisEmbed.svelte
+++ b/ui/packages/comhairle/src/lib/tools/polis/PolisEmbed.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { Button } from '$lib/components/ui/button';
+	import { Button, LoadingButton } from '$lib/components/ui/button';
 	import { fly, fade } from 'svelte/transition';
 	import { cubicOut } from 'svelte/easing';
 	import {
@@ -14,6 +14,7 @@
 	} from 'lucide-svelte';
 	import PolisApi, { type PolisApiState, type PolisStatement } from './PolisApi';
 	import { getVoteData, incrementVotes, resetVoteCount } from './polisVoteStore';
+	import { opinionCounter } from './polisCounter';
 	import * as m from '$lib/paraglide/messages';
 	import Separator from '$lib/components/ui/separator/separator.svelte';
 	import { apiClient } from '@crownshy/api-client/client';
@@ -22,7 +23,7 @@
 		polis_id: string;
 		polis_url: string;
 		user_id: string;
-		onDone: () => void;
+		onDone: () => void | Promise<void>;
 		requiredVotes?: number;
 		workflowStepId?: string;
 		showRemainingStatementCount?: boolean;
@@ -67,10 +68,6 @@
 			previousText = newTxt;
 			waitingForNext = false;
 		}
-
-		if (screen === 'voting' && s.ready && !s.loading && !s.currentStatement && !s.error) {
-			screen = 'completed';
-		}
 	}
 
 	const polis = new PolisApi(user_id, polis_id, handlePolisChange, 'en', polis_url);
@@ -85,8 +82,19 @@
 	let voteCooldown = $state(false);
 	let opinionText = $state('');
 	let opinionSubmitted = $state(false);
+	let opinionSubmitting = $state(false);
+	let opinionError = $state(false);
+	let returningToVoting = $state(false);
+	const submitBusy = $derived(opinionSubmitting || returningToVoting);
 	let previousText = '';
 	let visibleStatementWhenOpened: PolisStatement | undefined = undefined;
+
+	// `required_votes` is optional in the tool config and can arrive as null/0,
+	// which would break the threshold and progress maths. Fall back to a sane
+	// positive default so "continue" still unlocks correctly.
+	const safeRequiredVotes = $derived(
+		typeof requiredVotes === 'number' && requiredVotes > 0 ? Math.floor(requiredVotes) : 10
+	);
 
 	async function createStatementAux(
 		newStatement: { tid: number; pid: number },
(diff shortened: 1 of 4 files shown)


alpinegeorge (approved):
LGTM

PR #520, opened 2026-06-30
Fixing up audio recodring interface to algin with new api

### LINKED ISSUE #491: Audio recording upload (frontend)
We currently have a dev-art interface for the audio recording upload (Issue #477) admin interface, however we could use a properly designed interface in the comhairle frontend.

stuartlynn:
Aligns our UI with #490 .

### DIFF
--- a/ui/packages/api-client/src/api.ts
+++ b/ui/packages/api-client/src/api.ts
@@ -1241,6 +1241,35 @@ export const UpdateUserProgress = z
   .partial()
   .passthrough();
 export type UpdateUserProgress = z.infer<typeof UpdateUserProgress>;
+export const RecruitmentTargetDto = z
+  .object({
+    bucket: z.string(),
+    createdAt: z.string().datetime({ offset: true }),
+    id: z.string().uuid(),
+    metric: z.string(),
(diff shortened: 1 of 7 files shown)


ui/packages/api-client/src/api.ts
alpinegeorge, line 1: We should bump the api client version in ui/packages/api-client/packages.json, right?

dtodorowa:
Looks good, one thing that came to mind: if the client never completes the upload (closed tab, dropped network), we're left with a pending recording (if i understand correctly) A retry with the same name would now 409 on the unique constraint.

What should the cleanup path be? I assume the expectation is not picking a new name?

dtodorowa:
Tried testing, but got a 500 "No bulk storage service configured". Do I just add an env variable to configure it? 
What do we use for testing?

dtodorowa (dismissed):

ui/packages/api-client/src/api.ts
stuartlynn: Yeah going to do that on a separate PR

stuartlynn:
Added a delete button to clear up failed uploads and more specific statuses for uploads

PR #526, opened 2026-07-01
413 allow users to upload or select uploaded media in rich text

### LINKED ISSUE #413: Allow users to upload or select uploaded media in rich text fields
Currently images and videos in the rich text contexts can only be linked to. We should allow users to upload media or pick it from a library of media they have already uploaded.

The motivation for this is two fold. 

Firstly not all media is already hosted online and asking our users to get it somewhere publicly accessible is a barrier  

Secondly  safari  is pretty strict on cross domain image / video hosting. This can mean that images that are just linked sometimes dont work.

Notes

We should keep the ability to link in media, this should be in addition to that.

We should create a generic media browser / upload component that can be also used in other places in the application. 

We should add a media tab in admin pane (probably at the same level as the conversation list) to allow a user to manage their media.

### Acceptance criteria:
For the following media includes: images, video and audio

#### Front end:
- [x] Media library should be accessible from the admin nav menu
- [x] Media library page should have the ability to upload media
  - [ ] Image uploads should be restricted to media file types, any incorrect file types should error
- [ ] Media library should give the ability to show/play the media
- [ ] Media library should allow the user to delete media
- [ ] Media library should allow the user to bulk delete media
- [ ] Rich text editor should give the user the ability to upload media that will then be put into the rich text editor
  - [ ] Any media uploaded to the rich text editor should appear in the media library

#### Back end:
- [ ] Images should be optimised when uploaded to the backend (convert to webp)
- [ ] Videos should be optimised when uploaded to the backend (convert to webm/mp4)

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/413

Still a bit buggy and rough around the edges, but wanted to get an initial PR up and ready for review, I'll spin off all of the functionality and bug fixes we need into separate tickets 🙂️

Uploading video, audio and images, as well as deleting them:

https://github.com/user-attachments/assets/ba9dd7d5-98d5-4695-a328-36ac5e481cf2

Also found this bug where the user wasn't able to unlink a link without deleting the text, so I fixed that quickly as well
- Before

https://github.com/user-attachments/assets/ff0c4677-cfb2-44ff-9cf8-03fe2fa86433

- After

https://github.com/user-attachments/assets/c0e4ffe2-6a19-48e3-89ea-ddd78a48256e


ui/packages/comhairle/src/lib/components/ui/input/input.svelte
```diff
@@ -1,16 +1,16 @@
...
 	type Props = WithElementRef<
-		Omit<HTMLInputAttributes, "type"> &
-			({ type: "file"; files?: FileList } | { type?: InputType; files?: undefined })
+		Omit<HTMLInputAttributes, 'type'> &
+			({ type: 'file'; files?: FileList } | { type?: InputType; files?: undefined })
 	>;
 
 	let {
-		ref = $bindable(null),
+		ref = $bindable(),
```
ADSmith-0, line 13: There was an issue here where the `null` was causing the binding to error the first time it was trying to use it, so I removed it to fix that

ui/packages/comhairle/src/lib/components/ui/input/input.svelte
```diff
@@ -19,14 +19,14 @@
 	}: Props = $props();
 </script>
 
-{#if type === "file"}
+{#if type === 'file'}
```
ADSmith-0, line 22: Also sorry about the `'` and the changed tailwind classes, I have a formatter that I run in my setup, so I think this might be the eslint file or the prettier config or something? It was doing it all automatically

ui/packages/comhairle/src/lib/components/ui/sidebar/sidebar-inset.svelte
```diff
@@ -10,15 +10,15 @@
 	}: WithElementRef<HTMLAttributes<HTMLElement>> = $props();
 </script>
 
-<main
+<div
```
ADSmith-0, line 13: We already have a `main` tag [here](https://github.com/crownshy/comhairle/blob/7ab6b9aea9df33cad4d0ad4c5321dcdf637e6adb/ui/packages/comhairle/src/routes/(admin)/%2Blayout.svelte#L29) and based  on the spec you should only have 1 `main` tag, https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element, so I changed it to a `div`

ui/packages/comhairle/src/lib/utils/errorHandling.ts
ADSmith-0, line 1: Made some functions to make error handling a bit easier, rather than having to nest try-catch's, we can instead read the errors as variables, similar to something like Go or Rust

Since we're in a repo with Rust, I copied the syntax (`Ok` and `Err`), so it would be something like

```ts
const result = await tryCatchAsync(() => response.json());
if (result.err !== null) {
  // Handle the error
  notifications.send({
    message: "Couldn't parse json",
    priority: "ERROR"
  });
  return;
}
console.log(result.ok) // Discriminated union, so we know this value is definitely not `null` anymore;
```

ui/packages/comhairle/src/lib/utils/errorHandling.ts
```diff
@@ -0,0 +1,66 @@
...
+export type FetchErr =
+	| { id: 'NETWORK_ERROR'; message: typeof NO_INTERNET }
+	| { id: 'HTTP_ERROR'; status: number; message: string };
+/**
+ * Wrapper for fetch that will feed back an error whenever the response is not 200-299
+ * @param endpoint - URL to fetch
+ * @param params - Fetch parameters
+ * @param [fetchRef=fetch] - Fetch reference, will use the default fetch, but pass the reference on the backend if used in a +page.server.ts file
+ */
+export async function tryFetch(
```
ADSmith-0, line 43: Similarly with `fetch`, it can be pretty frustrating because the promise might have resolved ok, but then the request failed with like a `500` or something, so I did this as a wrapper to try and make it a little easier to use

Example usage:
```ts
const response = await tryFetch("/api/media");
if (response.err !== null) {
  switch (response.err) {
    case "NETWORK_ERROR":
    // Handle here
    break;
    case "HTTP_ERROR":
    notifications.send({
      message: response.err.message,
      priority: "ERROR"
    });
    break;
  }
}
console.log(response.ok);
```

ui/packages/comhairle/src/lib/components/UrlInputPopover/UrlInputPopover.svelte
```diff
@@ -1,117 +0,0 @@
...
-		onOpenChange(false);
-	}
-
-	function handleKeydown(e: KeyboardEvent) {
-		if (e.key === 'Enter') {
-			e.preventDefault();
-			handleSubmit();
-		} else if (e.key === 'Escape') {
-			e.preventDefault();
-			onOpenChange(false);
```
ADSmith-0, line 77: Removed this when I copied it over as it's handled by the library natively: https://www.bits-ui.com/docs/components/dialog#escape-key-handling

ui/packages/comhairle/src/lib/components/UrlInputPopover/UrlInputPopover.svelte
```diff
@@ -1,117 +0,0 @@
...
-
-	// Focus input when popover opens
-	$effect(() => {
-		if (open && inputElement) {
-			// Small delay to ensure popover is fully rendered
-			setTimeout(() => {
-				inputElement?.focus();
-			}, 100);
-		}
-	});
```
ADSmith-0, line 49: Removed this when I copied the file over as it's handled natively by the library: https://www.bits-ui.com/docs/components/dialog#open-focus

ui/packages/comhairle/src/lib/components/UrlInputPopover/UrlInputPopover.svelte
ADSmith-0, line 1: I moved this file to the rich text editor as we weren't using it anywhere else, so I thought it would be better to colocate it, until we need it accessible globally

ui/packages/comhairle/src/lib/components/RichTextEditor/RichTextEditor.svelte
```diff
@@ -219,18 +214,10 @@
 		<EditorToolbar
 			{editor}
 			{activeStates}
-			bind:showLinkPopover
-			bind:showImagePopover
-			bind:showVideoPopover
-			bind:showDocumentPopover
```
ADSmith-0, line 225: The state for opening and closing popovers is handled internally by `bits-ui` already

Since we weren't using the listeners for anything except opening and closing, I thought it would make sense to remove them for now

ui/packages/comhairle/src/lib/components/RichTextEditor/EditorToolbar.svelte
```diff
@@ -23,48 +18,64 @@
...
 	} from 'lucide-svelte';
+	import type { ComponentType } from 'svelte';
 
 	type Props = {
 		editor: Editor | undefined;
 		activeStates: ActiveStates;
-		showLinkPopover: boolean;
-		showImagePopover: boolean;
-		showVideoPopover: boolean;
-		showDocumentPopover: boolean;
```
ADSmith-0, line 37: Same as here: https://github.com/crownshy/comhairle/pull/526/changes#r3520834159

ui/packages/comhairle/src/lib/components/RichTextEditor/EditorToolbar.svelte
```diff
@@ -103,62 +114,42 @@
...
+				active: activeStates.underline,
+				onclick: () => editor?.chain().focus().toggleUnderline().run(),
+				text: 'U',
+				classes: 'underline'
+			})}
 		</div>
 
 		<!-- Mobile/Compact "more" toggle -->
-		{#if !compact}
+		{#if compact}
```
ADSmith-0, line 152: `!compact` meant it wasn't showing up on mobile so I switched this over

ui/packages/comhairle/src/lib/components/RichTextEditor/EditorToolbar.svelte
```diff
@@ -177,195 +168,140 @@
...
+			<UrlInputPopover
+				type="audio"
+				allowLocalSelection
+				onSubmit={(url) => {
+					editor
+						?.chain()
+						.focus()
+						.setAudio({ src: url, autoplay: false, controls: true })
+						.run();
+				}}
```
ADSmith-0, line 285: Moved a lot of the functionality to be inside of the `UrlInputPopover` to reduce repetition

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/UrlInputPopover.svelte
```diff
@@ -0,0 +1,157 @@
...
+	function validate(url: string) {
+		switch (type) {
+			case 'video':
+				return validateIframeUrl(url, DEFAULT_ALLOWED_DOMAINS);
+			case 'audio':
+			case 'image':
+			case 'link':
+				return validateUrl(url);
+		}
+	}
```
ADSmith-0, line 56: Moved the validate to be inside of the `UrlInputPopover` so it gets handled automatically

ui/packages/comhairle/src/lib/interfaces/Media.ts
```diff
@@ -0,0 +1,92 @@
...
+				filesArray.push(f);
+			}
+		} else {
+			filesArray.push(files as File);
+		}
+		return filesArray;
+	}
+
+	async upload(to: string, files: File, opts?: Opts): Promise<UploadReturn>;
+	async upload(to: string, files: File[], opts?: Opts): Promise<UploadReturn[]>;
```
ADSmith-0, line 75: Overloaded function signature to hopefully make it easier to deal with, so

single file -> single response
array of files -> array of responses

ui/packages/comhairle/src/lib/utils/contentDetection.ts
```diff
@@ -1,36 +1,37 @@
...
+ *
  * detectContentType('# Hello World')
  * // Returns: { content: '# Hello World', type: 'markdown' }
  * ```
  */
-export function detectContentType(content: string | undefined): { 
-	content: any; 
-	type: ContentType 
+export function detectContentType(content: string | null | undefined): {
+	content: Content;
```
ADSmith-0, line 26: Fixed TS errors

ui/packages/comhairle/src/lib/interfaces/Media.ts
ADSmith-0, line 1: Interface to hopefully unify the uploading and processing of files as we have the same logic repeated in a few different files throughout the codebase

ui/packages/comhairle/src/routes/(admin)/admin/media-library/+page.svelte
```diff
@@ -0,0 +1,101 @@
...
+					{#if bulkEdit}
+						<label>
+							<input
+								type="checkbox"
+								name="media"
+								value={media.id.toString()}
+								class="accent-primary absolute top-2 left-2 z-2 h-4 cursor-pointer"
+								bind:group={selected}
+							/>
+							<span>
```
ADSmith-0, line 84: Added `span` so we can target the component with the selector `input:checked + * `

ui/packages/comhairle/src/routes/(admin)/+layout.svelte
```diff
@@ -26,7 +26,7 @@
 	<AdminNav user={data.user} conversations={conversations.records} path={page.url.pathname} />
 	<SideBar.Inset>
 		<SidebarFloatingTriggers />
-		<main class="bg-muted flex flex-grow flex-col overflow-y-auto">
+		<main class="bg-muted flex grow flex-col overflow-y-auto">
```
ADSmith-0, line 29: Fixed TS error about incorrect tailwind class

ui/packages/comhairle/svelte.config.js
```diff
@@ -7,7 +7,8 @@ const TRUSTED_SOURCES = [
 	'https://polis.comhairle.scot',
 	'https://www.youtube.com',
 	'https://youtube.com',
-	'https://www.youtube-nocookie.com'
+	'https://www.youtube-nocookie.com',
+	'https://comhairle-media-test.s3.amazonaws.com'
```
ADSmith-0, line 11: Not sure if we need this here, but thought it might be good to add, to make sure it doesn't blocked?

api/src/routes/media/dto.rs
```diff
@@ -32,6 +33,7 @@ impl From<Media> for MediaDto {
     fn from(m: Media) -> Self {
         Self {
             id: m.id,
+            url: m.url().clone(),
```
rosswilson2306: @ADSmith-0 Did you definitely need the clone here now that you hoisted the `m.url()` call above moving `m.store_name` and `m.storage_key`? `m.url()` returns and owned String so I would have thought that you wouldn't need the clone now that the ordering had been changed?

ui/packages/comhairle/messages/en.json
```diff
@@ -6,6 +6,10 @@
 	"this_is_your_anonymous_id": "This is your anonymous ID",
 	"save_it_youll_need_it_to_log_in": "Save it - you'll need it to log in.",
 	"continue_": "Continue",
+	"cancel": "Cancel",
+	"delete": "Delete",
+	"edit": "Edit",
+	"upload": "Upload",
```
rosswilson2306, line 12: @ADSmith-0 This is a little annoying but you'll need to add the translations for these fields in the other `messages/*.json` files so that the language switcher works. We just usually use claude or something similar to translate the text then update the other files.

ui/packages/comhairle/src/lib/components/Media/MediaUpload.svelte
```diff
@@ -0,0 +1,95 @@
...
+				if (urls.length > 0) {
+					notifications.send({
+						message: `${urls.length} files uploaded`,
+						priority: 'SUCCESS'
+					});
+				}
+				oncomplete?.(urls);
+				return;
+			}
+			uploadForm?.submit();
```
rosswilson2306: This might be a personal thing but I kind of feel that if you have more than a couple of lines in a handler it is nicer to move it out of the html and into the `<script>`, so that we have separation of concerns for logic and markup. I'm happy if you or others disagree though.

We don't have anything in the styleguides about this (they are pretty light) so maybe it's worth discussing with others and making a decision. Or just ignore me!

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
```diff
@@ -0,0 +1,123 @@
...
+	import Button from '$lib/components/ui/button/button.svelte';
+	import { Images, SquareArrowOutUpLeft } from 'lucide-svelte';
+	import MediaLibrary from '$lib/components/Media/MediaLibrary.svelte';
+	import MediaItem from '$lib/components/Media/MediaItem.svelte';
+	import type { MediaDto } from '@crownshy/api-client/api';
+	import { tryCatchAsync, tryFetch, type FetchErr, type Result } from '$lib/utils/errorHandling';
+	import Spinner from '$lib/components/ui/spinner/spinner.svelte';
+
+	async function getData(): Promise<Result<'ok', MediaDto[], FetchErr>> {
+		const request = await tryFetch('/api/media');
```
rosswilson2306: If this is a GET request, is there any benefit to not using the api client? The client should do some error handling automatically and use typescript for the return types but maybe there's something I'm not seeing?

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
```diff
@@ -0,0 +1,123 @@
...
+		return { ok: media, err: null };
+	}
+
+	let dataRequest = $state.raw<ReturnType<typeof getData> | null>(null);
+
+	interface Props {
+		onconfirm: (url: string) => void;
+	}
+
+	const { onconfirm }: Props = $props();
```
rosswilson2306: Again, this might be a personal preference thing but I find it is nice (and consistent with the rest of the codebase) when the props destructuring is above any logic defined in the component. It just helps with knowing where to look for things quickly if it is consistent across the codebase.

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
```diff
@@ -0,0 +1,123 @@
+<script lang="ts" module>
+	let cache: MediaDto[] | null = null;
```
rosswilson2306, line 2: Does this behave differently to using `$state`?

rosswilson2306 (commented):
Left a few comments / questions.

api/src/routes/media/dto.rs
ADSmith-0: Oh yeah good catch! I think I accidentally left it in there from when we were debugging, just checked and it seems to all work fine without it, I'll remove it now!

ui/packages/comhairle/messages/en.json
ADSmith-0: Oh right yeah that makes sense, I think Daniela mentioned that when there aren't any translations it will default to English? So I was mainly putting them in to make sure we translated them at some point, I can do the translations in this ticket if we want though!

ui/packages/comhairle/src/lib/components/Media/MediaUpload.svelte
ADSmith-0: Yeah that's fair! I know what you mean, I generally prefer for event handlers to be where they're called, but I agree with you that's a really big function that kind of distracts from the flow of the html, I'll move that to the script area!

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
ADSmith-0: Yeah that's fair! I'll move it

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
ADSmith-0: Yeah so this is shared across every instance of the component, so if we have a page with a few different rich text boxes, then it will only make the api call once and then every component instance will have the cache available to them - https://svelte.dev/tutorial/svelte/sharing-code

rosswilson2306: Ah, nice!

ui/packages/comhairle/src/lib/components/RichTextEditor/UrlInputPopover/MediaLibraryDialog.svelte
ADSmith-0: Oh yeah I didn't see that we had that! I'll change that over!

rosswilson2306 (approved):

PR #585, opened 2026-07-07
[Media library] Update the UI

### LINKED ISSUE #568: [Media library] Update the UI
### Description 
Currently the UI for the media library doesn't look very good, we should update it to look better

### Inspiration
The Strapi media library is a good starting place

Image

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/568

- Checked styling

Screenshot from 2026-07-07 09-22-02

Screenshot from 2026-07-07 09-30-13

https://github.com/user-attachments/assets/69bf4f2e-4465-4d28-acad-ffbb1dfb879d


ui/packages/comhairle/src/lib/components/Media/MediaItem.svelte
```diff
@@ -1,37 +1,73 @@
 <script lang="ts">
+	import { capitalise } from '$lib/utils/string';
```
rosswilson2306, line 2: @ADSmith-0 We have another utils module for stuff like this https://github.com/crownshy/comhairle/blob/f916e17a6720347c86735550cb7aad3bb2b42245/ui/packages/comhairle/src/lib/utils/casingUtils.ts?plain=1#L1. Maybe they could be combined (I release this was from a previous PR).

rosswilson2306 (approved):
Left a comment but it's not really for something added in this PR so I've approved.

ui/packages/comhairle/src/lib/components/Media/MediaItem.svelte
ADSmith-0: Oh that's good to know! I'll move that in the next PR!

PR #588, opened 2026-07-07
584/conversation images

rosswilson2306:
Actions:

+ migration to replace image_url TEXT column with image UUID column on
  conversation table
+ model updates
+ MediaResolver and accompanying FromWithMedia trait to batch load media
  and resolve urls for DTOs
+ refactor admin configure page to use MediaLibraryDialog for image
  selection
+ refactor MediaLibraryDialog and usage to return media record in
  oncomplete handler
+ refactor media upload endpoint to allow multiple uploads
+ simplify Media interface upload functionality and usage
+ add upload UI for conversation images

Motivation:

+ allow conversations to reference images from media table instead of
  requiring urls from pre-existing hosts
+ simplify DB transactions and types for models with media references
+ allow selecting conversation image from existing media uploads
+ conversation image update requires access to media id instead of
+ allow multiple uploads in a single request for media instead of making
  multiple requests
+ allow uploading for conversation image


ui/packages/comhairle/src/lib/interfaces/Media.ts
```diff
@@ -9,19 +9,22 @@ type Opts = {
...
-			return {
-				ok: null,
-				err: {
-					id: 'MAX_SIZE_EXCEEDED',
-					message: `${file.name} exceeds max size ${opts.maxSizeMB}MB`
-				}
-			};
-		}
-
+	async #upload(to: string, files: File[], opts?: Opts): Promise<UploadReturn> {
```
ADSmith-0: I think we could probably just make this the `upload` function now? As I did the public and private distinction so that I could do the overloading for multiple files, it's not a massive difference though

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte
```diff
@@ -246,6 +253,30 @@
...
+
+			await invalidateAll();
+		} catch (e) {
+			console.error(e);
+			notifications.send({
+				message: 'Something went wrong updating conversation media',
+				priority: 'ERROR'
+			});
+		}
+	}
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte:286]: Could write this with the new `tryCatchAsync` if we wanted? Like:
```ts
async function updateConversationMedia(media: MediaDto, field: string) {
		const response = await tryCatchAsync(() => apiClient.UpdateConversation(
				{
					[field]: media.id
				},
				{ params: { conversation_id: conversation.id } }
			));
     
     if (response.err !== null) {
    	  console.error(response.err);
        notifications.send({
          message: 'Something went wrong updating conversation media',
          priority: 'ERROR'
        });
        return;
     }

			notifications.send({
				message: 'Successfully updated conversation media',
				priority: 'INFO'
			});
			await invalidateAll();
	}
```

I'll approve it either way because I don't want to force people to use it! It's just I find it easier to read when the error handling comes first, and also the TS compiler will flag if you don't handle the error, which I quite like, but it's all in all, it's personal preference!

ui/packages/comhairle/src/lib/components/Media/MediaLibraryDialog.svelte
ADSmith-0, line 1: Might be good to move this to the `lib` directory if we're using it in more than 1 place?

ui/packages/comhairle/src/lib/interfaces/Media.ts
rosswilson2306: Yeah, that makes sense. I'll update it.

ui/packages/comhairle/src/lib/components/Media/MediaLibraryDialog.svelte
rosswilson2306: Good shout, will do.

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte:286]: I'll have a look at that.

ADSmith-0 (approved):

stuartlynn (approved):
This looks good to me. Lets get it merged and tested out on stage

PR #593, opened 2026-07-08
333 ragflow chat empty response

rosswilson2306:
Actions:

+ ensure empty_response field carried through in conversion of comhairle create chat params to ragflow create chat params
+ tweak empty response default message

Motivation:

+ more helpful, friendly message for users

### DIFF
--- a/api/src/bot_service.rs
+++ b/api/src/bot_service.rs
@@ -23,8 +23,7 @@ If something in the background materials is unclear, you can ask me about it.
 
 You can ask me any question or ask me to explain something in simpler language.
 "#;
-pub const DEFAULT_CHAT_NOT_FOUND_RESPONSE: &str =
-    "Sorry I couldn't find an answer to that question.";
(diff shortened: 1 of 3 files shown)


api/src/bot_service.rs
```diff
@@ -23,8 +23,7 @@ If something in the background materials is unclear, you can ask me about it.
 
 You can ask me any question or ask me to explain something in simpler language.
 "#;
-pub const DEFAULT_CHAT_NOT_FOUND_RESPONSE: &str =
-    "Sorry I couldn't find an answer to that question.";
+pub const DEFAULT_CHAT_NOT_FOUND_RESPONSE: &str = "That one's outside what I can help with based on the available materials. If it's about the topic or the platform, try asking in a different way. I'm happy to keep trying!";
```
stuartlynn: I think we should change that to: "That one's outside what I can help with based on the  materials I have available. Your question is about the topic, try asking in a different way. I'm happy to keep trying!"

stuartlynn (changes requested):
one small language change and then this is good to go

PR #594, opened 2026-07-08
[Media library] User should be able to see and modify the details of a media file

### LINKED ISSUE #572: [Media library] User should be able to see and modify the details of a media file
### Description
Currently the user can't change anything once it's been uploaded

### Acceptance criteria
- [ ] Clicking on a media file in the media library should open a view which shows the file along with all of the metadata
- [ ] User should be able to edit the metadata for a given file
- [ ] User should be able to delete the file from this window

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/572

### Testing
- API tests
Screenshot from 2026-07-10 13-43-42
Screenshot from 2026-07-10 13-45-08

- Responsiveness

https://github.com/user-attachments/assets/1f9b4d92-ee0d-46c3-8603-c085be411c41

- Functionality

https://github.com/user-attachments/assets/d4301d66-24f2-4d23-815d-7e164849dc57


api/src/routes/media.rs
```diff
@@ -188,6 +205,17 @@ curl -X POST \\
                     .response::<201, Json<Vec<MediaDto>>>()
             }),
         )
+        .api_route(
+            "/{media_id}",
+            patch_with(update, |op| {
```
ADSmith-0, line 216: Using `PATCH` instead of `PUT` as I think that's more common for updating resources? 

"`PATCH` serves as a set of instructions for modifying a resource, whereas `PUT` represents a complete replacement of the resource" from [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PATCH), so I'm not sure which one we'd prefer?

api/src/models/media.rs
```diff
@@ -30,6 +27,8 @@ pub struct Media {
     /// Identifier in bulk_storage_service
     pub storage_key: String,
     pub filename: String,
+    // User defined identifier
```
rosswilson2306: @ADSmith-0 Small thing, but I would use the `///` documentation comment so that this explainer comment appears on hover.

api/src/routes/media.rs
```diff
@@ -106,9 +109,11 @@ async fn upload(
             .upload_file(&storage_key, bytes, metadata)
             .await?;
 
+        // TODO: Change name to be able to be set when uploading instead of copied from filename
         let create_media = CreateMedia {
             store_name: bulk_storage_config.store_name.to_string(),
             storage_key,
+            name: filename.clone(),
```
rosswilson2306: @ADSmith-0 I would have thought this would have been the filename minus the extension rather than just a straight copy of the filename? Maybe I misunderstood though.

api/src/routes/media.rs
rosswilson2306: I think that makes sense. I think we've went with PUT because there were a lot of places initially where we would have a form that would just take pretty much all the fields in say a conversation. We've probably changed a lot of those forms to be more targeted to specific fields instead of the entire data object but we've just stuck with PUT. I don't really have any preference and as we're using the generated apiClient it's not as if you have to remember which method to use on the frontend.

rosswilson2306 (dismissed):

api/src/models/media.rs
ADSmith-0: Oh that's really cool I didn't know that was a thing!

api/src/routes/media.rs
ADSmith-0: Yeah that might be a better idea tbf, I was setting it as the same by default just because it was easier and I thought if the user wants to change it then they can just get rid of the file extension at the end, but it probably makes more sense to do it that way, I can change it over!

ui/packages/comhairle/src/lib/utils/errorHandling.ts
```diff
@@ -3,7 +3,7 @@ type Ok<K extends string, T> = Record<K, T> & { err: null };
 type Err<K extends string, E extends ErrorType> = Record<K, null> & { err: E };
 export type Result<K extends string, T, E extends ErrorType> = Ok<K, T> | Err<K, E>;
 
-export async function tryCatchAsync<T, E extends string>(
+export async function tryCatchAsync<E extends string, T>(
```
ADSmith-0, line 6: Just realised these were the wrong way around and wanted to add this in here while I remembered, I ran the svelte checker and didn't see any issues coming from it

ui/packages/comhairle/src/routes/(admin)/admin/media-library/DetailsDialog.svelte
```diff
@@ -0,0 +1,181 @@
...
+				message: `Failed to update ${filename}. Please try again.`,
+				priority: 'ERROR'
+			});
+			return;
+		}
+		notifications.send({
+			message: `Updated ${filename} successfully`,
+			priority: 'SUCCESS'
+		});
+		editable = false;
```
rosswilson2306, line 84: @ADSmith-0 This looks like `editable` won't be set back to `false` if an error occurs because of the early return. Is that the desired behaviour?

rosswilson2306 (commented):
Just left one question.

ui/packages/comhairle/src/routes/(admin)/admin/media-library/DetailsDialog.svelte
ADSmith-0: Yeah I was thinking that if the error comes back then keeping them in edit mode means they can try and submit another request, but I'm happy to change it!

rosswilson2306: Nah, that's cool. Was just making sure it was deliberate.

rosswilson2306 (approved):

PR #615, opened 2026-07-10
613/Polis admin: Moderation tab (PR 2/4)

dtodorowa:
First UI PR in the effort to replace the Polis admin iframe. Adds the native **Moderation** tab where moderators sync, seed, and accept/reject statements. Builds on the endpoints from PR 0.

## What changed

- **New `PolisModeration.svelte`**: sync from Polis, add seed statements (single or CSV import), accept/reject with optimistic updates, and status filter chips (all / seeded / accepted / pending / rejected) with live counts.
- **Retabbed the Polis step editor** to `Configure / Setup / Moderation / Insights`. Dropped the dead `Statements`, `Participants`, and `Moderate` tabs. Insights is a placeholder for PR 2.
- **`+page.ts`** now loads `statement_aux` for Polis steps under `depends('polis:statement-aux')`, so Moderation and the coming Insights share one fetch and mutations invalidate just that key.

## Decisions and assumptions

- Ported from the civic_os moderation page, with two changes: seeds go through the new server-side `PolisPostSeed` (no browser-side Polis auth or CORS), and colours use theme tokens only (no hardcoded colours, which we need to resolve in civic os). "Accepted" green maps to `--primary` since there is no `--success` token.
- Draft conversations show only moderator seeds (preview poll); once launched, the same tab shows participant statements (live poll). Same UI, driven by the existing `isLive` split.
- Statements only appear after a sync. Aux rows are synced from Polis, not created live.

## Testing
- Manual testing -> create a conversation with a polis step, add some statements, test if moderation and seeding works
- Not yet driven end to end (needs a running frontend plus a Polis step with synced statements).

image
image

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
+++ b/ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
@@ -75,10 +75,10 @@
 		const text = draftText.trim();
 		if (!text || addingSeed) return;
 		addingSeed = true;
-		const { err } = await tryCatchAsync(() => postSeeds([text]));
+		const result = await tryCatchAsync(() => postSeeds([text]));
 		addingSeed = false;
-		if (err) {
-			console.error('PolisPostSeed failed', err);
+		if (result.err !== null) {
+			console.error('PolisPostSeed failed', result.err);
 			notifications.send({ priority: 'ERROR', message: 'Failed to add statement' });
 			return;
 		}
@@ -93,7 +93,7 @@
 		if (!file || csvImporting) return;
 		csvImporting = true;
 
-		const { ok: count, err } = await tryCatchAsync(async () => {
+		const result = await tryCatchAsync(async () => {
 			// One statement per line; strip wrapping quotes and a leading header row.
 			const lines = (await file.text())
 				.split(/\r?\n/)
@@ -108,15 +108,15 @@
 		csvImporting = false;
 		input.value = '';
 
-		if (err) {
-			console.error('CSV import failed', err);
+		if (result.err !== null) {
+			console.error('CSV import failed', result.err);
 			notifications.send({ priority: 'ERROR', message: 'CSV import failed' });
 			return;
 		}
-		if (count) {
+		if (result.ok) {
 			notifications.send({
 				priority: 'INFO',
-				message: `Imported ${count} statement${count === 1 ? '' : 's'}`
+				message: `Imported ${result.ok} statement${result.ok === 1 ? '' : 's'}`
 			});
 		}
 		showAddForm = false;
--- a/ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
+++ b/ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
@@ -1,0 +1,341 @@
+<script lang="ts">
+	import { invalidate } from '$app/navigation';
+	import { Button, LoadingButton } from '$lib/components/ui/button';
+	import { Textarea } from '$lib/components/ui/textarea';
+	import * as Card from '$lib/components/ui/card';
+	import * as Dialog from '$lib/components/ui/dialog';
+	import { notifications } from '$lib/notifications.svelte';
+	import { apiClient } from '@crownshy/api-client/client';
+	import type { PolisStatementAux } from '@crownshy/api-client/api';
+	import { Check, X, Plus, Upload, RefreshCw } from '@lucide/svelte';
+
+	let {
+		workflowStepId,
+		statements: initialStatements
+	}: {
+		workflowStepId: string;
+		statements: PolisStatementAux[];
+	} = $props();
+
+	// Local mutable copy so optimistic accept/reject re-renders without a refetch.
+	// Reset whenever the load re-runs (sync/seed invalidate `polis:statement-aux`).
+	let statements = $state<PolisStatementAux[]>(initialStatements);
+	$effect(() => {
+		statements = initialStatements;
+	});
+
+	// --- Sync from Polis ---
+	// Participant submissions only appear here after a sync — aux rows are synced
+	// from Polis, not created live. Existing moderation/themes are preserved.
+	let syncing = $state(false);
+
+	async function syncFromPolis() {
+		if (syncing) return;
+		syncing = true;
+		try {
+			const res = await apiClient.PolisSyncStatementAux({ workflow_step_id: workflowStepId });
+			notifications.send({
+				priority: 'INFO',
+				message: `Synced ${res.synced} statement${res.synced === 1 ? '' : 's'} from Polis${
+					res.skipped_invalid_xid ? ` (${res.skipped_invalid_xid} skipped)` : ''
+				}`
+			});
+			await invalidate('polis:statement-aux');
+		} catch (e) {
+			console.error('PolisSyncStatementAux failed', e);
+			notifications.send({
+				priority: 'ERROR',
+				message: 'Failed to sync statements from Polis'
+			});
+		} finally {
+			syncing = false;
+		}
+	}
+
+	// --- Add seed statements (moderator authoring) ---
+	// Seeds are posted server-side to the active poll via PolisPostSeed (no
+	// browser-side Polis auth / CORS), then we re-sync so the new comment comes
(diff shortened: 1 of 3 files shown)


ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
```diff
@@ -0,0 +1,341 @@
...
+	// browser-side Polis auth / CORS), then we re-sync so the new comment comes
+	// back with its real Polis-issued ids.
+	let showAddForm = $state(false);
+	let draftText = $state('');
+	let addingSeed = $state(false);
+	let csvImporting = $state(false);
+	let fileInput = $state<HTMLInputElement>();
+
+	async function postSeeds(texts: string[]) {
+		for (const statement_text of texts) {
```
rosswilson2306, line 67: This should probably be in a try / catch block or use the new tryCatchAsync that Adam made.

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
```diff
@@ -0,0 +1,341 @@
...
+	// Track in-flight requests per aux row so the buttons can disable mid-call.
+	let pending = $state<Record<string, boolean>>({});
+
+	async function setStatus(row: PolisStatementAux, status: 'accepted' | 'rejected') {
+		if (pending[row.id] || row.moderation_status === status) return;
+		const decision = status === 'accepted' ? 'accept' : 'reject';
+		pending = { ...pending, [row.id]: true };
+
+		// Optimistic update; roll back on failure.
+		const prevStatus = row.moderation_status;
```
rosswilson2306, line 183: Just curious what the benefit was of doing this? Does the request take a long time or something?

rosswilson2306 (commented):
Just a small comment about a try / catch

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
dtodorowa: the accept/reject request is a network round-trip, so without the optimistic update the accent bar and disabled state don't change until it resolves, which I think feels kinda laggy when moderating a long list, we flip the status locally right away and roll back to prevStatus if the request fails. If you think it's an overkill, I'm happy to to strip it down? wdyt?

rosswilson2306: Nah, I was just curious what was the reason for it. Nothing wrong with a bit of optimistic UI.

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
dtodorowa: postSeeds is called by importCsv and addSeed and they wrap the call in a try catch already, but will try out the `tryCatchAsync` Adam made 😇

rosswilson2306: Ah, I missed that. Probably just personal choice then.

rosswilson2306 (dismissed):

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
```diff
@@ -0,0 +1,345 @@
...
+		notifications.send({ priority: 'INFO', message: 'Seed statement added' });
+	}
+
+	async function importCsv(e: Event) {
+		const input = e.currentTarget as HTMLInputElement;
+		const file = input.files?.[0];
+		if (!file || csvImporting) return;
+		csvImporting = true;
+
+		const { ok: count, err } = await tryCatchAsync(async () => {
```
ADSmith-0: @dtodorowa

Hey Daniela!

Just to give a bit more context to this, I'm not sure that this would work fully unfortunately because of the way that TS handles discriminated unions, basically the `Result` type is either

`{ ok: T, err: null }`
or
`{ ok: null, err: string }`

So if you say
```ts
const result = await tryCatchAsync(() => { ... });

if (result.err !== null) {
  return;
}

result.ok; // type T (we know it's not `null` now)
```

But with how TS does discriminated unions, I think if you destructure them then TS sees them as separate variables at that point, so

```ts
const { ok, err } = await tryCatchAsync(() => { ... });

if (err !== null) {
  return;
}

ok; // TS thinks this could be `T` or `null`
```

So yeah the way I generally use it is like
```ts
const count = await tryCatchAsync(...);

if (count.err !== null) {
		console.error('CSV import failed', err);
		notifications.send({ priority: 'ERROR', message: 'CSV import failed' });
		return;
}
// We know it must have succeeded from here on out because we're returning when the error isn't null
notifications.send({
		priority: 'INFO',
		message: `Imported ${count} statement${count.ok === 1 ? '' : 's'}`
});
```

I basically saw how GO does it and thought yeah that's much nicer haha

Ideally it should allow you to wrap errors just around the bit that could possibly error and then it makes sure that you handle it, rather than having to put everything inside a `try...catch` because of scoping rules, so like here you could maybe do it like

```ts
const lines = await tryCatchAsync(() => file.text()); // Bit that can error

if (lines.err !== null || lines.ok.length === 0) {
  notifications.send({ priority: 'ERROR', message: 'CSV import failed' });
  return;
}

// ...rest of the counting logic using lines.ok

const postSeedsResult = await tryCatchAsync(() => postSeeds(lines.ok)); // If this can error?

if (postSeedsResult.err !== null) {
  // Handle error
  return;
}

// ...more logic
```

But yeah like I said to Ross I don't want to force people to use it or anything! I just find it so much easier to work with, and especially if you have multiple `try catch` calls in one block I think it makes it way easier to read

I hope that all makes sense, please let me know if not! I'm always happy to jump on a call 😄️

--- 
Btw another bonus is that it lets you type errors

```ts
const count = await tryCatchAsync<"INCORRECT_FILE_TYPE" | "NO_VALUES_FOUND">(...);

if (count.err !== null) {
	 console.error('CSV import failed', err);
   switch (count.err) { // Type-safe, you get autocomplete with it
   case "INCORRECT_FILE_TYPE":
        notifications.send({ priority: 'ERROR', message: 'Incorrect file type. Only CSV files are allowed' });
   case "NO_VALUES_FOUND":
        notifications.send({ priority: 'ERROR', message: 'Incorrect format. Please try a different file' });
   }
		return;
}
// We know it must have succeeded from here on out because we're returning when the error isn't null
notifications.send({
		priority: 'INFO',
		message: `Imported ${count} statement${count.ok === 1 ? '' : 's'}`
});
```

dtodorowa: th for taking the time to write all this 💛
I pushed some changes, let me know if it looks good 😇

ADSmith-0: No worries! 😄️

Yeah it's looking really good! 😄️

I think there were a few other places in the file where there are `try...catch`s in case we wanted to change them over, but it looks good either way! 🙂️

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
```diff
@@ -0,0 +1,366 @@
...
+		accepted: statements.filter((s) => s.moderation_status === 'accepted').length,
+		pending: statements.filter((s) => s.moderation_status === 'pending').length,
+		rejected: statements.filter((s) => s.moderation_status === 'rejected').length
+	});
+
+	const visible = $derived.by(() => {
+		let list = statements;
+		if (filter === 'seeded') list = list.filter((s) => s.is_seed);
+		else if (filter !== 'all') list = list.filter((s) => s.moderation_status === filter);
+		return [...list].sort((a, b) => b.polis_statement_id - a.polis_statement_id);
```
ADSmith-0, line 162: Just for future I think there's now a `toSorted` array function now that will sort and then return a new array 🙂

ADSmith-0 (approved):

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
dtodorowa: sweet! adding a follow-up!

PR #618, opened 2026-07-10
613/Polis admin: Insights tab (PR 3/4)

dtodorowa:
Second UI PR. Adds the native **Insights** tab: the per-step Polis report showing where opinions cluster, what people agree and disagree on, and how statements group by theme. Uses the typed `report_data` endpoint from PR 0.

## What changed

- **Analytics core** (`report.ts`, `reportTypes.ts`): ported from civic_os. Consensus (all groups agree >80% or <20%), difference (group agree gap >30pts), theme controversy (low <15, moderate 15–30, high >30), low data quality (any group <10 votes). Pure functions, thresholds as tunable constants.
- **`reportTypes.ts`**: re-exports the wire types from `@crownshy/api-client` instead of re-declaring them. Only adds what the client can't know (see Decisions).
- **`PolisInsights.svelte`**: stats row, Themes card, three sections (Consensus, Difference, All Statements). Theme filter chips with `?theme=` deep-linking, per-column filters (include hosts, include passes), low-quality reveal, CSV export, per-row theme picker.
- **9 reskinned components** in `lib/components/polis-report/`: StatementRow, StatementSection, ThemeBar, ThemeChip, GroupCircle, ColumnFilterHeader, FilterToggle, RowAccentStripe, ThemePicker.
- **`+page.ts`**: loads `report_data` for Polis steps under a `polis:report` depends key. Returns null when the poll has no votes yet. Assigns the client's return type directly, no cast.

## Decisions

Need help with this: **Report types reuse the typed client, they don't duplicate it.** `reportTypes.ts` re-exports `WikiPollReport` and its friends from `@crownshy/api-client` under the names this feature uses, and only adds:
- `topics`/`subtopics`, a client-side overlay from `polis_statement_aux.themes` (the backend payload has neither).
- view models computed here: `GroupVotePercent`, `ThemeControversy`, `ThemeSummary`.

Why this matters: the first cut hand-copied the wire shapes as parallel interfaces. That forced an `as unknown as PolisReportData` cast in the loader (the two type trees weren't the same identity), and the copy had already drifted from the source (it wrongly made `group_informed_consensus`/`divisiveness` required and `is_seed` optional). Deriving from the client removes the drift and drops the cast.

One gotcha, documented in-file: `PolisReportData` lists its three fields explicitly instead of `Omit<WikiPollReport, 'comments'>`, because the client's zod `.passthrough()` adds a `[k: string]: unknown` index signature that makes `Omit` collapse the other props to `unknown`.

(A second, older copy of these types lives at `lib/types/report.ts` for the public `waves/report` feature. It predates this branch. Worth unifying the same way later, kept out of scope here.)

**Other:**
- Themes are human authored and live in `polis_statement_aux.themes`. The view overlays them onto a cloned copy, so loaded `reportData` is never mutated. Theme edits go through the api client, then invalidate `polis:statement-aux`.
- Theme tokens only, no hardcoded colours. No success/warning token exists, so consensus green maps to `--primary` and low-quality amber maps to `text-muted-foreground`. GroupCircle agree scale: `text-destructive` (<33), `text-muted-foreground` (33–66), `text-primary` (>66).
- Uncertainty section deferred. CSV helper, stat row, and meter bar inlined since comhairle has no equivalents.

## Testing

- Not yet driven end to end (needs a launched poll with votes across 2+ groups).
- `svelte-check` clean on the touched files. The only remaining errors nearby are pre-existing `step`-undefined warnings in a sibling `+page.svelte` this PR doesn't touch.

image
image

### DIFF
--- a/ui/packages/comhairle/src/lib/components/polis-report/ColumnFilterHeader.svelte
+++ b/ui/packages/comhairle/src/lib/components/polis-report/ColumnFilterHeader.svelte
@@ -1,0 +1,48 @@
+<script lang="ts">
+	import { ChevronDown, Check } from '@lucide/svelte';
+	import * as Popover from '$lib/components/ui/popover';
+
+	interface Props {
+		/** Column header label, e.g. "Author". */
+		label: string;
+		/** Popover option label, phrased as the positive action, e.g. "Include all statements". */
+		optionLabel: string;
+		/** Controlled — checked = the option is INCLUDED (shown). */
+		checked?: boolean;
+		align?: 'left' | 'center' | 'right';
+		onchange?: (next: boolean) => void;
+	}
+
+	let { label, optionLabel, checked = false, align = 'left', onchange }: Props = $props();
+
+	const justify = {
+		left: 'justify-start',
+		center: 'justify-center',
+		right: 'justify-end'
+	} as const;
+
+	function toggle() {
+		onchange?.(!checked);
+	}
+</script>
+
+<Popover.Root>
+	<Popover.Trigger
+		class={`text-foreground hover:text-foreground/70 flex w-full cursor-pointer items-center gap-1 text-sm font-semibold whitespace-nowrap uppercase transition-colors ${justify[align]}`}
+	>
+		{label}
+		<ChevronDown class="size-3 shrink-0" />
+	</Popover.Trigger>
+	<Popover.Content align="start" class="w-72 overflow-hidden rounded-[10px] p-0 shadow-md">
+		<button
+			type="button"
+			onclick={toggle}
+			class="hover:bg-muted/50 border-border flex w-full cursor-pointer items-center justify-between gap-3 border-b px-4 py-3 text-left transition-colors"
+		>
+			<span class="text-foreground text-base font-medium whitespace-nowrap"
+				>{optionLabel}</span
+			>
+			{#if checked}<Check class="text-primary size-5 shrink-0" />{/if}
+		</button>
+	</Popover.Content>
+</Popover.Root>
--- a/ui/packages/comhairle/src/lib/components/polis-report/ColumnFilterHeader.svelte
+++ b/ui/packages/comhairle/src/lib/components/polis-report/ColumnFilterHeader.svelte
@@ -10,7 +10,7 @@
 		/** Controlled — checked = the option is INCLUDED (shown). */
 		checked?: boolean;
 		align?: 'left' | 'center' | 'right';
-		onchange?: (next: boolean) => void;
+		onchange: (next: boolean) => void;
 	}
 
 	let { label, optionLabel, checked = false, align = 'left', onchange }: Props = $props();
@@ -22,7 +22,7 @@
 	} as const;
 
 	function toggle() {
-		onchange?.(!checked);
+		onchange(!checked);
 	}
 </script>
 
--- a/ui/packages/comhairle/src/lib/components/polis-report/StatementSection.svelte
+++ b/ui/packages/comhairle/src/lib/components/polis-report/StatementSection.svelte
@@ -1,0 +1,186 @@
+<script lang="ts">
+	import type { Snippet } from 'svelte';
+	import { ChevronDown } from '@lucide/svelte';
+	import ColumnFilterHeader from './ColumnFilterHeader.svelte';
+	import * as Card from '$lib/components/ui/card';
+
+	interface Props {
+		/** Optional id on the root section (scroll anchor for deep-links). */
+		id?: string;
+		title: string;
+		/** Trailing description text after the "N STATEMENTS" count, e.g. "with greater than 80%…". */
+		description: string;
+		/** Count shown as the coloured "N STATEMENTS" lead-in (main + low-quality). */
+		count?: number;
+		/** Accent colour for the count lead-in. */
+		countAccent?: 'consensus' | 'difference' | 'all';
+		/** Header label for the per-variant metric column, e.g. "Min Agree". */
+		metricLabel: string;
+		/** Number of opinion groups — sizes the trailing agree-rings column so the
+		    header grid and the (separate) row grids share identical column widths. */
+		groupCount?: number;
+		/** Full main-list count. When it exceeds what's rendered, show the expander. */
+		total?: number;
+		/** Rows shown while collapsed (parent does the slicing). */
+		collapsedCount?: number;
+		/** Bindable expand state — parent slices its list on this. */
+		expanded?: boolean;
+		/** How many low-quality rows are hidden. 0 → no low-quality reveal. */
+		lowQualityCount?: number;
+		/** Bindable — reveal the low-quality rows. */
+		showLowQuality?: boolean;
+		excludePasses?: boolean;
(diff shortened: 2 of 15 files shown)


ui/packages/comhairle/src/lib/components/polis-report/ColumnFilterHeader.svelte
```diff
@@ -0,0 +1,48 @@
...
+	let { label, optionLabel, checked = false, align = 'left', onchange }: Props = $props();
+
+	const justify = {
+		left: 'justify-start',
+		center: 'justify-center',
+		right: 'justify-end'
+	} as const;
+
+	function toggle() {
+		onchange?.(!checked);
```
rosswilson2306: @dtodorowa The button that calls this handler will do nothing if `onchange` is undefined. Would it not be better just making that prop required or is there a reason it needs to be optional?

ui/packages/comhairle/src/lib/components/polis-report/StatementSection.svelte
```diff
@@ -0,0 +1,186 @@
...
+		excludePasses?: boolean;
+		excludeHosts?: boolean;
+		onExcludePassesChange?: (next: boolean) => void;
+		onExcludeHostsChange?: (next: boolean) => void;
+		children: Snippet;
+		/** Optional action rendered on the title row, right-aligned (e.g. a Download button). */
+		headerAction?: Snippet;
+		/** Optional controls rendered between the description and the table (e.g. theme chips). */
+		toolbar?: Snippet;
+		/** Low-quality rows, rendered below the reveal toggle when expanded. */
```
rosswilson2306, line 41: Not a big deal but if you changed these to `/***` they would appear as jsdoc comments on hover.

ui/packages/comhairle/src/lib/components/polis-report/ThemeChip.svelte
```diff
@@ -0,0 +1,37 @@
...
+	interface Props {
+		label: string;
+		selected: boolean;
+		/**
+		 * Colour scheme — both variants use the comhairle primary accent. `default`
+		 * renders primary text on a neutral muted surface; `primary` uses the solid
+		 * primary fill when selected and the subtle secondary surface when idle.
+		 */
+		variant?: 'default' | 'primary';
+		onclick?: () => void;
```
rosswilson2306: Same comment here, the button does nothing if the `onclick` prop is undefined. Could it be required instead of optional? Seems unusual to have a button without a click handler.

ui/packages/comhairle/src/lib/components/polis-report/ThemePicker.svelte
```diff
@@ -0,0 +1,172 @@
...
+
+	/**
+	 * Dropdown uses `position: fixed` to escape ancestor `overflow: hidden`
+	 * (e.g. the Card wrapping the statements list). So we have to position it
+	 * manually from the trigger's bbox, and re-position on scroll/resize.
+	 *
+	 * When there isn't room below the trigger (statement near the bottom of the
+	 * viewport), flip the menu above it so it never gets clipped off-screen.
+	 */
+	function reposition() {
```
rosswilson2306: This seems like a lot of work for a css workaround. Does the ancestor component definitely need to have `overflow: hidden`?

ui/packages/comhairle/src/lib/tools/polis/PolisInsights.svelte
```diff
@@ -0,0 +1,608 @@
...
+		for (const a of statementAux) map[a.polis_statement_id] = a;
+		auxByTid = map;
+	});
+
+	/**
+	 * Overlay aux.themes onto each comment so the existing utils (which read
+	 * `comment.topics`) get aux-sourced themes for free. Themes are human-authored
+	 * today and live only in aux; Polis carries none, so `comment.topics` is empty
+	 * until a future source populates it. Comments with no aux row keep that empty
+	 * `topics` and simply contribute no themes. Works on a cloned copy so the
```
rosswilson2306, line 62: Just out of curiousity, what is `aux` in this context? I keep seeing references to it and have no idea what it means!

ui/packages/comhairle/src/lib/tools/polis/report.ts
```diff
@@ -0,0 +1,248 @@
...
+	comment: ReportComment,
+	groups: ReportGroup[],
+	{ excludePasses = false }: { excludePasses?: boolean } = {}
+): GroupVotePercent[] {
+	return comment.group_votes.map((gv) => {
+		const group = groups.find((g) => g.group_id === gv.group_id);
+		const totalMembers = group ? group.total_members : gv.agrees + gv.disagrees + gv.passes;
+		const totalVoted = gv.agrees + gv.disagrees + gv.passes;
+		const denom = excludePasses ? gv.agrees + gv.disagrees : totalVoted;
+		const pct = (n: number) => (denom > 0 ? (n / denom) * 100 : 0);
```
rosswilson2306: What does `pct` stand for?

ui/packages/comhairle/src/lib/tools/polis/PolisInsights.svelte
dtodorowa: stands for "auxiliary" (I think!) in `PolisStatementAux` since it's like a supplementary metadata table holding themes, status, reason, step id, etc etc

Might be worth adding a little comment  like
 `// "aux" = PolisStatementAux, our supplementary per-statement record (themes, moderation, seed) that Polis doesn't store`
 
 wdyt?

ui/packages/comhairle/src/lib/tools/polis/report.ts
dtodorowa: ohh pct is just "percent". `pct(gv.agrees)` gives the share of agrees.
But I think yeh, it's not clear. gonna rename. and add info, sorry abbreviations keep coming! Experimenting with LLMs and will add it to the context. Forgot to do it this time

ui/packages/comhairle/src/lib/components/polis-report/StatementSection.svelte
dtodorowa: wait, i thought two stars already does this? What does three stars do?

rosswilson2306: Ah yeah! Sorry, I'm getting confused with Rust comments!

ui/packages/comhairle/src/lib/tools/polis/PolisInsights.svelte
rosswilson2306: Nah, I was just curious. I figured it meant "auxiliary" but just wasn't sure what that referred to. I wasn't sure if it was something coming from Polis. Didn't realise it was just additional data.

rosswilson2306 (approved):

ui/packages/comhairle/src/lib/components/polis-report/StatementSection.svelte
dtodorowa: lmaoo i feel ya

PR #621, opened 2026-07-10
613/Polis admin: native Setup, remove the iframe (PR 4/4)

dtodorowa:
Final PR of the arc. Replaces the embedded Polis admin iframe with native controls, so configuring a Polis conversation matches the rest of the admin and no Polis login happens in the browser.

## What changed

- **Removed the iframe** and the `POLIS_LOGIN` postMessage from `PolisManage.svelte` (and the duplicate Sync button, since Moderation owns syncing).
- **Native Setup form:** Content (Topic, Description) and Settings (Required votes, Show remaining statements, No comments without moderator approval, Label seed statements as Conversation Starter). Seed authoring points to the Moderation tab.
- **Backend:** extended `PolisToolConfig` with `topic`, `description`, `is_active`, `strict_moderation`, and `label_seeds_as_conversation_starter` (all default, so existing configs still load). Regenerated the client.

## Decisions and assumptions

- **Mirror in `tool_config`:** Polis has no read path for topic/description/strict_moderation, and `PolisToolConfig` drops unknown fields, so the form cannot pre-fill from Polis. Config fields are written to Polis first (`PolisUpdateConfig`), then mirrored into `tool_config`, which the form reads back. Matches how `required_votes` already works. Trade-off: `tool_config` can drift from Polis if a write half-fails or Polis is changed out of band. (Maybe there is a better way to do this? Let me know if you can think of anything.)
- **`is_active` (open/closed) is plumbed but not surfaced.** Participant access in comhairle is governed by conversation launch and the Workflow, so a per-Polis on/off toggle is redundant. The field stays on the config in case we need it, because maybe Stuart has to confirm.

## Testing

- `svelte-check` clean on `PolisManage.svelte` (only the benign `state_referenced_locally` warnings). `cargo check` (lib and tests) clean. Client regenerated.
- Not driven end to end. Writes go through the two typed endpoints; behavior verified by types plus the port of the existing update pattern.

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
+++ b/ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
@@ -1,11 +1,10 @@
 <script lang="ts">
 	import { invalidateAll } from '$app/navigation';
-	import { LoadingButton } from '$lib/components/ui/button';
 	import Input from '$lib/components/ui/input/input.svelte';
 	import Label from '$lib/components/ui/label/label.svelte';
 	import Switch from '$lib/components/ui/switch/switch.svelte';
+	import { Textarea } from '$lib/components/ui/textarea';
 	import { notifications } from '$lib/notifications.svelte';
-	import { camelToSentenceCase, camelToSnakeCase } from '$lib/utils/casingUtils';
 	import { apiClient } from '@crownshy/api-client/client';
 	import { useDebounce } from 'runed';
 
@@ -16,7 +15,7 @@
 		workflowStepId,
 		isLive
 	}: {
-		toolConfig: any; // TODO:
+		toolConfig: any; // TODO: type once tool config types are generated
 		conversationId: string;
 		workflowId: string;
 		workflowStepId: string;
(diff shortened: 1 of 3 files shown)


ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
```diff
@@ -16,54 +17,60 @@
...
+	let descriptionInput = $state(description);
 	$effect(() => {
 		requiredVotesInput = requiredVotes;
 	});
+	$effect(() => {
+		topicInput = topic;
+	});
+	$effect(() => {
+		descriptionInput = description;
+	});
```
ADSmith-0, line 48: I think these could be `$derived()` potentially?

ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
```diff
@@ -16,54 +17,60 @@
...
-
-	const debouncedUpdateToolConfig = useDebounce(async (e: Event, field: string) => {
-		const raw = (e.target as HTMLInputElement).value;
-		const value = Number(raw);
-		// Don't persist an empty, invalid, or non-positive number
-		if (raw.trim() === '' || !Number.isFinite(value) || value < 1) return;
-
+	// Config keys that must be proxied to Polis (the rest are comhairle-only
+	// display flags that live in tool_config).
+	const POLIS_KEYS = new Set(['topic', 'description', 'is_active', 'strict_moderation']);
```
ADSmith-0, line 52: Do these need to be constants outside of this file? Just wondering if we use them anywhere else

ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
```diff
@@ -74,157 +81,174 @@
...
-			syncing = false;
-		}
-	}
+	const saveTopic = useDebounce((v: string) => saveField('topic', v), 500);
+	const saveDescription = useDebounce((v: string) => saveField('description', v), 500);
+	const saveRequiredVotes = useDebounce((raw: string) => {
+		const value = Number(raw);
+		// Don't persist an empty, invalid, or non-positive number.
+		if (raw.trim() === '' || !Number.isFinite(value) || value < 1) return;
+		saveField('required_votes', value);
```
ADSmith-0, line 97: I think you can do `parseInt()` here instead of `Number()` and then `Number.isNan(parseInt(Infinity))` and `Number.isNan(parseInt(""))` both are false, but I think either way is fine 🙂

ADSmith-0 (commented):
Not sure if I have the full context for this ticket, but tried to give a review anyway 🙂

Just a few small things, but happy to approve as-is!

ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
dtodorowa: nah not used, not yet, but the moment we do, I'd move out

ui/packages/comhairle/src/lib/tools/polis/PolisManage.svelte
dtodorowa: I think Number is the stricter validator. 
`Number("12abc") → NaN` (rejected), whereas `parseInt("12abc") → 12` (silently accepts garbage)

Something that I'm thinking about now is that neither floors decimals (oops) so "2.5" still passes for `required_votes` that's a separate issue, tho, will add a ticket to resolve 🙈

ADSmith-0: Oh wow I never knew that!

Would be nice for JS to just pick 1 function for parsing ints and sticking with that one :sweat_smile:

ADSmith-0 (approved):

PR #623, opened 2026-07-13
613/polis moderation search bulk (5/4)

dtodorowa:
TODO: add description 

image
image
image


ui/packages/comhairle/src/lib/tools/polis/polis-moderation/AddSeedStatementsDialog.svelte
ADSmith-0, line 1: It looks like this was copied over from `PolisModeration.svelte`, but I think that there were changes from #615 that might not have been pulled over

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/StatementModerationRow.svelte
```diff
@@ -0,0 +1,97 @@
...
+		onModerate
+	}: {
+		row: PolisStatementAux;
+		selected: boolean;
+		// This row has an accept/reject request in flight.
+		pending: boolean;
+		// A bulk moderation is running for the whole table.
+		bulkWorking: boolean;
+		onToggle: (checked: boolean) => void;
+		onModerate: (status: 'accepted' | 'rejected') => void;
```
ADSmith-0: I'm not sure that we have a unified style across the codebase, but it looks like we extract them `Props` type in quite a few places like:
```ts
type Props = {
   ...
}
// or
interface Props {
  ...
}

const { ... }: Props = $props();
```

Just wanted to point it out here, as I think it might be a bit more readable? Might just be my personal preference though!

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/AddSeedStatementsDialog.svelte
```diff
@@ -0,0 +1,143 @@
...
+	import { Plus, Upload } from '@lucide/svelte';
+
+	let {
+		workflowStepId,
+		onSeeded
+	}: {
+		workflowStepId: string;
+		// Called after seeds are posted + synced so the parent can refresh its data.
+		onSeeded: () => void | Promise<void>;
+	} = $props();
```
ADSmith-0: Same here as https://github.com/crownshy/comhairle/pull/623#discussion_r3571146695

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/StatementsTable.svelte
```diff
@@ -0,0 +1,128 @@
...
+		selected: Record<string, boolean>;
+		pending: Record<string, boolean>;
+		// Which bulk action is in flight, or null when idle.
+		bulkAction: 'accepted' | 'rejected' | null;
+		onToggleSelect: (id: string, checked: boolean) => void;
+		onToggleAll: (checked: boolean) => void;
+		onClear: () => void;
+		onBulkModerate: (status: 'accepted' | 'rejected') => void;
+		onModerate: (row: PolisStatementAux, status: 'accepted' | 'rejected') => void;
+	} = $props();
```
ADSmith-0: Same here as https://github.com/crownshy/comhairle/pull/623#discussion_r3571146695

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/StatementsTable.svelte
```diff
@@ -0,0 +1,128 @@
...
+					<LoadingButton
+						size="sm"
+						loading={bulkAction === 'accepted'}
+						disabled={bulkWorking}
+						onclick={() => onBulkModerate('accepted')}
+					>
+						<Check class="size-4" />
+						Approve
+					</LoadingButton>
+					<LoadingButton
```
ADSmith-0, line 69: For this and the `ModerationRow` it might be worth using the HTML table element? https://www.w3schools.com/html/html_tables.asp, I think it does stuff like automatic aligning with columns and provides better semantics, but we can make this something we look into later if we want!

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/AddSeedStatementsDialog.svelte
dtodorowa: ohh goood point .. :/ I tried to be helpful to reviewers by making smaller PRs but made it easier to make mistakes, oops

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/StatementModerationRow.svelte
dtodorowa: yehh, i agree with this. will try to change everything i touch to follow that. Would be nice to start drafting out a style guideline doc, bc you are making some really good suggestions

ui/packages/comhairle/src/lib/tools/polis/polis-moderation/StatementsTable.svelte
dtodorowa: good call to flag it! I'd lean toward deferring for now: right now each row isn't really tabular data, it's a `role="button"` click-to-select target with a status accent bar, and the header doubles as the sticky bulk-actions bar. A semantic `<table>` fights that interaction model more than it helps. Happy to revisit if these rows ever become read-only/tabular or maybe I am not understanding this fully, but I'd keep it out of this PR, so we can merge faster and test on staging

ADSmith-0: Yeah that's totally fair! :slightly_smiling_face:

ADSmith-0 (approved):

PR #624, opened 2026-07-13
613/ Add batch moderation endpoint for Polis statements with single-login optimization (6/4)

dtodorowa:
Actions:
+ add get_by_ids function to fetch multiple polis_statement_aux rows in one query
+ add update_moderation_status_many function to bulk-update moderation_status with RETURNING clause
+ add moderate_statement_aux_batch endpoint that logs in to Polis once and moderates multiple statements
+ add ModerateStatementAuxBatchRequest/Response types with per-row failure tracking
+ update PolisModeration.svelte to use batch endpoint


api/src/tools/polis.rs
```diff
@@ -737,6 +752,113 @@ async fn moderate_statement_aux(
...
+    // A selection always comes from a single moderation view (one workflow step,
+    // one poll), so we authorize and log in to Polis exactly once for the batch.
+    let workflow_step_id = rows[0].workflow_step_id;
+    if rows.iter().any(|r| r.workflow_step_id != workflow_step_id) {
+        return Err(ComhairleError::BadRequest(
+            "all statements must belong to the same workflow step".into(),
+        ));
+    }
+
+    polis_statement_aux::check_can_moderate(&state.db, &user, &workflow_step_id).await?;
```
ADSmith-0, line 810: Just wanted to check I'm not super familiar with Rust I know this function can return an error, I assume that just throws to the root of the stack right? Like you don't have to catch it here and then pass it up?

ADSmith-0 (dismissed):
Not sure I'm the best one to review the Rust stuff? But from everything I can see, it all looks good to me!

The `tryCatchAsync` stuff is all really good too! 🙂

api/src/tools/polis.rs
dtodorowa: I'm not super familiar either but i think that's right. The `?` means if `check_can_moderate` returns an `Err`, it returns early from this function with that error, so we don't have to catch it and pass it up manually. It doesn't go to the "root" exactly, it propagates to our caller, which here returns `Result<..., ComhairleError>`, and that implements axum's `IntoResponse`, so axum turns it into the HTTP error response for us. Would need some advice/confirmation/reassurance from Ross maybe haha but that's my current understanding

api/src/models/polis_statement_aux.rs
```diff
@@ -262,6 +262,56 @@ pub async fn get_by_id(db: &PgPool, id: &Uuid) -> Result<PolisStatementAux, Comh
     Ok(aux)
 }
 
+/// Fetch many rows by id in a single query. Returns whatever ids exist (order is
+/// unspecified); the caller compares the count to detect missing ids.
+#[instrument(err(Debug))]
+pub async fn get_by_ids(
+    db: &PgPool,
+    ids: &[Uuid],
+) -> Result<Vec<PolisStatementAux>, ComhairleError> {
```
rosswilson2306, line 285: @dtodorowa Could this just be a field on the list method's `filter_options` instead of a new method? The return type looks the same so it looks like we could just combine it.

api/src/models/polis_statement_aux.rs
```diff
@@ -262,6 +262,56 @@ pub async fn get_by_id(db: &PgPool, id: &Uuid) -> Result<PolisStatementAux, Comh
...
+}
+
+/// Bulk-set `moderation_status` for many rows in one statement, returning the
+/// updated rows. Used by batch moderation once the decisions have been forwarded
+/// to Polis. An empty id slice is a no-op.
+#[instrument(err(Debug))]
+pub async fn update_moderation_status_many(
+    db: &PgPool,
+    ids: &[Uuid],
+    status: ModerationStatus,
```
rosswilson2306: @dtodorowa Could this be just a `bulk_update` method that allows you to update other fields as well? Or is this the only thing you'll ever need to update en masse?

api/src/tools/polis.rs
rosswilson2306: Yeah, exactly right. In order to be able to use the `?` syntax, either the function we are calling needs to return our top level `ComhairleError` or there needs to be a `From` conversion implemented for `ComhairleError` for any other error. And all the variants of `ComhairleError` correspond to an http status code when we implement `IntoResponse` https://github.com/crownshy/comhairle/blob/a8efa53e06dcecca7b9657e4962d0a0c56bb6255/api/src/error.rs?plain=1#L356-L384

api/src/models/polis_statement_aux.rs
dtodorowa: `moderation_status` is the only field batch moderation needs today I think, but generalizing to `update_many` costs nothing extra, so I made it general. I'm thinking we might want to batch tag at some point too, but that's slightly different because it is `add theme X to this selection` rather than `change X to Y on this field`.
Definitely not in scope tho, will worry about it when/if the time comes

api/src/models/polis_statement_aux.rs
```diff
@@ -201,38 +201,49 @@ pub struct UpdatePolisStatementAux {
     pub moderation_reason: Option<String>,
 }
 
+impl UpdatePolisStatementAux {
+    /// The `(column, value)` pairs for the fields that are `Some`. Shared by the
+    /// single-row `update` and the bulk `update_many` so both apply the same
+    /// partial-update semantics.
+    fn changed_columns(&self) -> Vec<(PolisStatementAuxIden, SimpleExpr)> {
```
rosswilson2306: The convention we have went with in the other models for this method is `to_values()`. It might be helpful to understand at a glance if it used the same naming convention.

rosswilson2306 (dismissed):
@dtodorowa I left a comment about a naming convention we've used elsewhere that would be good to continue for consistency / reading at a glance. But it's not a massive deal to do now if you are still working on this so I've approved.

api/src/models/polis_statement_aux.rs
dtodorowa: oh yehh, this makes sense, adding it to my provisional style guide, pushed

PR #626, opened 2026-07-13
581/Workflow Design page: two-column Add-a-step dialog + single-row step board

### LINKED ISSUE #581: Workflow Manage/Design page
[Figma Link](https://www.figma.com/design/BoTwbAaBrhJQFFLvlKf6R6/Comhairle-Design?node-id=661-9333&t=e3TIUivZrB5zJqUU-0)

Image

dtodorowa:
TODO:
- [ ] Break this down into multiple PRs because this will be annoying to review

Reworks the **Workflow → Design** page (`/admin/conversations/[id]/design`) for issue #581. Two user-facing changes:

1. **Adding a step** now happens through a two-column **"Add a step" dialog** (step type on the left, a rich detail panel on the right), reusing the same shell as the "Choose a template" dialog. This replaces the old left-hand tool palette.
2. **The board itself** is now a simple vertical list of **single-row step cards** — number, name, tool type, and a `⋮` actions menu. Click a card to configure it; drag or use the menu to reorder.

Plus supporting work: the tool **guide pages** (`/admin/info/tools/*`) were refactored onto a shared, data-driven article component, and the per-step **data-protocol** control moved into the step's Configure screen.

Design: [Figma](https://www.figma.com/design/BoTwbAaBrhJQFFLvlKf6R6/Comhairle-Design?node-id=661-9333)

> Note: this branch had a couple of iterations. An earlier version added a tool palette + inline step previews; both were **removed** later in favour of the simpler dialog + single-row list described here. The PR title/description now reflect the final state.

---

### Add-a-step dialog
- **`AddStepDialog.svelte`** (new) — two-column picker. Left: the 7 backed step types. Right: description, a "setup guide" callout linking to the tool's guide page, tokenised "Example" media blocks, and **BEST FOR / FEATURES / WHAT YOU'D GET** lists. Confirming emits the tool's creation key; the layout creates the step and closes the dialog.
- **`SelectableOptionRow.svelte`** (new) — the shared radio-row used by *both* this dialog and `NewConversationButton` (the template picker), so the two stay visually consistent.
- Opened from three places, all wired to one dialog: the workflow step-strip "+ Add step", the board's empty-state / footer button, and the `?addStep=true` deep link.

### Step board (`design/+page.svelte`)
- Left tool-palette rail removed. Board is a vertical list of single-row cards.
- **Whole card is clickable → Configure** (accessible "stretched link" so keyboard users get a real focusable link).
- **Reorder two ways:** drag the card, *or* use **Move up / Move down** in the `⋮` menu (keyboard-friendly, for devices/users where drag is awkward). Both commit through the same `step_order` patch.
- `⋮` menu holds Rename / Move up / Move down / Delete. Hover state (border + shadow lift + name colour shift) signals the card is clickable.
- Freshly-added steps scroll into view and flash a highlight ring.

### Consolidation
- One shared **`createWorkflowStep.ts`** helper replaces the create-step logic that was duplicated in `+page` and `+layout`.
- Cross-component signals: **`stores/addStepDialog.svelte.ts`** (open the dialog from the board) and **`stores/newStepHighlight.svelte.ts`** (tell the board which new step to highlight).
- **Deleted** the old, now-unused `ToolSelectionModal.svelte` and `avaliable_tools.ts` (a legacy parallel step registry).

### Registry & content (`tool_meta.ts`)
- Renamed tool display names to the friendlier Figma copy (see assumptions below) and added detail-panel fields (`description`, `bestFor`, `features`, `whatYoudGet`).

### Configure screen (`CommonStepConfig.svelte`)
- Replaced the "Ask for sharing consent" switch with a **4-level Data protocol picker** (Confidential / Restricted / Collaborative / Open — the last two disabled pending backing). Same underlying `requestUserSharePermission` boolean.

### Tool guide pages
- New **`tool_guides.ts`** (guide content) + **`GuideArticle.svelte`** (renderer). The per-tool pages under `/admin/info/tools/*` now render from this shared data instead of bespoke markup (e.g. the Polis page dropped ~700 lines).
- The guide layout uses a SvelteKit **layout reset** (`+layout@.svelte`) so guides render full-width without the admin app sidebar (auth still enforced in the load).

---

## Key decisions & assumptions

These came out of a design/grilling pass before building — flagging them so reviewers can push back:

- **Tool display names were renamed app-wide** to match the Figma copy: `learn → Topic onboarding`, `polis → Participant-led poll`, `thinkingspace → Individual view exploration`, `prioritization → Proposal prioritisation` (Survey / Elicitation Bot / Lived Experience unchanged). `tool_meta.displayName` is read in only the board card label, so the blast radius is small.
- **`tool_meta.ts` is treated as the canonical registry.** The legacy `avaliable_tools.ts` was removed.
- ⚠️ **Detail-panel copy is placeholder for 6 of 7 tools.** Only `Topic onboarding` uses finalised Figma copy; the rest is drafted from taglines and **needs product/design sign-off** (marked with `// PLACEHOLDER` comments in `tool_meta.ts`).
- **No photos in the dialog — tokenised CSS "Example" blocks instead.** The app CSP is `img-src 'self' data:`, so external (Unsplash) images would be blocked; this avoids a CSP change and works offline.
- **Data protocol lives in Configure, not on the board.** The board's old time / lock pills were unbacked placeholders (time was client-only, lock decorative) and were dropped; only data protocol is real, so it was re-homed rather than lost.
- **Preview of the participant UI was intentionally dropped** (for now) to ship something simple. Rendering the real participant `UserUI` inline isn't practical — several tools are external iframes (Polis, HeyForm) and all need fully-configured data a fresh step doesn't have. A "preview the real thing" escape hatch via the existing `[[preview]]` route is a possible follow-up.

---

## Testing

- `pnpm test:unit` → **96 passed** (added unit tests for `nextStepOrder` and the `moveItem` reorder helper).
- 
## Known limitations / follow-ups

- **Placeholder tool copy** needs a design/product pass before this is truly done.
- **Not yet driven in a running browser** — the board needs the full stack + a seeded conversation. The interactive bits (stretched-link vs drag, cross-component dialog signal, highlight) are type-valid and unit-covered but unexercised at runtime. Worth a manual smoke test.
- **ESLint `no-navigation-without-resolve`**: the new `<a href>` / `goto` calls don't use `resolve()`. This rule is already unsatisfied ~80× across the repo and the new nav matches the surrounding admin-route style (including the pre-existing Configure link); left consistent rather than diverging.
- A pre-existing type error in `[conversation_id]/+layout.ts` (a possibly-`undefined` stats/steps value) is present in already-committed code on this branch and **not introduced by this PR**.

image

image

image


ui/packages/comhairle/src/routes/(admin)/admin/info/tools/[tool_id]/+page.svelte
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/info/tools/[tool_id]/+page.svelte:1]: All of these info pages seem to have the same layout, so we could probably do `tools/[tool_id]/+page.svelte` and just have the 1 page I think?

ui/packages/comhairle/src/lib/components/GuideArticle.svelte
ADSmith-0, line 1: If we're changing the `info/tools` to be 1 page, then we could probably just inline this inside of the `+page.svelte` and get rid of it as a component? Just having a search I don't think we use it anywhere except the `info/tools` pages right?

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/AddStepDialog.svelte
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/AddStepDialog.svelte:1]: Might just be personal preference, but generally I think it's good to have globally reused stuff inside of `lib` and everything else colocated where possible, so like this I would probably put inside of the `[conversation_id]/design/` folder as it's only used there currently

I guess it's a trade-off though, as being in `lib` will give better discoverability for the devs, but if the component in `lib` is too specific then it can't be reused globally anyway, idk what are your thoughts on it?

ui/packages/comhairle/src/routes/(admin)/+layout.svelte
```diff
@@ -20,13 +20,14 @@
...
+	class="h-svh"
 	style="--sidebar-width: {sidebarWidth.width}px;"
 	data-sidebar-resizing={sidebarWidth.resizing || sidebarWidth.initializing ? '' : undefined}
 >
 	<AdminNav user={data.user} conversations={conversations.records} path={page.url.pathname} />
-	<SideBar.Inset>
+	<SideBar.Inset class="min-h-0 min-w-0">
 		<SidebarFloatingTriggers />
-		<main class="bg-muted flex grow flex-col overflow-y-auto">
+		<main class="bg-muted flex min-h-0 w-full min-w-0 grow flex-col overflow-y-auto">
```
ADSmith-0, line 30: Just curious what are these classes for? Was there an issue with sizing the sidebar?

ui/packages/comhairle/src/lib/utils/reorder.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Return a new array with the item at `index` moved one slot in `direction`
+ * (`-1` = towards the start, `+1` = towards the end). If the move would fall off
+ * either end (or `index` is out of range) the array is returned unchanged (as a copy).
+ * Pure, so it powers the board's button/keyboard reorder without touching the DOM.
+ */
+export function moveItem<T>(items: readonly T[], index: number, direction: -1 | 1): T[] {
+	const next = [...items];
```
ADSmith-0: I think that `items.slice()` is quite a bit quicker when making a copy, it doesn't matter massively, but I only say this because at my last job I had to do a bunch of stuff around efficiency with arrays so feel free to ignore this, just wanted to mention it haha

ui/packages/comhairle/src/lib/stores/addStepDialog.svelte.ts
ADSmith-0, line 1: I think this could potentially be a module export? https://svelte.dev/tutorial/svelte/module-exports,

Basically inside of the `AddStepDialog.svelte`
```html
<script lang="ts" module>
  export function openAddStepDialog() {
    open = true;
  }
  export function closeAddStepDialog() {
    open = false;
  }
</script>
```

Then when you want to open it you can just say

```html
<script lang="ts">
  import { openAddStepDialog, closeAddStepDialog } from "$lib/components/AddStepDialog.svelte";

  openAddStepDialog(); // Opens the dialog
  closeAddStepDialog(); // Close the dialog
</script>
```

ui/packages/comhairle/src/lib/createWorkflowStep.ts
```diff
@@ -0,0 +1,83 @@
...
+		Polis: basic_polis_config,
+		Learn: basic_learn_config,
+		Survey: basic_survey_config,
+		'Lived Experience': basic_lived_experience_config,
+		// basic_elicitation_bot_config only reads `conversation.id`; the cast avoids
+		// requiring the full LocalizedConversationDto that callers don't have to hand.
+		'Elicitation Bot': basic_elicitation_bot_config(conversation as LocalizedConversationDto),
+		'Thinking Space': basic_thinking_space_config(),
+		Prioritization: basic_prioritization_config
+	}[creationKey];
```
ADSmith-0: This might be better as a `switch` statement as it currently looks a bit like we're returning the entire object? I think `switch`s are quicker too

ui/packages/comhairle/src/lib/tool_meta.ts
```diff
@@ -0,0 +1,266 @@
...
+	/** ToolConfig discriminant. */
+	type: ToolType;
+	/** Admin-facing label (Figma display name). Distinct from the internal creation key. */
+	displayName: string;
+	/** One-line tooltip / palette description. */
+	tagline: string;
+	/** Icon shown on cards and the palette. */
+	icon: typeof Icon;
+	/** Route key under /admin/info/tools/<infoKey>. */
+	infoKey: string;
```
ADSmith-0: I think this might make more sense to call `path` or something like that?

ui/packages/comhairle/src/lib/tool_meta.ts
```diff
@@ -0,0 +1,266 @@
...
+	/** Key passed to the add-step switch (see {@link import('./createWorkflowStep').createWorkflowStep}). */
+	creationKey: string;
```
ADSmith-0: Is there a reason we're splitting this out into a separate key? I think we could just use the `type` for the switch? Or would that not work?

Either way I think it would be good for this to be typed so that there's auto-complete and safety, especially if it's on a switch

ui/packages/comhairle/src/lib/components/SelectableOptionRow.svelte
```diff
@@ -0,0 +1,53 @@
...
+		/** Whether this row is the currently-selected option. */
+		selected?: boolean;
+		/** Option title. */
+		name: string;
+		/** Option supporting line. */
+		description: string;
+		/** Called when the row is clicked. */
+		onSelect: () => void;
+		class?: string;
+	} = $props();
```
ADSmith-0: Just based on the other ticket we did, I'm not sure if we wanted this to be a `type Props` as well? Sorry idk if you want me to mention it when I see it or we can just leave it for now until we get like an `eslint` rule that automatically picks it up or something

dtodorowa: oh yess, this was before we had this conversation. It's like a week-old code. Will go over now and change stuff

dtodorowa: it was absolutely appropriate to comment on it, bc I would've not noticed 🙈

ui/packages/comhairle/src/routes/(admin)/+layout.svelte
dtodorowa: yeh i noticed a flexbox overflow issue when we had the horizontally scrollable steps. Flex items default to` min-width/min-height: auto`, which refuses to shrink below their content's intrinsic size, that was letting wide content (and the resizable sidebar width) push the layout out and break the overflow-y-auto scroll container. `min-h-0 min-w-0` lets the inset/main shrink properly so content scrolls instead of blowing out the layout. 

We don't have the horizontally-scrollable layout anymore, so this won't be an issue anymore, but idk, do you vote for keeping or nuking? @ADSmith-0

ui/packages/comhairle/src/lib/utils/reorder.ts
dtodorowa: meh, it's a tiny change, so I'm happy to do it. pushed

ui/packages/comhairle/src/lib/stores/addStepDialog.svelte.ts
dtodorowa: hmm, actually I think this can be simplified. Store now just owns the open boolean directly as the single source of truth, which let me delete the counter, the local mirror state, and the effect.

dtodorowa:
Random but should resolve - notifications cover the button
image

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/AddStepDialog.svelte
```diff
@@ -0,0 +1,177 @@
...
+		adding = false,
+		onAdd
+	}: {
+		/** Two-way open state. */
+		open?: boolean;
+		/** Disables the confirm button while the parent is creating the step. */
+		adding?: boolean;
+		/** Called with the chosen tool's `creationKey` when the user confirms. */
+		onAdd: (creationKey: CreationKey) => void;
+	} = $props();
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/AddStepDialog.svelte]: There's also a `prop` bit here if you want to change it, but I'm happy either way 🙂

ADSmith-0 (dismissed):

ui/packages/comhairle/src/lib/tool_meta.ts
dtodorowa: called it Slug, is that an okay middle-ground?

ADSmith-0: Yeah that sounds good! :slightly_smiling_face:

PR #631, opened 2026-07-13
Adds an arbitrary metadata field to conversations

stuartlynn:
Adds a metadata field to conversations that allows us to store arbitrary data on them. 

Adds methods to read / write the metdata

### DIFF
--- a/api/src/models/conversation.rs
+++ b/api/src/models/conversation.rs
@@ -82,13 +82,14 @@ pub struct Conversation {
     pub call_to_action: Option<TextContentId>,
     pub enable_signup_prompts: bool,
     pub show_thank_you_page_annon_instructions: bool,
+    pub metadata: serde_json::Value,
     #[partially(omit)]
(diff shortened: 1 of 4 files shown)


api/src/models/conversation.rs
```diff
@@ -545,6 +550,40 @@ pub async fn update(
...
+                updated_at = NOW()
+            WHERE id = $2
+            RETURNING *",
+    )
+    .bind(patch)
+    .bind(id)
+    .fetch_one(db)
+    .await
+    .map_err(|e| match e {
+        sqlx::Error::RowNotFound => ComhairleError::ResourceNotFound("Conversation".into()),
```
rosswilson2306: Similar to the other PR, you could use the `resolve_db_err` trait method to try and standardise / dry up the error handling for this but not essential.

rosswilson2306 (dismissed):
Left a similar comment about error handling but not essential so approving.

PR #632, opened 2026-07-14
608/conversation permissions

rosswilson2306:
Actions:

+ create new struct for ConversationContentEditor permission
+ adjust params for granting permission to include email lookup for user
+ remove reference usage in UserOrOrganizationId
+ form for granting permissions to users
+ new mailer method to notify user of permission update
+ send notification email from permission `grant` endpoint
+ model method and endpoint to list conversations a user is permitted
  access to
+ refactor AdminNav to render list of permitted conversations as well as
  owned conversations
+ create ConversationResource struct with implementation for extracting
  conversation_id from request
+ create OwnedResource trait for extracting owner_id from resource
+ refactor RequiredUserPermission extractor to allow access if user is
  resource owner as well as permitted users

Motivation:

+ allow conversation owners to grant edit access to other users
+ need to be able to grant permissions without exposing user ids on
  shared servers
+ Uuid are copy and tying lifetime to input arguments caused ownership
  issues when requesting data from the db
+ notify user that their permissions have been updated for a given
  resource
+ block conversation endpoints to users without appropriate permission
+ allow conversation owners to access same endpoints without having to
  backfill owner permissions for all conversations and keep in sync


api/src/models/permissions.rs
```diff
@@ -43,8 +48,46 @@ impl FromRequestParts<Arc<ComhairleState>> for SystemResource {
...
+        let Path(ConversationPath { conversation_id }) =
+            Path::<ConversationPath>::from_request_parts(parts, state)
+                .await
+                .map_err(|_| {
+                    ComhairleError::ResourceNotFound(
+                        "Path must contain a conversation_id".to_string(),
+                    )
+                })?;
+
+        let conversation = models::conversation::get_by_id(&state.db, &conversation_id).await?;
```
rosswilson2306, line 78: Should maybe try to refactor to reuse the permission redis cache to store whether a user is the resource owner instead of running this lookup on every request that uses `RequiredUserPermission<_, ConversationResource>`.

api/src/models/permissions.rs
```diff
@@ -131,16 +218,16 @@ const DEFAULT_COLUMNS: [ResourcePermissionIden; 9] = [
 ];
 
 /// Represents either a user or an organization for role assignment purposes.
-#[derive(Debug)]
-pub enum UserOrOrganizationId<'request> {
-    User(&'request Uuid),
-    Org(&'request Uuid),
+#[derive(Debug, Copy, Clone)]
+pub enum UserOrOrganizationId {
```
rosswilson2306, line 222: I removed the reference as `Uuid`s  impl Copy so can be passed around without references easily enough. The references also caused an ownership issue when I had to refactor the `resolve_actor` to run a db lookup for `get_user_by_email` because the lifetime tied the reference to the input args, which the db lookup couldn't satisfy.

api/src/models/users.rs
```diff
@@ -555,6 +567,93 @@ pub async fn upgrade_account(
...
+                .order_by((UserIden::Table, UserIden::CreatedAt), order.into())
+                .to_owned();
+        }
+
+        query
+    }
+}
+
+#[instrument(err(Debug))]
+pub async fn list(
```
rosswilson2306, line 641: Started creating this in case I needed it and it seems like we may need a way to list users if we build a separate interface for managing users / permissions. Currently no endpoint for this so not exposed to the api.

api/src/routes/permissions.rs
```diff
@@ -63,18 +68,26 @@ pub struct ListPermissionsQuery {
...
-    match (user_id, organization_id) {
-        (Some(uid), None) => Ok(Some(UserOrOrganizationId::User(uid))),
-        (None, Some(oid)) => Ok(Some(UserOrOrganizationId::Org(oid))),
-        (None, None) if allow_none => Ok(None),
+) -> Result<Option<UserOrOrganizationId>, ComhairleError> {
+    match (user_id, organization_id, user_email) {
+        (Some(uid), None, None) => Ok(Some(UserOrOrganizationId::User(uid))),
+        (None, Some(oid), None) => Ok(Some(UserOrOrganizationId::Org(oid))),
+        (None, None, Some(u_email)) => {
+            let user = users::get_user_by_email(&u_email, db).await?;
```
rosswilson2306, line 84: Needed to provide a way of conversation owners assigning permissions to other users without exposing a list of admin users on shared servers.

api/src/routes/permissions.rs
```diff
@@ -96,7 +109,15 @@ async fn grant(
     Path(path): Path<TargetResourceId>,
     Json(body): Json<GrantPermissionBody>,
 ) -> Result<(StatusCode, Json<permissions::ResourcePermission>), ComhairleError> {
-    let actor_id = resolve_actor(&body.user_id, &body.organization_id, false)?.unwrap();
+    let actor_id = resolve_actor(
+        &state.db,
+        body.user_id,
+        body.organization_id,
+        body.user_email,
+        false,
```
rosswilson2306, line 117: Maybe it would be better to make `resolve_actor` use a builder pattern?

api/src/models/permissions.rs
```diff
@@ -1189,4 +1330,63 @@ mod tests {
...
+            users_with_permission.iter().any(|u| u.id == user_a_id),
+            "missing user_a"
+        );
+        assert!(
+            users_with_permission.iter().any(|u| u.id == user_b_id),
+            "missing user_b"
+        );
+        assert!(
+            !users_with_permission.iter().any(|u| u.id == user_c_id),
+            "user_b incorrectly included"
```
alpinegeorge: Should be "user_c" in error message

alpinegeorge (dismissed):

api/src/models/permissions.rs
alpinegeorge: This sounds like a good idea. Modifying the cache key to include new information shouldn't cause issues across versions since the keys won't cross over. Old keys will also simply expire out after the TTL, so we shouldn't need to clean them up.

api/src/routes/permissions.rs
alpinegeorge: That would be useful. Ideally, it could be generified so it's not just a permissions only concept.

api/src/routes/conversations.rs
```diff
@@ -70,6 +75,10 @@ async fn create_conversation(
 async fn update_conversation(
     State(state): State<Arc<ComhairleState>>,
     RequiredAdminUser(_user): RequiredAdminUser,
```
dtodorowa, line 77: Think this blocks the edit feature, an invited editor will get bounced before the permission check runs? 

I think the test passes because `test@crown-shy.com` got added to admin users

I am wrapping my head around this still, but I assume the permission extractor should be the gate? If this is bs, let me know 🤭

api/src/routes/permissions.rs
dtodorowa, line 106: grant requires `SystemAdminRole` but we show the grant form to conversation owners. It works rn now because admin signup happens to hand out that DB role. But an an owner who isn't a system admin would hit a 403? 

Should the gate here be owner-or-admin instead?

dtodorowa (approved):

api/src/models/permissions.rs
rosswilson2306: Cool, I'll look at that as part of a separate PR.

api/src/routes/permissions.rs
rosswilson2306: Likewise, I'll look at this as part of a separate PR.

api/src/routes/conversations.rs
rosswilson2306: An invited editor will need to be an admin user otherwise that shouldn't have access to do anything in admin.

api/src/routes/permissions.rs
rosswilson2306: Same as before, users can have many roles and a conversation owner will need to be SystemAdmin (same as the previous admin users).

api/src/routes/conversations.rs
dtodorowa: oh I seeee! My imagination was going wild for a sec 🤭 yeh yeh this makes sense

PR #648, opened 2026-07-17
Update the learn step setup UI

### LINKED ISSUE #629: Update the learn step setup UI
### Description
Currently the learn step setup UI doesn't have great UX, we should update it to match the new designs

Image

### Designs
Link: https://www.figma.com/design/BoTwbAaBrhJQFFLvlKf6R6/Comhairle-Design?node-id=953-5439&t=5erhAzU9j8qxYfXh-4

Image

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/629

### Testing

https://github.com/user-attachments/assets/ffb1df74-475f-4630-b501-8ab79aee59c9

- Checked that the translations still worked correctly

https://github.com/user-attachments/assets/186d78db-2d12-452a-ac9f-28707c4d45ef


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/+page.svelte
```diff
@@ -21,27 +19,11 @@
...
-		apiClient
-			.ListDocuments({ params: { conversation_id: cid } })
-			.then((docs) => {
-				availableDocuments = docs.filter((d) => d.parse_status === 'DONE');
-			})
-			.catch(() => {
-				availableDocuments = [];
-			});
-	});
-
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/+page.svelte]: My editor was saying that all this is unused?

ui/packages/comhairle/src/lib/tools/learn/Pages.svelte.ts
ADSmith-0, line 1: I split the pages functionality into a class because I was finding hard to keep track of everything when it was in the script tags

ui/packages/comhairle/src/lib/tools/learn/Pages.svelte.ts
```diff
@@ -0,0 +1,143 @@
...
+							if (page[translation].lang !== lang) {
+								page[translation].requires_validation = true;
+							}
+						}
+						break;
+					case 'target':
+						break;
+				}
+				this.items[this.currentId] = page;
+				console.debug('requires_validation:', page);
```
dtodorowa: should prob nuke this

ADSmith-0: Oh yeah good catch! :slightly_smiling_face:

dtodorowa:
Spotted a bug:

Deleting the first (leftmost) page → blank editor, nothing selected

1. Open a Learn step with 3 pages.
2. Click Page 1 (leftmost).
3. Click Delete Page.


(I THINK, but might have to double check)
delete sets `currentId = Math.max(index - 1, 0) → max(-1, 0) = 0`, but it just deleted item key "0". So zcurrentId = 0z now points at a page that no longer exists. zitems[0]z is undefined → blank editor, and no button matches so nothing highlights

https://github.com/user-attachments/assets/c5198cfe-868a-4692-b7fe-887d001975fb

dtodorowa:
Probably nit-picky, but if we add a new page, I would expect the UI to navigate to it, so I can start editing?
But maybe this is a design question, hmm

https://github.com/user-attachments/assets/7e30214b-de9f-417c-a24e-84e8abcaf08f

dtodorowa:
I think this is something we always had, but now it's much faster to change pages.
Super edge case. Also I doubt people would be editing so quickly.

(quickly change after editing overwrites the content of the page) 
Fix would potentially be blocking user actions until we save (but we gotta refactor the translatable field stuff I think) 



https://github.com/user-attachments/assets/f0b892b8-37f9-46ba-b3de-059923fb00d5

ADSmith-0:
> Probably nit-picky, but if we add a new page, I would expect the UI to navigate to it, so I can start editing? But maybe this is a design question, hmm
> 
>  Screen.Recording.2026-07-21.at.10.22.58.mov

That's a really good shout I can add that now! :slightly_smiling_face:

ADSmith-0:
> I think this is something we always had, but now it's much faster to change pages. Super edge case. Also I doubt people would be editing so quickly.
> 
> (quickly change after editing overwrites the content of the page) Fix would potentially be blocking user actions until we save (but we gotta refactor the translatable field stuff I think)
> 
>  Screen.Recording.2026-07-21.at.10.25.38.mov

@dtodorowa 

Yeah I was a bit worried about this, I wasn't sure how much of an issue it was going to be, I guess that the simplest solution would be to noop page switching until the save has completed? What do you think?

ADSmith-0:
> I think this is something we always had, but now it's much faster to change pages. Super edge case. Also I doubt people would be editing so quickly.
> 
> (quickly change after editing overwrites the content of the page) Fix would potentially be blocking user actions until we save (but we gotta refactor the translatable field stuff I think)
> Screen.Recording.2026-07-21.at.10.25.38.mov

This should be fixed now! I've just blocked page switching until the save happens, it will probably be better once we refactor then we can tell when the save has completed and have a callback or something a bit better?

dtodorowa:
Super minor: 

not sure how i got to this state, but worth flagging

image

https://github.com/user-attachments/assets/e331735b-8eab-4c61-842f-225185950fe6

dtodorowa:
We have some dead skeleton code now (Example - refreshing: dark: staging, light: local)
Should either wire `isInitialLoad` to a real initial state or drop the skeleton branches. What do you think?


https://github.com/user-attachments/assets/3bd396c7-c153-497c-95ec-f82fd4e8a2d4

dtodorowa:
Minor bug. I don't even know if we care about this dialog title so much (should prob reduce the text size too lol)

**Reorder (or delete), then the Translate dialog title lies**

- Open a Learn step with 3 pages.
- Drag Page 3 to the front.
- Click the now-first page and open Translate.
- What happens: dialog says "Translate: Page 3" while the button you clicked reads "Page 1".

Why: I think because the button label uses order position (i + 1), but the dialog uses pages.currentId + 1, which is the item id, not the position. They will diverge after any reorder or delete.

https://github.com/user-attachments/assets/74d69e4d-ea54-431d-82f7-e23b9ae98499


And on delete:

https://github.com/user-attachments/assets/d2438b24-1248-4178-925d-0da6eb4e7539

ADSmith-0:
> Minor bug. I don't even know if we care about this dialog title so much (should prob reduce the text size too lol)
> 
> **Reorder (or delete), then the Translate dialog title lies**
> 
>     * Open a Learn step with 3 pages.
> 
>     * Drag Page 3 to the front.
> 
>     * Click the now-first page and open Translate.
> 
>     * What happens: dialog says "Translate: Page 3" while the button you clicked reads "Page 1".
> 
> 
> Why: I think because the button label uses order position (i + 1), but the dialog uses pages.currentId + 1, which is the item id, not the position. They will diverge after any reorder or delete.
> Screen.Recording.2026-07-21.at.11.36.12.mov
> 
> And on delete:
> Screen.Recording.2026-07-21.at.11.25.59.mov

Yeah that's a fair point! It's a bit of a pain to read the correct index at that part of the code, so instead I've switch it to just be "translate", as you can only do it for the current page anyway I think it should still make sense?

Also changed the title styling a bit to be a bit more subtle :slightly_smiling_face: 

Screenshot from 2026-07-21 12-09-46

ADSmith-0:
> We have some dead skeleton code now (Example - refreshing: dark: staging, light: local) Should either wire `isInitialLoad` to a real initial state or drop the skeleton branches. What do you think?
> Screen.Recording.2026-07-21.at.11.24.29.mov

Oh yeah whoops! I forgot to re-enable the `isInitialLoad` I disabled it for testing :sweat_smile: Great catch!

In terms of skeletons, I've not used them much before but I'm happy to use them! :slightly_smiling_face:

ADSmith-0:
> Super minor:
> 
> not sure how i got to this state, but worth flagging
> image
> Screen.Recording.2026-07-21.at.11.23.00.mov

Yeah I'm able to recreate it, but it's not super consistent, that's annoying I think I might have an idea as to what's causing it though, I'll take a look!

ADSmith-0:
> Super minor:
> 
> not sure how i got to this state, but worth flagging
> image
> Screen.Recording.2026-07-21.at.11.23.00.mov

@dtodorowa 

I can't seem to consistently recreate this, do you know how to? Otherwise would we be happy to spin this off into a separate ticket?

dtodorowa (approved):
Let's test this bad boi on staging

dtodorowa:
> @dtodorowa
> 
> I can't seem to consistently recreate this, do you know how to? Otherwise would we be happy to spin this off into a separate ticket?

nahh, not so sure, let's address in the future, would love to merge this and play with it on staging.

I tried to do the refactor we spoke about, there are some test cases in the description that might be worth going over
(#675)

PR #654, opened 2026-07-19
Server-render conversation tab strips from data

dtodorowa:
The refactor this stack builds toward: **every conversation-admin sub-tab strip now server-renders from layout data instead of being injected post-hydration via a client `$effect`.** Strips no longer flash in or vanish on a hard refresh.

## Why
Strips were pushed up into the header via a context (`tabExtras`) `$effect`, which doesn't run
during SSR, so on refresh the row was briefly empty (and sometimes stayed gone). Rendering each
strip from the layout that owns its data fixes it everywhere, removes a Polis over-fetch, and
lets us delete the whole context + skeleton-injection subsystem.

## Changes
- **Workflow step strip (Row 3)** rendered from `data.workflowSteps` in the conversation layout,
  plus section-switching skeletons; tab-strip gutter padding aligned.
- **Step sub-tabs → real nested routes** (`configure` / `setup` / `moderation` / `insights`
  under `design/step/[step_id]/`), each with its own load. Polis Moderation/Insights now fetch
  their data **only when opened** (kills the previous over-fetch); non-Polis steps guard-redirect
  to `configure`.
- **Configure, Recruit, and Events strips** server-render from layout data / small tab constants
  (`configure/tabs.ts`, `invites/tabs.ts`, `events/[event_id]/tabs.ts`). Configure/Recruit/Events
  sub-tabs stay `?tab=`/`?subtab=` (sections of one shared dataset); only the step sub-tabs became
  routes (distinct data + weight).
- **`tabExtras` / `CONVERSATION_TAB_EXTRAS_CTX` deleted** — no strip is injected via `$effect`
  anymore. Events list is loaded in the conversation layout (parallel with workflows, no added
  latency) so its strip can render there too.

## Testing
- Verified in-browser end-to-end: strips survive hard refresh; sub-tab nav swaps content with no skeleton flicker; non-Polis steps redirect correctly; event-detail page renders Row 3 + Row 4 with `?subtab` switching.

## Notes
- Top of the 6-PR stack; spans the `align-workflow-step-strip` + `refactor-step-subtab-strip-from-data`
  branches. Merge last, after 1–5.


alpinegeorge (approved):
This all looks good, as far as I can tell.
I'd get Ross's review as well and then it's good to merge.

rosswilson2306 (approved):

PR #675, opened 2026-07-21
Refactor translatable field

dtodorowa:
Also contains #678 #680 #683 Just merged those

`TranslatableField` had grown **two personalities**, forked at runtime on whether you passed a `translation` prop: a `textContent` mode that saved itself and showed a Saving/Saved indicator, and a `callback` mode that delegated persistence to `onSave*` props and showed **no** indicator. One `isTextContentMode` flag branched the component in ~8 places, and it bridged two different data shapes internally (`TextContent` rows vs learn's inline `ExtendedLocalizedPage[]`).

That muxing is what caused the learn/setup-step bugs:
- **No save indicator** (indicator state only ever set in the textContent branch).
- **Content clobbered on fast page switches** the component's own 1s debounce fired after the page changed, writing the old page's text into the new one.
- **A full reload per keystroke**, plus two out-of-sync copies of the same content (field debounced at 1000ms, dialog at 500ms).

### What this does

Introduces one **`TranslationSource`** contract that owns both reading and persisting a field's translations, and reduces `TranslatableField` / `TranslationEditor` to dumb views over it. The source owns the debounce, `saveState`, and `flush()`; the components just render and forward intent. Two implementations: `createTextContentSource` (backed by a `TextContent`) and `createLearnSource` (a thin adapter over learn's `Pages` controller). An optimistic overlay keeps `source.contents` current on every keystroke, so `RichTextEditor` stops fighting the cursor and the dialog no longer needs its own copy.

Full rationale + the decisions behind it are in **documentation/adr/0005-translations-single-source-contract.md**.

All consumers are migrated: learn, step name/description, conversation `configure` (incl. the create-then-autosave nullable fields), events, and prioritization. The old `translation`/`onSave*` API and the dual-mode component are gone.

> Heads up: events + configure still show some **pre-existing** superForm/zod type errors (predating this PR)

## QA checklist

**Learn / setup step** (the main fix)
- [ ] Editing a page shows Saving → Saved.
- [ ] Type in page A, immediately switch to page B before it saves: A keeps its text, B isn't overwritten.
- [ ] Typing doesn't fire a reload per keystroke; no flicker mid-typing.
- [ ] Sustained typing never resets the editor / jumps the cursor.
- [ ] Add / delete / reorder pages works; save state settles after each.

**Step name & description, conversation configure, events, prioritization**
- [ ] Each field saves, shows the indicator, and survives a reload.
- [ ] Target-locale editing, AI translate, Approve, Mark as draft all work.
- [ ] **configure**: required-field validation errors still show/clear as you type.
- [ ] **configure**: first edit of an empty nullable field (privacy policy, FAQs…) creates + saves it.
- [ ] **prioritization**: add/remove/edit sections + the per-section translate dialog.

**Cross-cutting**
- [ ] Type into a field and immediately navigate away: the pending save flushes, no lost edits.
- [ ] Force a failed save (offline / 500): indicator shows an error, content isn't silently dropped.


ui/packages/comhairle/src/lib/components/Translation/TranslatableField.svelte
```diff
@@ -35,31 +23,19 @@
 		maxHeight?: string;
 		dialogMinHeight?: string;
 		dialogTitle?: string;
-		inputProps?: Record<string, any>;
+		inputProps?: Record<string, unknown>;
```
ADSmith-0: I think that we might be able to do a discriminated union here, like this:

```ts
	type Props = BaseProps &
		(
			| { inputType?: 'input'; inputProps?: ComponentProps<typeof Input> }
			| { inputType?: 'textarea'; inputProps?: ComponentProps<typeof Textarea> }
		);

```

I've been playing around with it for about 20 minutes but I can't get it quite right, happy to leave it as is though 🙂

ui/packages/comhairle/src/lib/components/Translation/translationSource.svelte.ts
```diff
@@ -0,0 +1,249 @@
...
+				// flip the indicator to "saved" while another is still in flight.
+				if (inFlightCount === 0) {
+					saveState = ok ? 'saved' : 'error';
+					if (ok) {
+						savedResetTimer = setTimeout(() => {
+							if (saveState === 'saved') saveState = 'idle';
+						}, 2_000);
+					}
+				}
+			}
```
ADSmith-0: Can use `tryCatchAsync` here if we want

ui/packages/comhairle/src/lib/tools/learn/LearnManage.svelte
```diff
@@ -77,6 +78,8 @@
 
 		if (response.err !== null) {
 			notifications.send({ message: 'Failed to save changes', priority: 'ERROR' });
+			// Rethrow so Pages can flip saveState to 'error' and leave the page marked dirty.
+			throw response.err;
```
ADSmith-0, line 88: Should this be a `return`? Based on the comment here it sounds as though it would make the user lose all their changes?

ui/packages/comhairle/src/lib/components/Translation/TranslatableField.svelte
dtodorowa: Let me know if [this commit](https://github.com/crownshy/comhairle/pull/675/changes/1fd19b3888f63b3f3d356f275fcb8cfd7c1c6a7d) makes sense. couldn't get it to narrow cleanly either :/ 

left a comment explaining why the cast's there. Happy to revise

ui/packages/comhairle/src/lib/components/Translation/translationSource.svelte.ts
dtodorowa: we want we want! (we just forgot 👉👈 🤭)

ui/packages/comhairle/src/lib/tools/learn/LearnManage.svelte
dtodorowa: yeah throw is deliberate here, it's what keeps their edits. i think the name is a problem maybe: `restore` actually means "mark clean", not "recover their work". 

will rename it to `markSaved()` so it stops being a trap? wdyt? is this enough? good catch btw, that would've confused me going back to this code too

ui/packages/comhairle/src/lib/components/Translation/TranslatableField.svelte
ADSmith-0: Yeah I think that's good! :slightly_smiling_face:

ui/packages/comhairle/src/lib/tools/learn/LearnManage.svelte
ADSmith-0: Oh no sorry I meant the throw response.err, should it be return response.err, idk if a throw would break the app?

With restore I named after like git convention like git restore removes local changes, but I'm happy to change it to markSaved() if we want 🙂

dtodorowa: ahh wait maybe i'm misunderstanding, but i think if we use return here, the save would look successful and it'd mark everything clean, so the user's failed save silently looks fine. the throw is what protects their edits, not what loses them, right?

on the naming, I think the git comparison is tripping me up a bit. `git restore` throws your local changes away and takes you back to the saved version, right? this function does the opposite, me thinks:
- takes whatever you've currently got
- marks *that* as the saved state (`areDirty = false`)

so if we keep the `restore` name I think it reads as "revert my edits" when it actually keeps them? idk, let me know if i've got your point backwards, I sometimes get stuck in a logic branch and don't consider other important things. happy to talk it through more 😊

ADSmith-0: Oh right ok I think I see what you mean! Yeah I guess maybe `restore()` isn't the right phrasing, if we're continuing the git analogy I guess it would be more like `pull`?

But yeah I think that's fine! No worries! 🙂

ADSmith-0 (approved):

PR #678, opened 2026-07-21
Add "unsaved changes" guard to prevent losing edits on the Learn Edit

dtodorowa:
Add "unsaved changes" guard to prevent losing edits

Builds on top of the translatable field refactor in #675 

https://github.com/user-attachments/assets/f1b0e939-e1a9-41a6-8f30-d5a4693731ee

### DIFF
--- a/ui/packages/comhairle/src/lib/utils/unsavedChangesGuard.svelte.ts
+++ b/ui/packages/comhairle/src/lib/utils/unsavedChangesGuard.svelte.ts
@@ -1,15 +1,18 @@
 import { beforeNavigate } from '$app/navigation';
+import { onMount } from 'svelte';
 
 /**
  * Warn before leaving the page while there are unsaved changes, e.g. a refresh mid-save on the learn
  * step. Covers a full-page refresh / tab close / external navigation (via the browser's native
  * prompt) and in-app SvelteKit navigation.
  *
- * Call during component initialisation (it uses `$effect` and `beforeNavigate`) and pass a getter
+ * Call during component initialisation (it uses `onMount` and `beforeNavigate`) and pass a getter
  * for the "dirty" condition, e.g. `guardUnsavedChanges(() => pages.areDirty)`.
  */
 export function guardUnsavedChanges(hasUnsavedChanges: () => boolean) {
-	$effect(() => {
+	// onMount (not $effect): the listener is wired up once and never depends on reactive state. The
+	// dirty check runs at event time via the getter, so there is nothing to re-run on state change.
+	onMount(() => {
 		function handleBeforeUnload(event: BeforeUnloadEvent) {
 			if (!hasUnsavedChanges()) return;
 			// Triggering the native "you have unsaved changes" prompt: modern browsers key off
--- a/ui/packages/comhairle/src/lib/utils/unsavedChangesGuard.svelte.ts
+++ b/ui/packages/comhairle/src/lib/utils/unsavedChangesGuard.svelte.ts
@@ -1,0 +1,31 @@
+import { beforeNavigate } from '$app/navigation';
+
+/**
+ * Warn before leaving the page while there are unsaved changes, e.g. a refresh mid-save on the learn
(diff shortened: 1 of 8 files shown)


ui/packages/comhairle/src/lib/utils/unsavedChangesGuard.svelte.ts
```diff
@@ -0,0 +1,31 @@
...
+		function handleBeforeUnload(event: BeforeUnloadEvent) {
+			if (!hasUnsavedChanges()) return;
+			// Triggering the native "you have unsaved changes" prompt: modern browsers key off
+			// preventDefault(); older ones need returnValue set. The message itself is not customisable.
+			event.preventDefault();
+			event.returnValue = '';
+		}
+		window.addEventListener('beforeunload', handleBeforeUnload);
+		return () => window.removeEventListener('beforeunload', handleBeforeUnload);
+	});
```
ADSmith-0: Would this be better as a `onMount` or would that not work?

dtodorowa: Good q! I think `onMount` would work here too, so happy to change and document when we should use one or the other in the style guide.

I went with `$effect` since we're aiming to use Svelte 5/runes,etc, and we already do the same listener + cleanup thing in `ReportNav.svelte` and `CollapsibleChat.svelte`. Behaviour is the same either way tho.

lmk if you want me to switch it, I'm cool either way, mostly because I do not have a deep enough understanding probably. But this sounds like a great opportunity for me to dive deeper 🤭

dtodorowa: Did some research:
I guess `onMount` signals the intent better and says "we are just setting this up once, won't rerun it"
With `$effect`, someone reading would expect the block to rerun and maybe waste some time looking for when this happens? 
 
Another thing that popped out was that Svelte team's guidance is basically "use effects to sync with reactive state"

So I think it's prob a consistency call at this point? Should we add a "when do we reach for onMount vs $effect"  bit somewhere in the style-guide? I'm cool with changing all things to onMount, now I'm mostly confused and have decision paralysis haha

ADSmith-0: Yeah no worries! :smile: 

I think that Svelte 5 also allows for life cycle methods like `onMount` and `onDestroy`

I think of it more like:
- `$effect` is more for side effects
- `onMount` only fire once when we mount
- `onDestroy` only fire once when we unmount

With the Svelte docs they say to [avoid `$effect` if you can help it](https://svelte.dev/tutorial/svelte/effects)

Screenshot from 2026-07-21 17-34-20

But yeah `onMount` only runs once so it's better for performance, like if we don't need to run it multiple times then we should avoid it if we can, so I think it's good practice use `onMount` by default unless we can't - the only reason I was wondering here is because of the callback, I was wondering if that wouldn't be set in time, but if it is then I'd definitely go for `onMount`

I think generally garbage collection and assigning event listeners is quite slow in JS so the more we can avoid it the better I think :slightly_smiling_face:

dtodorowa: thanks for taking the time to explain this 🥹 <3 changing!

ADSmith-0: No worries, any time! :smile:

dtodorowa: added [this](https://github.com/crownshy/comhairle/pull/678/changes#diff-476f8d3866159e359d47fe8844c04fa4fc9e658d9e31f771b73701474081c488R319) to the style guide as well. maybe it doesn't belong there, but I want to have a place to document learnings from reviews

ADSmith-0 (approved):

PR #686, opened 2026-07-22
551 audio recording streaming endpoint

### LINKED ISSUE #551: Audio Recording streaming endpoint
We want to enable recording of live events as well as uploading  of recordings of past events. The use case for this would be for in person meetings with breakout rooms.  The goal is to generate a link that attendees can use for the event which would lead to a page with a big record button. When breakout discussions are formed, one participant would navigate to that link either through a QR code or clicking the link and it would show them a big record button (potentially also ask for how many people are in the conversation).

Pressing that button should start a recording. They should be able to pause it and if their browser refreshes or the connection drops, to be able to renew the upload.

Once done they should be able to hit a Done button and the recording would be finalized and submitted to the transcription / processing pipeline.

WE might not know ahead of time how many breakout rooms there will be so we will need to have this be dynamic. 

Some considerations / requirements

1. Add an endpoint on events that will create a new audio recording and return the steam upload link
2. That link should allow streamed recording with upload 
3. It should allow interruptions but unintentional or deliberate 
4. It should allow finalization of submitted audio recordings.
5. IT might need some kind of timeout as well so that if people forget to hit done we automatically finalize what we have

This might be possible with multipart signed upload urls but I suspect it might be easier to stream it through a  rust endpoint.

alpinegeorge:
Motivations:

 - We want participants of an event to be able to record conversations, live, uploading the audio data as they go for faster processing and to ensure the data is not lost if the recording is interrupted for any reason.
 - In order to do this, our bulk storage service should support a continuous upload process, such as multipart upload.
 - Users should also not be able to modify other's recordings and must only be able to record in a single session at a time.

Actions:

 - Added multipart upload support to BulkStorageService and S3StorageService.
 - Added endpoints and models for creating, managing and finalising live recordings.
 - Added user ownership and locking mechanisms to live recordings.
 - Added websocket handler to automatically pause live recordings on disconnect and allow for simple part uploads via messaging.


api/src/models/live_audio_recording.rs
```diff
@@ -0,0 +1,309 @@
...
+        .build_sqlx(PostgresQueryBuilder);
+
+    let row = sqlx::query_as_with::<_, RawLiveAudioRecording, _>(&sql, values)
+        .fetch_one(db)
+        .await?;
+
+    row.try_into()
+}
+
+pub async fn list_by_event(
```
alpinegeorge, line 164: It would be better to combine this with the other list methods using a builder pattern.

api/migrations/20260716150000_create_live_audio_recording_table.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE live_audio_recording (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    audio_recording_id uuid NOT NULL UNIQUE REFERENCES audio_recording(id) ON DELETE CASCADE,
+    multipart_upload_id TEXT NOT NULL,
+    next_part_number INTEGER NOT NULL DEFAULT 1,
+    uploaded_parts JSONB NOT NULL DEFAULT '[]'::jsonb,
+    owner_id uuid NOT NULL REFERENCES comhairle_user(id) ON DELETE CASCADE,
+    locked BOOLEAN NOT NULL,
```
alpinegeorge, line 8: Thinking about this, this should probably belong with runtime state to ensure that a locked recording doesn't stay locked over an unexpected server restart.

api/src/bulk_storage_service/s3_storage.rs
```diff
@@ -195,4 +196,107 @@ impl BulkStorageService for S3StorageService {
...
+        let expires_in: Duration = Duration::from_secs(PUT_EXPIRES);
+
+        let presigned = self
+            .s3_client
+            .upload_part()
+            .bucket(&self.bucket)
+            .upload_id(&upload_id.0)
+            .key(path)
+            .part_number(part_number.0)
+            .presigned(PresigningConfig::expires_in(expires_in).unwrap())
```
rosswilson2306: @alpinegeorge Is the `.unwrap()` here safe?

api/src/models/live_audio_recording.rs
alpinegeorge, line 1: Needs model unit tests

api/src/routes/audio_recordings.rs
alpinegeorge, line 1: Could also use some tests.

api/src/websockets/handlers/audio_recording.rs
```diff
@@ -0,0 +1,1083 @@
...
+                    "audio_recording:acquire_result",
+                    Some(request.request_id),
+                    err.to_string(),
+                )
+                .await
+            }
+        }
+    }
+
+    async fn handle_presign_part(
```
alpinegeorge, line 377: All of the handler functions have a similar signature and shape. Refactor to re-use code?

api/src/websockets/handlers/audio_recording.rs
```diff
@@ -0,0 +1,1083 @@
...
+        message: &WebSocketMessage,
+        connection: &WebSocketConnection,
+        state: &Arc<ComhairleState>,
+    ) -> Result<(), WebsocketError> {
+        let WebSocketMessage::Custom { event, data } = message else {
+            return Ok(());
+        };
+
+        match event.as_str() {
+            "audio_recording:acquire" => {
```
alpinegeorge, line 789: The bodies here also have similar shapes. Refactor to re-use?

api/src/models/live_audio_recording.rs
```diff
@@ -0,0 +1,309 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct UploadedPart {
+    pub part_number: i32,
+    pub etag: String,
+    pub size_bytes: i64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
```
rosswilson2306: @alpinegeorge I'm obviously missing something here but why is this struct needed as well as the `RawLiveAudioRecording`? Also, is `#[serde(rename_all = "snake_case")]` doing anything if all the keys are already in snake case?

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
```diff
@@ -0,0 +1,60 @@
...
+	completeLiveRecording(liveRecordingId: string): Promise<ProcessRecordingResponse> {
+		return this.call<ProcessRecordingResponse>('POST', `/${liveRecordingId}/complete`);
+	}
+
+	private basePath(): string {
+		return `/api/conversation/${this.options.getConversationId()}/events/${this.options.getEventId()}/audio_recordings/live`;
+	}
+
+	private async call<T>(method: string, path: string, body?: unknown): Promise<T> {
+		const response = await fetch(`${this.basePath()}${path}`, {
```
alpinegeorge: Replace fetch with API client use

api/src/models/live_audio_recording.rs
rosswilson2306: As I read through more of this I guess it's to do with deserializing the jsonb column? Maybe there's a reason this wouldn't work but elsewhere we have implemented some sqlx traits on the column structs instead of creating two separate structs for the models, see https://github.com/crownshy/comhairle/blob/1f6c293105693d308ae4c48c182f1f9bb48aaa8a/api/src/models/email_template_config.rs?plain=1#L72-L93. 

We should maybe pick a single approach of handling jsonb columns?

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
ADSmith-0, line 1: For this would we be able to use the `apiClient`? Or would that not work?

api/src/models/live_audio_recording.rs
```diff
@@ -0,0 +1,309 @@
...
+    row.try_into()
+}
+
+pub async fn delete(db: &PgPool, live_audio_recording_id: Uuid) -> Result<(), ComhairleError> {
+    let (sql, values) = Query::delete()
+        .from_table(RawLiveAudioRecordingIden::Table)
+        .and_where(Expr::col(RawLiveAudioRecordingIden::Id).eq(live_audio_recording_id))
+        .build_sqlx(PostgresQueryBuilder);
+
+    sqlx::query_with(&sql, values).execute(db).await?;
```
rosswilson2306: @alpinegeorge Elsewhere we typically return the object in our delete methods. So maybe we should do that here so that the delete endpoints are consistent? Or change the other endpoints to not return anything, what do you think?

api/src/models/live_audio_recording.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    let current = get_by_id(db, live_audio_recording_id).await?;
+    if current.locked {
+        return Err(ComhairleError::LiveAudioRecordingLocked);
+    }
+
+    Err(not_found())
+}
+
+pub async fn unlock_live_recording(
```
rosswilson2306: @alpinegeorge Could you stick `#[instrument(err(Debug)]` on these methods?

api/src/routes/audio_recordings.rs
```diff
@@ -86,9 +111,10 @@ async fn create_recording(
 async fn list_recordings(
     State(state): State<Arc<ComhairleState>>,
     Path((_conversation_id, event_id)): Path<(Uuid, Uuid)>,
-    RequiredAdminUser(_user): RequiredAdminUser,
+    RequiredUser(user): RequiredUser,
 ) -> Result<(StatusCode, Json<Vec<AudioRecordingDto>>), ComhairleError> {
-    let recordings = audio_recording::list_by_event(&state.db, &event_id).await?;
+    let _ = ensure_live_recording_access(&state, event_id, &user).await?;
```
rosswilson2306: @alpinegeorge Just out of curiosity, who needs access to these who isn't an admin user?

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
alpinegeorge: Yes, I think we should use it. I was using fetch since I seemed to be having issues on my laptop. Trying again now.

api/src/websockets/handlers/audio_recording.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+        Ok(())
+    }
+
+    async fn on_connected(
+        &self,
+        connection: &WebSocketConnection,
+        _state: &Arc<ComhairleState>,
+    ) -> Result<(), WebsocketError> {
+        debug!(
```
rosswilson2306: @alpinegeorge Is this required? If so, could we use `tracing::info!` instead of `debug!`?

api/src/websockets/handlers/audio_recording.rs
```diff
@@ -0,0 +1,1083 @@
...
+            // disconnecting in most cases.
+            let _ = live_audio_recording::unlock_live_recording(
+                &_state.db,
+                live_connection_info.live_recording_id,
+                connection.user.id,
+            )
+            .await;
+        }
+
+        debug!(
```
rosswilson2306: @alpinegeorge Same here, is this required?

ui/packages/comhairle/src/lib/components/recordings/LiveRecorder.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+	let controller = $state<LiveRecorderController | null>(null);
+
+	function currentContext() {
+		return {
+			conversationId: conversation_id,
+			eventId: event_id,
+			recordings,
+			onComplete,
+			mode
```
rosswilson2306: @alpinegeorge Is any of this data expected to change across the lifecycle of the component? If so it would probably be better to use
```
let currentContext = $derived({
  ...
});
```
I'm not sure that what you have here will update if the data changes in svelte. I could be wrong though, I'm no svelte expert.

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
```diff
@@ -0,0 +1,60 @@
...
+			let message = `${method} ${path} failed (${response.status})`;
+			const parseResult = await tryCatchAsync(() => response.json());
+			if (parseResult.err === null) {
+				if (typeof parseResult.ok?.message === 'string') message = parseResult.ok.message;
+				else if (typeof parseResult.ok?.err === 'string') message = parseResult.ok.err;
+			}
+			throw new Error(message);
+		}
+
+		return response.json() as Promise<T>;
```
rosswilson2306: @alpinegeorge If these requests are all sending JSON and returning JSON responses could you use the `apiClient` from `@crownshy/api-client` instead of building out the custom functionality?

ui/packages/comhairle/src/lib/components/recordings/liveRecorderController.svelte.ts
```diff
@@ -0,0 +1,432 @@
...
+		}
+		this.liveRecordings = this.liveRecordings.filter(
+			(recording) => recording.id !== liveRecordingId
+		);
+		notifications.send({
+			message: completeResult.ok.message || 'Recording saved — transcription started',
+			priority: 'SUCCESS'
+		});
+		await this.onComplete?.();
+		void invalidateAll();
```
rosswilson2306: @alpinegeorge We're trying to move away from the blanket `invalidateAll()` calls now (we have it in so many places however) and instead using `invalidate('<indentifier>')`. See https://github.com/crownshy/comhairle/blob/1f6c293105693d308ae4c48c182f1f9bb48aaa8a/ui/packages/comhairle/src/routes/(admin)/admin/conversations/%5bconversation_id%5d/configure/+page.svelte?plain=1#L366 and https://github.com/crownshy/comhairle/blob/1f6c293105693d308ae4c48c182f1f9bb48aaa8a/ui/packages/comhairle/src/routes/(admin)/admin/conversations/%5bconversation_id%5d/+layout.ts?plain=1#L37 for how to set it up. It just means that you only refresh targeted requests instead of absolutely everything.

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
alpinegeorge: I just hadn't re-generated api.ts. That should be this fixed now.

api/src/bulk_storage_service/s3_storage.rs
alpinegeorge: Nope! Good spot!

api/src/models/live_audio_recording.rs
alpinegeorge: Implementing the trait sounds like a good approach. I'll use that.

api/src/models/live_audio_recording.rs
alpinegeorge: Yeah, returning is a good pattern. I'll fix this.

api/src/models/live_audio_recording.rs
alpinegeorge: Happily

api/src/routes/audio_recordings.rs
alpinegeorge: As said in the demo, this will be for any attendee wanting to record a breakout room.

api/src/websockets/handlers/audio_recording.rs
alpinegeorge: This was just for my own debugging, but may be useful as info later. I'll make the switch and if it's too noisy, we can remove it.

api/src/websockets/handlers/audio_recording.rs
alpinegeorge: For debugging just now, though I'll switch to info.

ui/packages/comhairle/src/lib/components/recordings/LiveRecorder.svelte
alpinegeorge: I think you're more of a svelte expert than I am, so I'll give it a shot.

ui/packages/comhairle/src/lib/components/recordings/liveRecorderApi.ts
alpinegeorge: Already transitioned in the latest fixup commit

ui/packages/comhairle/src/lib/components/recordings/liveRecorderController.svelte.ts
alpinegeorge: Thanks! I'll have a go at using that instead. I was thinking invalidating everything seemed a bit overkill.

alpinegeorge: I've ended up manually implementing the sqlx::prelude::FromRow trait and it works great.

rosswilson2306: Interesting.

ui/packages/comhairle/src/lib/components/recordings/LiveRecorder.svelte
alpinegeorge: It worked :)

PR #704, opened 2026-07-22
669/thinking space reporting

rosswilson2306:
Actions:

- insights endpoint for thinking_space tool, which returns organised data required for insights admin page
- insights admin UI for thinking_space
- migration to store ai_generated_summary if user edits thinking_space_summary to show comparison
- model method updates required for insights

Motivation:

- reporting and thinking_space insights

### DIFF
--- a/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
+++ b/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
@@ -1,0 +1,26 @@
+import type { PolisStatementAux } from '@crownshy/api-client/api';
+import type { PolisReportData } from '$lib/tools/polis/reportTypes';
+
+export async function polisInsightsLoader(apiClient: any, workflowStepId: string) {
+	let statementAux: PolisStatementAux[] = [];
+	let reportData: PolisReportData | null = null;
+	try {
+		statementAux = await apiClient.PolisListStatementAux({
+			queries: { workflow_step_id: workflowStepId }
+		});
+	} catch (e) {
+		console.error('Failed to load Polis statement aux', e);
+	}
+	try {
+		// PolisReportData is WikiPollReport (the client's return type) plus the client-only
+		// theme overlay, so this assigns without a cast. Fails (→ null) when the poll has no
+		// votes yet.
+		reportData = await apiClient.PolisGetReportData({
+			queries: { workflow_step_id: workflowStepId }
+		});
+	} catch (e) {
+		console.error('Failed to load Polis report data', e);
+	}
+
+	return { polis: { statementAux, reportData } };
+}
--- a/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
+++ b/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
@@ -1,26 +1,15 @@
-import type { PolisStatementAux } from '@crownshy/api-client/api';
-import type { PolisReportData } from '$lib/tools/polis/reportTypes';
+import type { ApiClient } from '@crownshy/api-client/api';
 
-export async function polisInsightsLoader(apiClient: any, workflowStepId: string) {
-	let statementAux: PolisStatementAux[] = [];
-	let reportData: PolisReportData | null = null;
-	try {
-		statementAux = await apiClient.PolisListStatementAux({
-			queries: { workflow_step_id: workflowStepId }
-		});
-	} catch (e) {
-		console.error('Failed to load Polis statement aux', e);
-	}
-	try {
-		// PolisReportData is WikiPollReport (the client's return type) plus the client-only
-		// theme overlay, so this assigns without a cast. Fails (→ null) when the poll has no
-		// votes yet.
-		reportData = await apiClient.PolisGetReportData({
-			queries: { workflow_step_id: workflowStepId }
-		});
-	} catch (e) {
-		console.error('Failed to load Polis report data', e);
-	}
+export async function polisInsightsLoader(apiClient: ApiClient, workflowStepId: string) {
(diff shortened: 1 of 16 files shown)


ui/packages/comhairle/src/lib/reports/thinking-space/ThinkingSpaceInsights.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+	let numParticipants = $derived(insights.users.length);
+	let averageNumFollowUps = $derived.by(() => {
+		const sumFollowups = insights.users.reduce(
+			(acc, user) =>
+				acc + user.answers.reduce((acc, answer) => acc + answer.followUps.length, 0),
+			0
+		);
+
+		return sumFollowups;
```
dtodorowa: tiny one: the card says "Average follow-up questions" but this returns the total (never divided by participants) 

unless I misunderstand something which is also very berry probable

rosswilson2306: Ah! Great spot! Thanks.

ui/packages/comhairle/src/lib/reports/thinking-space/SummaryTable.svelte
```diff
@@ -0,0 +1,6 @@
+<script lang="ts">
+	type Props = {
+		insights: ThinkingSpaceUserInsights;
+	};
+	let { summaries }: Props = $props();
```
dtodorowa: leftover stub? props type says insights but it destructures summaries and nothing imports it. nuke nuke-y? 💣

ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function polisInsightsLoader(apiClient: any, workflowStepId: string) {
+	let statementAux: PolisStatementAux[] = [];
+	let reportData: PolisReportData | null = null;
+	try {
+		statementAux = await apiClient.PolisListStatementAux({
+			queries: { workflow_step_id: workflowStepId }
+		});
+	} catch (e) {
+		console.error('Failed to load Polis statement aux', e);
+	}
```
dtodorowa: quoting Adam and checking if we wanna use `tryCatchAsync` tehee

ui/packages/comhairle/src/lib/reports/thinking-space/InsightsTable.svelte
```diff
@@ -0,0 +1,160 @@
...
+							{#if showAiSourceSummary}
+								Viewing original AI summary
+							{:else}
+								Edited by participant
+							{/if}
+						</span>
+						<Switch
+							bind:checked={showAiSourceSummary}
+							onCheckedChange={(v) => (showAiSourceSummary = v)}
+						/>
```
dtodorowa: minor: `bind:checked` already does the two-way, so the `onCheckedChange` is doing the same job twice (i think!)

also spotted an empty `<p class="text-primary"></p>` a bit further down that can go?

dtodorowa:
Random question: summaries that were edited before this migration won't have the original AI snapshot, right? So the toggle won't show them? Sounds like correct behaviour, but noting this for myself, if we are expecting to look at the Maine data at some point and wondering what's going on

dtodorowa (dismissed):

ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
rosswilson2306: To be honest, I think the try catches can go from here as all the insights-loader functions will be called from a single `try / catch`

dtodorowa:
- Think the cards in the design have a blue hover state that we are missing 
- and also the right side - question/responses - text is blue
- `Click on a row to see the full summary and responses associated with it` should use a different color

image
image

dtodorowa:
suuper tiny, but in the design the top text is different from the bottom one 20px/16px text-sm font-normal / text-xs font medium
image

dtodorowa:
this is a pet peeve - the ui jumping (extra row added) when we toggle, but this is probably a design question. maybe no action required... would just treat as me venting 🤡 

https://github.com/user-attachments/assets/36070cae-e016-40eb-a029-a03812f348ee

dtodorowa:
would make the dialog a bit taller, maybe 70-80vh? i feel like we can't see stuff properly there 
image

dtodorowa:
also, noticed the dialog has a "close" button at the top right, but I don't actually mind to not have it, because we don't have it in any other dialog in the codebase + you can tap outside or click esc to leave? 

so all good i think

image

dtodorowa (dismissed):

ui/packages/comhairle/src/lib/reports/thinking-space/SummaryTable.svelte
rosswilson2306: Yeah, this was just vim buffers not closing files I thought I had removed.

rosswilson2306:
> Random question: summaries that were edited before this migration won't have the original AI snapshot, right? So the toggle won't show them? Sounds like correct behaviour, but noting this for myself, if we are expecting to look at the Maine data at some point and wondering what's going on

@dtodorowa Correct. There's no way to know whether a user edited their summary up until now.

rosswilson2306:
> this is a pet peeve - the ui jumping (extra row added) when we toggle, but this is probably a design question. maybe no action required... would just treat as me venting 🤡

@dtodorowa Yeah, it is a little jumpy. I'll add an animation or something on the icon.

dtodorowa (approved):

PR #707, opened 2026-07-22
[Media library] User should be able to upload additional data with the media

### LINKED ISSUE #571: [Media library] User should be able to upload additional data with the media
### Description
Additional data can help with things like accessibility

### Acceptance criteria
- [x] Media should allow for a different `name` to be supplied
- [x] Media should allow for `alt` text to be defined on upload

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/571

### Changes
- Added details dialog to view the metadata in more detail
- Added the ability to modify this metadata
- Updated learning-assistant [`+page.ts`](https://github.com/crownshy/comhairle/pull/707/changes#diff-c24c4f96454ef9bc05f9860ec17e38ecf5ac2f2e65d0641c7a280ec92332292e), as it will be quicker to load
- Split the `FileUpload` from knowledge base into a separate component [`FileInput`](https://github.com/crownshy/comhairle/pull/707/changes#diff-cf4ac3c3829ca5d3da12b3c5639533cce6b1ba174a9a3d5b24ec5e86fdb3d6a4) so it could be reused for the upload dialog
- [Easyforms](https://github.com/crownshy/comhairle/pull/707#issuecomment-5134209786)
- Added [`HttpStatus`](https://github.com/crownshy/comhairle/pull/707/changes#diff-26bc5cca47aecd01bc18af57d045f1bd465ae316a672ec7821eaf257a549fe26) to mirror the backend
- Added [units](https://github.com/crownshy/comhairle/pull/707/changes#diff-75a7352676137c784ca802480da3dc5a1c6be847bb98ac4d15fb41a7c80bc948) to help explain common magic numbers

### Testing
<details>
<summary>Should be able to upload and view metadata</summary>

https://github.com/user-attachments/assets/1b3534d7-8ce3-4f87-9ab4-fe8951062161

</details>

<details>
<summary>Should be able to edit metadata</summary>

https://github.com/user-attachments/assets/95ea449e-c703-418b-a825-a2f16c80f94e

</details>

<details>
<summary>Should be able to drag and drop files to upload</summary>

https://github.com/user-attachments/assets/b12d7aef-5fc2-459a-9e26-4e52b9d096e6

</details>

<details>
<summary>Should prevent the user from uploading media without alt text</summary>

https://github.com/user-attachments/assets/1cfc6db6-dcd6-405d-8d03-d0d35ce04548

</details>

<details>
<summary>Should prevent the user from uploading media without a name</summary>

https://github.com/user-attachments/assets/e9d4d8e3-77b1-469e-bb32-42dd4a14bc5e

</details>

<details>
<summary>Should error when a broken file is uploaded</summary>

https://github.com/user-attachments/assets/86278200-5b9e-4660-abdf-34ad69f00802

</details>

<details>
<summary>Should error when a file with an incorrect extension is uploaded</summary>

https://github.com/user-attachments/assets/1a8ac6f4-9924-4a59-8582-23256530331f

</details>

<details>
<summary>Should error when a max file size is exceeded</summary>

https://github.com/user-attachments/assets/417a21cc-d52f-4d9c-9de3-bc0e2a7c25db

</details>

<details>
<summary>Should be able to delete the file to upload another</summary>

https://github.com/user-attachments/assets/c6efba37-5871-42e5-a171-9a3dd57f3d3f

</details>


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/+layout.ts
```diff
@@ -11,7 +11,6 @@ import type {
 	WorkflowStepWithTranslations
 } from '@crownshy/api-client/api';
 import type { LayoutLoad } from './$types';
-import type { ConfigureTab } from './configure/tabs';
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/+layout.ts:14]: My editor flagged that this was unused

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.server.ts
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.server.ts:1]: I was changing over the `FileUpload` in the learning assistant and I saw the `+page.ts`:
```ts
import { apiClient } from '@crownshy/api-client/client';
import type { PageLoad } from './$types';
export const load: PageLoad = async ({ parent }) => {
	const { conversation } = await parent();
	let documents;
	try {
		documents = await apiClient.ListDocuments({
			params: { conversation_id: conversation.id }
		});
	} catch (e) {
		console.error(e);
	}
	return { documents, conversation };
};
```
I hope it's alright I changed this a bit too, basically with `+page.ts` it would waterfall, so we'd have to wait for the previous requests to finish before we started this one, but with `+page.server.ts` it will fire straight away so should be faster hopefully

For the conversation id I just pulled it out from the request parameters, and then the conversations object gets pulled through via data anyway because it's loaded from `+layout.server.ts` so all child pages will have access to it automatically

ui/packages/comhairle/src/lib/components/FileInput.svelte
ADSmith-0, line 1: This is the `FileUpload` component from the `learning-assistant` broken out into a reusable component

ui/packages/comhairle/src/lib/components/FileInput.stories.svelte
ADSmith-0, line 1: I tried to create some stories for the test cases, but I wasn't super sure how it would work in this case, as it's about user input with html elements? So I've left them as stubs currently but hopefully can go back to them at some point and finish them

ui/packages/comhairle/src/lib/components/ui/input/Input.stories.svelte
```diff
@@ -18,8 +18,7 @@
 </script>
 
 <Story name="Default">
-	<div class="w-[300px]">
```
ADSmith-0, line 21: The linter flagged saying to switch it from this class to the other

ADSmith-0:
One thing I wanted to mention for this PR is the form solution I went for, basically I was having a lot of trouble getting superforms to work as expected, I'm not sure if I was using it wrong, but I spent about a day going through the `FormSnap`, `SuperForms` and `Zod` docs trying to figure out why the errors and behaviour were not working correctly

I ended up using a simple form so I could continue with the dev, with the intention of circling back to the SuperForm issues, but I thought of a way we might be able to extend simple forms in a way that makes it very easy to build them in the future

I called it `EasyForms` and it's inside of the components directory for this PR, to briefly explain how it works, HTML and JS already have a lot of inbuilt validation for forms which is really good, they have automatic error messages that are localised, multiple validation types, and a pretty good API:
image

There are also a number of methods that are super helpful:
- [`checkValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/checkValidity) - Checks whether an input is valid or not
- [`reportValidity()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/reportValidity) - Checks whether an input is valid or not, and if it isn't, flags the errors
- [`setCustomValidity`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity) - Lets you set a custom error message

This is all coming from the [HTML validation related attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Constraint_validation#validation-related_attributes), effectively for any input you can give a number of attributes to check if the field has been left empty, the input is too short, it doesn't match a regex pattern, etc

So the idea was to create validation around these attributes, with 3 parts
- Schema
- Frontend validation
- Backend validation

**Schema**
The schema is the validation that will work across the Frontend and Backend, it's key-value pairs for the HTML validation attributes (plus some extra for file types because their validation is a special case), e.g. The [MediaSchema](https://github.com/crownshy/comhairle/blob/29277357f0059696f6b036c12c8a84658eafdc2e/ui/packages/comhairle/src/lib/components/Media/schema.ts)
```ts
import { typedObj } from '$lib/utils/types';
import { MB } from '$lib/utils/units';
import { type FileAttr, type InputAttr } from '../EasyForm/types';

export const MediaSchema = {
	media: typedObj<FileAttr>({
		name: 'media',
		required: true,
		accept: '.png,.jpg',
		maxSize: 50 * MB
	}),
	name: typedObj<InputAttr>({
		name: 'name',
		required: true
	}),
	alt: typedObj<InputAttr>({
		name: 'alt',
		required: true
	})
} as const;

export default MediaSchema;
```

**Frontend validation**
To use it on the Frontend then it's as simple as spreading it into the input, and because it's HTML validation it will work for any HTML `input`, e.g.
```html
<FileInput
	{...MediaSchema.media}
	onfile={(newFile) => (file = newFile)}
	class={file !== null ? 'hidden' : ''}
/>
<Input
	{...MediaSchema.name}
	label="Filename"
	type="text"
	value={file?.name ? Media.getFilename(file.name) : ''}
/>
<Input {...MediaSchema.alt} label="Alt" type="text" />
```

**Backend validation**
This is a validate function [found here](https://github.com/crownshy/comhairle/blob/29277357f0059696f6b036c12c8a84658eafdc2e/ui/packages/comhairle/src/lib/components/EasyForm/validate.ts), that just takes the schema and the `formData` and runs a check against it to make sure that it's valid, in the same way the HTML would, e.g. [from here](https://github.com/crownshy/comhairle/blob/29277357f0059696f6b036c12c8a84658eafdc2e/ui/packages/comhairle/src/routes/(admin)/admin/media-library/%2Bpage.server.ts#L29)
```ts
export const actions = {
  upload: async ({ request }: RequestEvent) => {
    const data = await request.formData();
    const form = validate(data, MediaSchema);

    console.log('form:', form); // Returns a `Result` ({ ok: boolean, err: ValidationError });
  }
}
```

I've written [unit tests](https://github.com/crownshy/comhairle/blob/29277357f0059696f6b036c12c8a84658eafdc2e/ui/packages/comhairle/src/lib/components/EasyForm/validate.spec.ts) to try and make sure that all of the validation is correct 

**Comparison**
The difference should hopefully be roughly the same functionality for less overhead, making the same form with both libraries:

*SuperForms*

```html
<script lang="ts">
	const form = superForm(
		{
			media: [],
			name: '',
			alt: ''
		},
		{
			validators: zodClient(mediaSchema),
			taintedMessage: false,
			validationMethod: 'onsubmit'
		}
	);

	const { form: formData, validateForm, errors } = form;
</script>

<form method="POST" action="/admin/media-library?/upload" enctype="multipart/form-data">
	<Form.Field {form} name="media" class="space-y-2">
		<Form.Control>
			{#snippet children({ props })}
				<Form.Label>File</Form.Label>
				<Input
					{...props}
					type="file"
					bind:value={$formData.media}
				/>
				<Form.FieldErrors />
			{/snippet}
		</Form.Control>
	</Form.Field>
	<Form.Field {form} name="name" class="space-y-2">
		<Form.Control>
			{#snippet children({ props })}
				<Form.Label>Filename</Form.Label>
				<Input
					{...props}
					bind:value={$formData.name}
				/>
				<Form.FieldErrors />
			{/snippet}
		</Form.Control>
	</Form.Field>
	<Form.Field {form} name="alt" class="space-y-2">
		<Form.Control>
			{#snippet children({ props })}
				<Form.Label>Alt</Form.Label>
				<Input
					{...props}
					bind:value={$formData.alt}
				/>
				<Form.FieldErrors />
			{/snippet}
		</Form.Control>
	</Form.Field>

	<div class="flex justify-end">
		<Button type="submit" disabled={saving}>
			{saving ? 'Saving...' : 'Save Changes'}
		</Button>
	</div>
</form>
```

*EasyForms*
```html
<Form method="POST" action="/admin/media-library?/upload" enctype="multipart/form-data">
	<Input type="file" label="Media" {...MediaSchema.media} />
	<Input label="Filename" {...MediaSchema.name} />
	<Input label="Alt" {...MediaSchema.alt} />
	<Submit class="mt-7 self-end" text="Upload" />
</Form>
```

**Behaviour**

*`onBlur`*

https://github.com/user-attachments/assets/5d584cb5-a87f-422d-9edc-29748378d9e5

*Custom errors*

https://github.com/user-attachments/assets/9c1f6f30-a091-4e53-89f5-123f59868ff3

*Trying to submit*

https://github.com/user-attachments/assets/a1b3c549-b88a-45d7-88af-490a9f832ff2


**Drawbacks**
To do more complex validation such as checking two passwords are the same or something like that, would required more effort, something like this:
Screenshot from 2026-07-31 11-19-22

Requires this:
```html
<!-- +page.svelte -->
<script lang="ts">
	import { Form, Input, Submit } from '$lib/components/EasyForm';

	let emailInput = $state<Input | null>(null);
	let emailConfirmInput = $state<Input | null>(null);

	function checkNotEqual() {
		if (!emailInput?.getValue() || !emailConfirmInput?.getValue()) {
			return;
		}

		if (emailInput.getValue() !== emailConfirmInput.getValue()) {
			emailInput?.setError('Emails do not match');
			emailConfirmInput?.setError('Emails do not match');
			return;
		}

		emailInput?.clearError();
		emailConfirmInput?.clearError();
	}
</script>

<Form
	method="POST"
	class="flex flex-col gap-5 pt-6"
	onsubmit={() => {
		checkNotEqual();
	}}
>
	<Input
		required
		label="Email"
		bind:this={emailInput}
		type="email"
		name="email"
		onblur={checkNotEqual}
	/>
	<Input
		required
		label="Confirm email"
		bind:this={emailConfirmInput}
		type="email"
		name="confirm-email"
		onblur={checkNotEqual}
	/>
	<Submit text="Confirm" />
</Form>
```
```ts
// +page.ts
export const actions = {
	upload: async ({ request }: RequestEvent) => {
		const data = await request.formData();
		const form = validate(data, MediaSchema);

		// Check on the backend too
		if (data.get("email") !== data.get("confirm-email") {
			return fail(...);
		}

		console.log('form:', form);
	}
}
```

However generally it does seem that validation like this is more the exception than the rule, in which case I don't think it would be too much hassle?

But just wanted to hear people's thoughts on it?

I personally would probably prefer a simpler form system, but there's nothing to say we can use both - SuperForms for more complex forms and EasyForms for the simpler ones? Or we can just use one or the other? What does everyone think?

rosswilson2306:
@ADSmith-0 What was it that wasn't working with Superform? What were the errors occurring and behaviour you wanted that wasn't working?

ADSmith-0:
@rosswilson2306

It was just more like little inconsistencies that I couldn't figure out how to fix, like I think the validation was working on blur, but then it wasn't working when I tried to submit? It was just allowing me to submit the form with no values in it even though in the schema I had specified all fields were required

Also I think that the file field especially, just wasn't having any validation applied, I tried to specify it based on what I could see in the docs, but I couldn't get any of the errors to show up

Then I seem to remember as well I wasn't getting any errors at all in the beginning (when I should of been), and I think Daniela and I figured out it's because you're meant to load the form in first? Like here https://github.com/crownshy/comhairle/blob/staging/ui/packages/comhairle/src/routes/(admin)/admin/conversations/%5Bconversation_id%5D/events/new/%2Bpage.ts

Although it looks like we're sometimes doing it and sometimes not? But I think that fixed the issue

But yeah overall it was just not being able to get the validation to trigger properly, so I'm not sure if I was doing something wrong or what was happening with it all

api/src/routes/conversations.rs
```diff
@@ -1568,7 +1568,9 @@ mod tests {
     }
 
     #[sqlx::test(migrator = "crate::SQLX_MIGRATOR")]
-    fn should_be_able_to_get_conversation_with_media(pool: PgPool) -> Result<(), Box<dyn Error>> {
+    async fn should_be_able_to_get_conversation_with_media(
```
ADSmith-0, line 1571: My linter was flagging that await was being used inside of a non-async environment, so I added it here, but I'm not sure if it's needed or not?

api/src/routes/media.rs
```diff
@@ -55,82 +55,97 @@ async fn upload(
...
         .as_ref()
         .ok_or(ComhairleError::NoBulkStorageServiceConfigured)?;
 
-    let mut files = vec![];
+    let mut upload_media_form = UploadMediaForm::new();
+
     while let Some(field) = form_data.next_field().await? {
-        let content_type = field.content_type().map(|ct| ct.to_string());
+        let content_type = field.content_type().map(|f| f.to_string());
+        let field_name = field.name().map(|n| n.to_string());
```
ADSmith-0, line 70: Weird Rust ownership stuff, apparently just cloning doesn't work? But I'm not super familiar with it all, so please let me know if there's a better way of doing it!

ui/packages/comhairle/src/lib/utils/constants.ts
ADSmith-0, line 1: Wanted to see what people thought of this? Might just be me but I sometimes find the numbers hard to remember what does what, and I know that on the API side we have this naming convention with stuff like `StatusCode::CREATED`, thought it might be helpful to have it this end as well? Happy to have it screaming snake case or something like that, if it's preferable!

ui/packages/comhairle/src/lib/utils/units.ts
ADSmith-0, line 1: Added these in too I thought they might be helpful going forward, it gives a bit more context when you're trying to work out numbers

Also just to note I'm times-ing by `1000` instead of `1024`, because apparently there's a `1000` variation (byte) and a `1024` variation (bibyte)

So:
1000B (Kilobyte KB), 1024B (Kibibyte KiB),
1000KB (Megabyte MB), 1024KiB (Mebibyte MiB),
1000MB (Gigabyte GB) 1024MiB (Gibibyte GiB)

Had no idea this was a thing, but I think most operating systems these days use normal bytes not the bibyte variant, in which case I think these conversions makes more sense? Open to ideas though this is something I'm not super familiar with - I always thought it was 1024 by default and there weren't any variations 😅

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/+page.svelte:1]: Had to change this over because we're not doing multiple file uploads anymore (as it doesn't work with adding in the name and alt)

api/src/models/media.rs
```diff
@@ -20,6 +20,43 @@ use crate::{
...
+            name: "".to_string(),
+            alt: "".to_string(),
+        }
+    }
+    pub fn update_field(&mut self, field: &str, value: String) {
+        match field {
+            "name" => self.name = value,
+            "alt" => self.alt = value,
+            &_ => (),
+        }
```
rosswilson2306: @ADSmith-0 I'm not very far through this PR but as you can only use `update_field` to update the `name` and `alt` fields is it a useful method? You initialise it with empty data, so I would imagine you would need to add the file by doing:
```rust
let mut form = UploadMediaForm::new()
form.file = MyFile;
```
rather than using the `update_field` method. So maybe just do the same with the `name` and `alt` fields so it's consistent and get rid of `update_field`?

api/src/models/media.rs
```diff
@@ -130,14 +166,30 @@ impl MediaContentType {
...
         }
     }
+    pub fn get_type(file: &File) -> Result<Self, ComhairleError> {
+        let mime = MediaContentType::try_from_mime(&file.content_type);
+        if mime.is_ok() {
+            return mime;
+        }
+
+        let extension: Vec<&str> = file.filename.rsplit('.').collect();
+        let Some(extension) = extension.get(0) else {
```
rosswilson2306: @ADSmith-0 Rust has a `path` module that might be a bit more reliable for working with extensions that this. See this playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0c98da5b10b5dea5d405aa640cc14496

api/src/routes/documents.rs
```diff
@@ -328,7 +328,8 @@ pub fn router(state: Arc<ComhairleState>) -> ApiRouter {
         .api_route(
             "/",
             post_with(upload, |op| {
-                op.tag("Documents")
+                op.id("PostDocuments")
```
rosswilson2306, line 331: @ADSmith-0 I had left out the id from this so that the api client didn't get a generated method for the frontend to use, as sending multiform data didn't work well with the client generator and we had to just use a raw `fetch` request. Were you able to get this working sending multiform data using `apiClient.PostDocuments`?

api/src/routes/media.rs
```diff
@@ -55,82 +55,97 @@ async fn upload(
...
+
+        if let None = content_type {
+            let Some(field_name) = field_name else {
+                continue;
+            };
+            let text = field.text().await.map(|t| t.to_string());
+            let Ok(text) = text else {
+                continue;
+            };
```
rosswilson2306, line 79: It might a little more idiomatic to do:
```
if content_type.is_none() {
```
and then do
```
let Ok(text) = field.text().await.map(str::to_owned) else {
  continue;
}
```
in a oner. But doesn't matter too much.

api/src/routes/media.rs
```diff
@@ -55,82 +55,97 @@ async fn upload(
...
             .file_name()
             .map(|f| f.to_string())
             .unwrap_or_else(gen_id);
         let bytes = field.bytes().await?.to_vec();
 
-        files.push((filename, bytes, content_type));
+        upload_media_form.file = File {
+            filename,
+            bytes,
+            content_type: content_type.unwrap_or("".to_string()),
```
rosswilson2306: You can do `.unwrap_or_default()` here (default for a String is an empty string) but again just a suggestion, doesn't matter too much. Just something the linter might suggest and keep down warnings.

api/src/routes/media.rs
```diff
@@ -55,82 +55,97 @@ async fn upload(
...
 
-    let mut files = vec![];
+    let mut upload_media_form = UploadMediaForm::new();
+
     while let Some(field) = form_data.next_field().await? {
-        let content_type = field.content_type().map(|ct| ct.to_string());
+        let content_type = field.content_type().map(|f| f.to_string());
+        let field_name = field.name().map(|n| n.to_string());
+
+        if let None = content_type {
```
rosswilson2306: I might be missing something but I don't see what this check is for. It checks whether the content_type is `None` but then doesn't do anything to extract the content_type and further down you return an error if the content_type isn't present. Have I missed something or would it be better to just bail out earlier if there is no content_type and there is no way to create it elsewhere?

ui/packages/comhairle/src/lib/utils/constants.ts
```diff
@@ -0,0 +1,10 @@
+export enum HttpStatus {
+	TemporaryRedirect = 307,
+	PermanentRedirect = 308,
+	BadRequest = 400,
+	Unauthoized = 401,
+	Forbidden = 403,
+	NotFound = 404,
+	UnprocessableContent = 422,
+	InternalServerError = 500
```
rosswilson2306, line 10: Maybe add `Conflict = 409` as I've found myself using that in a few places on the frontend when there is a conflict in the db and I want to notify with a specific message if that occurs.

ui/packages/comhairle/src/lib/utils/units.ts
rosswilson2306: I actually thought that KiB etc would have been the more common way, but looked it up and right enough. Learnt something new.

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.server.ts
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.server.ts:1]: We've tended to avoid using `+page.server.ts` because there was an idea of potentially creating a mobile app at some point and I think there is some way of wrapping up a svelte app as a mobile app but is dependent on things being able to work completely client side. 

I don't know much about it though and don't know if this is still an ambition.

@stuartlynn What's your thoughts?

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte
```diff
@@ -60,6 +69,68 @@
...
+		}
+
+		notifications.send({
+			message: 'File uploaded successfully',
+			priority: 'INFO'
+		});
+		await invalidate('knowledge-base:documents');
+	}
+
+	async function uploadFromUrl() {
```
Comment (rosswilson2306) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte]: Does this work now? For a while we've had this button that didn't work and I keep meaning to say to remove it but if it works now great!

api/src/models/media.rs
ADSmith-0: Yeah that's fair! I was thinking maybe if we had more string fields in the future that we wanted to change, like description or tags or something then it might be useful to have a function that handles that, but I guess it would pretty much be the same as just moving the match to the `upload` loop, so I'm happy to do that!

api/src/models/media.rs
ADSmith-0: Oh cool! Yeah I'll swap that over!

api/src/routes/documents.rs
ADSmith-0: Oh I see what you mean, yeah I did it more for discoverability because when I was trying to search through the `api.ts` to for `PostDocuments` I couldn't find it since this was it's current name

image

But I agree that with `PostDocuments` it sound like a method you can call when you wouldn't want to because it wouldn't work, so I guess it might be better to swap it back?

api/src/routes/media.rs
ADSmith-0: I'm using it to try and check whether the field is a string or a file, so I believe that if it's a file then the `content_type` is set, but if it's a string then it won't be, I guess you could use something like `file_name` instead for the check which might be a bit clearer? But I wasn't sure about `file_name` if it's always set, so I went with `content_type`

rosswilson2306: Right, I see. That makes sense. Maybe add a comment above just so it's clear what the check does?

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte
Comment (ADSmith-0) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte]: Oh yeah no it's not working unfortunately, I can remove it if we want?

ADSmith-0:
> We've tended to avoid using +page.server.ts because there was an idea of potentially creating a mobile app at some > point and I think there is some way of wrapping up a svelte app as a mobile app but is dependent on things being able to work completely client side.
>
> I don't know much about it though and don't know if this is still an ambition.
>
> @stuartlynn What's your thoughts?

Oh interesting! I didn't know about that, I think we'd be able to do a mobile app as a PWA which might be a bit easier? 

But yeah it wouldn't be an "app" app if that makes sense, but it would be installable

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte]: Yeah, I think just comment it out for now if it's not working. I would imagine most use cases would be uploading from a local machine.

ADSmith-0:
> > We've tended to avoid using +page.server.ts because there was an idea of potentially creating a mobile app at some > point and I think there is some way of wrapping up a svelte app as a mobile app but is dependent on things being able to work completely client side.
> > I don't know much about it though and don't know if this is still an ambition.
> > @stuartlynn What's your thoughts?
> 
> Oh interesting! I didn't know about that, I think we'd be able to do a mobile app as a PWA which might be a bit easier?
> 
> But yeah it wouldn't be an "app" app if that makes sense, but it would be installable

Just swapped this back to `+page.ts` after the conversation in the standup this morning

api/src/routes/media.rs
alpinegeorge: There's probably something you could do with `Cow::Borrowed`, but I don't think it's worth your time to do so

ui/packages/comhairle/src/lib/utils/constants.ts
alpinegeorge: I wouldn't add this, but keep this close by for reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status. If there's a widely available (ideally internal) library which handles this, we should use that instead.
https://github.com/prettymuchbryce/http-status-codes looks quite widely used.

stuartlynn (approved):

stuartlynn:
Merging this PR but probably needs some tidy up to remove dead code and look at getting supabase working with it

PR #715, opened 2026-07-23
Tiptap: Cell background colours

dtodorowa:
Builds on the table support (#714): lets you give table cells a background colour from a fixed palette.

### What you can do
- Caret in a table -> toolbar dropdown has a **Cell colour** row of swatches (none/gray/red/orange/yellow/green/blue/purple/pink)
- Pick one to colour the current cell, or drag-select several cells and colour them all at once
- The "/" swatch clears the colour

### Should be theme-safe 
Cells store a colour *key* (`data-cell-color="blue"`), not a raw hex. The CSS resolves each key with `color-mix(in srgb, <hue> 16%, var(--color-background))`, so it's a soft pastel in light mode and a muted tint in dark mode, readable in both. Nothing theme-specific is stored, and it adapts on its own.

Colours render in the editor and the read-only renderer alike: I disabled TableKit's built-in cell/header and swapped in colour-aware versions in the shared `getBaseExtensions`, so all render paths pick it up.

### Notes
- Pasted raw cell colours are still dropped (raw hex doesn't map cleanly to the theme keys), so re-apply from the palette.
- No data model change, it's just a new `cellColor` node attribute.
- To tweak: hexes live in `editor-content.css` + the `swatchBackground` map in `TableControlsPopover.svelte`, keys in `tableColors.ts`.

### Testing
Added an SSR test that a colour key renders `data-cell-color` in read-only output..
The live picker/dark-mode look wants a manual pass.
 
## Demos: 

Adding colors to cells: 

https://github.com/user-attachments/assets/e393d1fa-bd16-42ff-b03c-43211b251b1e

Pasting tables from google docs drops the color:

https://github.com/user-attachments/assets/fae0a27a-75dc-4e1c-b8d5-092a009d0572

But pasting our own tables, should work fine :)

https://github.com/user-attachments/assets/62bf50c0-6c60-4ecd-bc43-5bb09280858e

### DIFF
--- a/ui/packages/comhairle/src/lib/components/RichTextEditor/TableControlsPopover.svelte
+++ b/ui/packages/comhairle/src/lib/components/RichTextEditor/TableControlsPopover.svelte
@@ -9,6 +9,7 @@
 		tableDeleteAction,
 		type TableAction
 	} from './tableActions';
+	import { TABLE_CELL_COLORS, type TableCellColorOption } from './tableColors';
 
 	type Props = {
 		editor: Editor | undefined;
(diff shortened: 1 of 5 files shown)


rosswilson2306 (dismissed):

alpinegeorge (dismissed):
LGTM

PR #716, opened 2026-07-23
Ip logging

stuartlynn:
This PR does two things 

1. Adds IP extraction to the logging middleware so we see ip addresses in route spans. It additionally adds user_id to the spans if a user is logged in ("-") otherwise 

2. Associates the users IP at signup on the user model. This isn't exposed anywhere in the API or frontend and is just for internal use


rosswilson2306 (approved):

alpinegeorge (approved):
LGTM

PR #725, opened 2026-07-23
Fix proposal translations not saving + tidy up the editor dialogs

### LINKED ISSUE #722: Issue with translation tool in prioritisation tool (not saved)
https://github.com/user-attachments/assets/f4f1711d-3da8-4d95-a5a2-cff1a255f947

dtodorowa:
Editing proposal translations was broken in a few ways: text you typed would revert, AI translations vanished right after generating, and hitting Approve saved a blank. This sorts all of that out, plus some polish on the editor dialogs.

## The actual bugs

Two root causes, both around state:

1. **The proposals list wasn't refreshing after saves.** The translation fields reconcile by calling SvelteKit's `invalidateAll()`, but the prioritization tool keeps its own list that `invalidateAll()` doesn't touch. So every save reconciled against stale data and your edit snapped back to the old value. Same reason the AI translation disappeared (it cleared back to the empty stored value) and Approve then persisted that blank.
2. **The rich text editor echoed server content back as a fake edit.** `onChange` was firing from tiptap's `onTransaction`, which runs on every transaction, including the programmatic sync we use to push saved content back in. That created spurious saves and a feedback loop on top of bug 1.

## What changed

- Added a `refresh` hook to the translation source (defaults to `invalidateAll`); the prioritization store passes a silent `reload()` so saves reconcile against fresh data without flashing the list.
- Rich text editor now emits changes from `onUpdate` (which ignores programmatic content sync) instead of `onTransaction`.
- **Create proposal is now persist-first:** opening "New proposal" creates an empty draft up front, so create and edit are one flow and translation badges work immediately (they used to only show on edit). Untouched drafts get cleaned up on close, Cancel discards, Done keeps.
- **Unsaved-changes guard** on the proposal editor so a refresh / tab close / navigation mid-save warns you instead of silently dropping the edit.
- Skeleton cards for the proposals list (including first paint, so the "No proposals yet" empty state no longer flashes before the fetch), and the list no longer tears down / jumps scroll to the top when you close the editor.
- Both editor dialogs are now a consistent `90vh` x `80vw`, so the translate dialog fully covers the proposal editor behind it instead of floating short inside it.

Demo 

https://github.com/user-attachments/assets/f1cf8269-4604-4158-8727-594fc9993d1e

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { onDestroy } from 'svelte';
+	import { onMount } from 'svelte';
 	import TranslatableField from '$lib/components/Translation/TranslatableField.svelte';
 	import { createTextContentSource } from '$lib/components/Translation/translationSource.svelte';
 	import type { TranslationSource } from '$lib/components/Translation/translationUtils';
@@ -35,8 +35,10 @@
 		refresh
 	});
 
-	registerSource?.(section.id, source);
-	onDestroy(() => unregisterSource?.(section.id));
+	onMount(() => {
+		registerSource?.(section.id, source);
+		return () => unregisterSource?.(section.id);
+	});
 </script>
 
 <TranslatableField
--- a/ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
@@ -1,15 +1,29 @@
 <script lang="ts">
+	import { onDestroy } from 'svelte';
 	import TranslatableField from '$lib/components/Translation/TranslatableField.svelte';
 	import { createTextContentSource } from '$lib/components/Translation/translationSource.svelte';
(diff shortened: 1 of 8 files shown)


ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
```diff
@@ -1,24 +1,42 @@
...
 	// getter reads the live `section` prop, so a list refresh (new bodyTranslations) is picked up.
 	const source = createTextContentSource({
 		getTranslation: () => section.bodyTranslations,
 		getPrimaryLocale: () => primaryLocale,
-		getSupportedLanguages: () => supportedLocales
+		getSupportedLanguages: () => supportedLocales,
+		refresh
 	});
+
+	registerSource?.(section.id, source);
```
rosswilson2306: @dtodorowa Does it matter at all if this code is here vs inside `onMount()` ?

rosswilson2306 (commented):
Left a small question.

shuyanglin (dismissed):

ui/packages/comhairle/src/lib/tools/prioritization/components/ProposalSectionField.svelte
dtodorowa: functionally, i think not, but yeh, let's keep the lifecycle stuff together. I put it at init so it sat next to where the source is created, but it prob reads cleaner grouped in one place

Issue #736, opened 2026-07-23
Rename enableQaChatBot field to enable_learning_assistant (parked, needs migration)


dtodorowa:
Follow-up to #373. **Parked / low priority** - see below.

## Status
The user-facing rename shipped in #373: the tab, route (`/learning-assistant`), page, and toggle all read "Learning Assistant". The only thing left is the **internal** field name, which still says `enableQaChatBot` (FE) / `enable_qa_chat_bot` (DB + Rust).

Decided to **park this** for now.

## If/when we do it - scope
Backend:
- new migration: `ALTER TABLE conversation RENAME COLUMN enable_qa_chat_bot TO enable_learning_assistant`
- `api/src/models/conversation.rs` (struct field + the generated `ConversationIden::EnableQaChatBot` usages + `DEFAULT_COLUMNS`)
- `api/src/routes/conversations/dto.rs`
- regenerate `open-api-spec.json` (run the API with `--export-api-spec`)

Frontend (regenerate `@crownshy/api-client`, then update refs):
- `LearnUI.svelte`, `ConversationSupportSidebar.svelte`, the workflow `+layout.ts`, the step `+page.svelte`, the `learning-assistant/+page.svelte` toggle (the `enable_qa_chat_bot` UpdateConversation payload), `ConversationCard.stories.svelte`

## Notes 
Blast radius is tiny, so it's safe to do later as a single coordinated PR - just not worth doing now

PR #737, opened 2026-07-23
373/Hide Learning Assistant when the knowledge base is empty

### LINKED ISSUE #373: Show error state when no documents have been uploaded to the knowledge base
TODO: detect wrong state and display meaningful feedback to the user explaining that no document has been configured or parsed yet.

Acceptance Criteria:
- If no document has been uploaded/parsed, the Learning Assistant should display a clear message indicating this (e.g. "No document has been set up yet. Please upload and parse a document to get started.")
- The error state is visually distinct and actionable. not a raw error message
- Handles both "no document uploaded" and "document still parsing" as separate states if applicable

This is the bug:
Image

dtodorowa:
Closes #373

### The problem
When a conversation had the Learning Assistant enabled but no usable documents, the backend RAG call failed and streamed its error back as a normal answer, so participants saw a raw `**ERROR**: 'id'` bubble where an answer should be.

### The fix
Only show the Learning Assistant when the knowledge base has at least one fully parsed (`DONE`) document. "Nothing uploaded" and "uploaded but still parsing" both resolve to hidden, so there's no assistant to error out. No raw error.
### What changed

**Participant side**
- The "does the knowledge base have parsed docs" check is hoisted into the shared workflow `+layout.ts` (`hasKnowledgeBaseDocs`), so it's one source of truth consumed by both surfaces that render the assistant (`LearnUI` and the support sidebar). This also removes two duplicate client-side `ListDocuments` fetches that were doing the same thing.
- A failed doc fetch falls back to "no docs" and hides the assistant, rather than surfacing a backend error.

**Admin side**
- Renamed the "Knowledge base" tab to **Learning Assistant** (route is now `/learning-assistant`).
- Moved the enable toggle off Configure → Access and onto this tab, next to the documents that make it work. Removed it from Configure entirely.
- Added a warning callout under the toggle (using the Alert component): when it's on but no document has parsed yet, it tells the admin to upload a document or the assistant won't appear to participants.
- Reworked the page layout to match the Configure form style (label-left, control-right, divider rows).

### Notes
- ADR-0007 documents the gating decision and the one-shared-source enforcement.
- The internal field is still named `enableQaChatBot`; renaming it end to end needs a DB migration, so it's parked in #736 (the UI already reads "Learning Assistant", so it's cosmetic).
- Old `/knowledge-base` links now 404 (admin-only, internal, not worth a redirect, right?)

## Demo 
Alert: 
image

https://github.com/user-attachments/assets/7cac7bee-3ec8-4537-ab65-dc2cd698374c


alpinegeorge (approved):
Looks good.
Would it be worth adding a high-contrast note in the admin interface when there are no knowledge base docs uploaded?
I'm just thinking someone who's not paying attention might miss the current note and then get confused when it doesn't appear.

dtodorowa:
> Would it be worth adding a high-contrast note in the admin interface when there are no knowledge base docs uploaded? I'm just thinking someone who's not paying attention might miss the current note and then get confused when it doesn't appear.

Absolutely. will create a follow-up

PR #742, opened 2026-07-24
Live agenda push

stuartlynn:
Sets up a broadcast to room pattern for more easily broadcasting to video call users regardless of if they are on the same server or not. 

Uses  this to update the agenda when it changes across all rooms


alpinegeorge (approved):
LGTM

PR #759, opened 2026-07-27
758/Fix: HeyForm survey embed footer overlapping content on longer questions

### LINKED ISSUE #758: HeyForm survey embed: bottom buttons overlap content on longer questions (participant/public side)
## Summary

On the participant/public side, longer HeyForm surveys render with the form's bottom navigation buttons (Next / Skip and the nav toolbar) overlapping the question content. The participant has to scroll down inside the embed to un-overlap them before they can answer. Short forms are fine; only questions tall enough to exceed the embed's capped height are affected.

Image

## Where

`ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte` (the public embed used at
`routes/(public)/.../workflow/[workflow_id]/s/[workflow_step_id]/+page.svelte`).

## Cause

PR #660 (commit `782d9e8a`, "HeyForm: scale-to-fit builder + centered form embed + skeleton loading") changed the embed iframe from a tall, effectively content-sized height to a **capped viewport height**:

```diff
- class="h-full min-h-[900px] w-full border-none"
+ class="h-[60dvh] max-h-[700px] min-h-[440px] w-full border-none"
```

HeyForm's renderer lays out its bottom controls relative to *its own* viewport (the iframe height). When a question is taller than the capped iframe (`60dvh`, max `700px`), the content and the bottom button bar collide, so they overlap and the participant must scroll within the iframe. Before #660 the iframe was tall enough (`min-h-[900px]`) that this rarely happened.

## Repro

1. Open a Conversation with a HeyForm survey step that has a longer question (several options / long text), e.g. the "Take a survey" step in the screenshots.
2. View it on the participant/public side (or preview).
3. Note the Next / Skip buttons and nav toolbar overlapping the options; scrolling down inside the embed separates them.

## Suggested fix

Give the iframe enough height for HeyForm's own layout again, without reintroducing the huge empty area on short forms that #660 was trying to fix. Options to weigh:
- Raise the cap / `min-h` so the button bar has room, or
- Let the iframe auto-size to content via HeyForm's `autoResizeHeight` postMessage (the URL already passes `autoResizeHeight=true`) and listen for the height message instead of hard-capping.

Keep the centered-card framing and skeleton loading from #660.

dtodorowa:
The participant/public survey embed now sizes its iframe to the form's actual content height instead of a fixed guess. This fixes the bottom buttons (Next / Skip + nav) overlapping the answers on longer questions, and removes the big empty gap that short questions had. Fixes #758

**How it works**

The form is a cross-origin iframe, so comhairle can't measure it directly. Our HeyForm fork now measures its active question and posts the height out; this listens for that message and sizes the iframe to it:

- listens for `{ source: 'HEYFORM', eventName: 'FORM_RESIZE', height }`
- clamps to 440–2000px and applies it to the iframe, with a height transition so per-question changes are smooth
- before the message arrives, falls back to a compact fixed height (`min-h-110` / 440px) that matches the skeleton

Also adjusted the loading skeleton ([HeyFormEmbedSkeleton.svelte](ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbedSkeleton.svelte)) to be top-aligned and compact, so it matches how the form actually renders now (content anchored at the top, sized to fit) instead of a tall vertically-centered block.

**Testing**

Verified locally against a dockerised local heyform: short questions shrink to fit, long/grouped questions grow with no footer overlap, content top-aligns.

Before: 

https://github.com/user-attachments/assets/1db51d31-6ef0-4e86-a0cb-d9f5d681c666

After:

https://github.com/user-attachments/assets/899ee07a-287a-4c65-b661-0c9eb4c0ff24

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
+++ b/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
@@ -66,9 +66,8 @@
 <!-- The form renderer is a white, self-themed UI inside a cross-origin iframe: we can't restyle its
 	internals or match comhairle's light/dark per viewer. So we frame it as a centered white card
 	(bg-white is deliberate, matching the form's own paper) rather than a full-bleed slab, so it
-	reads as an intentional embedded form on any background, dark mode included. Bounding the height
-	keeps a short form from showing a large empty area; longer forms scroll inside the card. The
-	max-width and height are tunable. -->
+	reads as an intentional embedded form on any background, dark mode included. The max-width and
+	height are tunable. -->
 <!-- A single-cell grid rather than absolute positioning: both children claim the same cell, so the
 	skeleton inherits the iframe's exact box without restating its height clamp. -->
 <div class="grid w-full grid-cols-1 grid-rows-1">
@@ -83,7 +82,7 @@
 			title="survey"
 			onload={handleLoad}
 			allow="microphone; camera"
-			class="h-[60dvh] max-h-[700px] min-h-[440px] w-full border-none transition-opacity duration-300 {ready
+			class="h-[70dvh] max-h-225 min-h-140 w-full border-none transition-opacity duration-300 {ready
 				? 'opacity-100'
 				: 'opacity-0'}"
 		></iframe>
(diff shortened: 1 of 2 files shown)


alpinegeorge (approved):
LGTM

PR #765, opened 2026-07-28
Handle Polis insights loading states (empty / not-ready / error)

dtodorowa:
A brand-new Polis step (no votes, or only seed statements) rendered a blank insights page. The `report_data` endpoint doesn't return empty in that state, it *errors* (`"Failed to get comments Failed to parse PCA data..."`) because Polis can't run the PCA clustering until there's real vote data. The loader awaited that call directly, so the throw bubbled up, the whole `polis` payload came back `undefined`, and nothing rendered.

**What changed**
- Pull the real message off the Zodios/axios error (`response.data.err`) and split two failure modes:
  - **Pre-report** (matches `parse PCA data` / `Failed to get comments`) → expected "no votes yet" state, not a real error.
  - **Everything else** (network, our-API 500s, auth) → surfaced as a genuine error.
- `PolisInsights` now renders three states: the report, a **"No insights yet"** empty state (waiting on votes), or a **"Couldn't load insights"** error state. The underlying error is always logged for debugging.

**Notes**
- The pre-report match is tied to the current Polis error wording; if it changes, a new poll falls through to the error state (safe, retryable) and a comment flags it.
- Raw error text isn't shown to admins (logged only), since `"error decoding response body"` isn't useful to them.

**Testing**
- New Polis step, no votes → "No insights yet" (previously blank).
- To see the error state without a real outage, temporarily return `reportError: 'test'` from the loader's success path.

### DIFF
--- a/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
+++ b/ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
@@ -1,6 +1,29 @@
 import type { ApiClient } from '@crownshy/api-client/api';
 import { tryCatchAsync } from '$lib/utils/errorHandling';
 
+/**
+ * Pull a human-readable message out of whatever the api client throws. Zodios wraps
+ * axios, so a failed call carries the API's JSON body at `response.data.err`; fall back
+ * to the error's own `message`, then to `String(err)`.
+ */
+function errorText(err: unknown): string {
+	if (typeof err === 'string') return err;
+	if (err && typeof err === 'object') {
+		const e = err as { response?: { data?: { err?: unknown } }; message?: unknown };
+		if (typeof e.response?.data?.err === 'string') return e.response.data.err;
+		if (typeof e.message === 'string') return e.message;
+	}
+	return String(err);
+}
+
+// A brand-new poll has no report yet: until Polis has enough vote data to run PCA, the
+// report_data endpoint errors instead of returning empty. This matches that expected
+// pre-report state so we can show a "waiting on votes" empty state for it, and a real
+// "something broke" state for everything else (network, our-API 500s, auth). Tied to the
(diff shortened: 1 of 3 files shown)


ui/packages/comhairle/src/lib/reports/polis/insights-loader.ts
```diff
@@ -1,15 +1,53 @@
...
+	const reportResult = await tryCatchAsync(() =>
+		apiClient.PolisGetReportData({ queries: { workflow_step_id: workflowStepId } })
+	);
+
+	if (reportResult.err === null) {
+		return { polis: { statementAux, reportData: reportResult.ok, reportError: null } };
+	}
+
+	const message = errorText(reportResult.err);
+	if (PRE_REPORT_SIGNATURE.test(message)) {
```
rosswilson2306, line 43: @dtodorowa I mentioned this in the PR channel, but it might be better to use http status codes on the error instead of the message but we'd need to do a bit of backend work to implement `IntoResponse` for the `PolisError` enum and then make sure that is carried through in the `ComhairleError`'s implementation of `IntoResponse`. Then you could do something like
```
if (reportResult.err.response.status === 404) {
  ...
```
or something similar. 

That would probably be a better way to handle the service errors anyway as at the moment they are probably all just coming through as 500 errors.

We have this ancient ticket to sort out the error enums https://github.com/orgs/crownshy/projects/3/views/1?filterQuery=error&pane=issue&itemId=154488295&issue=crownshy%7Ccomhairle_old%7C364 but there isn't any detail in it and some of it is already done. I'll make a new ticket to handle the IntoResponse for the service errors.

rosswilson2306 (dismissed):
I've left a comment about some backend work that might provide a slightly stabler way of handling this. But I've approved anyway.

dtodorowa:
closing for now but will reopen later with the the better solution

PR #766, opened 2026-07-28
Reporting: Survey Tool

### LINKED ISSUE #662: Reporting: Survey Tool
We want to be able to show the results of the heyform surveys. We should 

1. Create an endpoint to get the results of the heyfrom survey from the tool and reformat them into a useful format for the front end 
2. Set up some simple charts for each of the responses types 
3. Create a way to show all responses on the page as a table.

Setup a summary view for participants to see where the fit into responses

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/662

### Testing
Checked that we could render all the different charts

https://github.com/user-attachments/assets/8c48a5ae-c847-4dda-a31d-2d7f30397785

#### No Response
Screenshot from 2026-08-12 15-01-10

#### Phone Number
Screenshot from 2026-08-12 15-14-24

#### Email
Screenshot from 2026-08-12 15-14-30

#### Full name
Screenshot from 2026-08-12 15-14-34

#### Address
Screenshot from 2026-08-12 15-14-38

#### Country
Screenshot from 2026-08-12 15-14-43

#### Website
Screenshot from 2026-08-12 15-14-46

#### Multiple choice (single)
Screenshot from 2026-08-12 15-14-51

#### Multiple choice (multiple)
Screenshot from 2026-08-12 15-14-56

#### Picture choice (single)
Screenshot from 2026-08-12 15-15-01

#### Picture choice (multiple)
Screenshot from 2026-08-12 15-15-04

#### Ranking
Screenshot from 2026-08-12 15-15-14

#### Yes/No
Screenshot from 2026-08-12 15-15-17

#### Short text
Screenshot from 2026-08-12 15-15-24

#### Long text
Screenshot from 2026-08-12 15-15-28

#### Number
Screenshot from 2026-08-12 15-21-01

#### File
Screenshot from 2026-08-12 15-21-36

#### Date
Screenshot from 2026-08-12 15-21-42

#### Date Range
Screenshot from 2026-08-12 15-21-45

#### Rating
Screenshot from 2026-08-12 15-21-51

#### Opinion scale
Screenshot from 2026-08-12 15-36-39

#### Matrix
Screenshot from 2026-08-12 15-21-58

#### Legal terms
Screenshot from 2026-08-12 15-22-04


ui/packages/api-client/src/client.ts
ADSmith-0, line 1: Copied over from Axios, used to help with typing the errors that come back from the API when not using the `apiClient`

ui/packages/comhairle/src/lib/components/Charts/KdePlot.svelte
ADSmith-0, line 1: Just for reference, these changes were previously approved with this PR https://github.com/crownshy/comhairle/pull/788

ui/packages/comhairle/src/lib/components/ui/chart/chart-container.svelte
```diff
@@ -31,46 +31,46 @@
 	data-chart={chartId}
 	data-slot="chart"
 	class={cn(
-		"flex aspect-video justify-center overflow-visible text-xs",
```
ADSmith-0, line 34: `aspect-video` was messing up the height of the graphs

ui/packages/comhairle/src/lib/components/Switcher.svelte
ADSmith-0, line 1: Context: 

Image

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationInsights.svelte
```diff
@@ -10,7 +10,7 @@
 	type Props = {
 		insights: PrioritizationInsightsResponse | null;
 		step: WorkflowStepDto;
-		error?: string;
+		error: string | null;
```
ADSmith-0, line 13: Was getting a TS error here, so changed it, I hope that's alright

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -0,0 +1,78 @@
...
+		<div class="py-10">
+			<h2 class="text-md font-bold">{section.title}</h2>
+			<!-- <div class="flex flex-row"> -->
+			<!-- 	<h3 class="text-muted-foreground mr-10 text-sm"> -->
+			<!-- 		{section.answers.length} -->
+			<!-- 		{section.answers.length === 1 ? 'response' : 'responses'} -->
+			<!-- 		· -->
+			<!-- 		{Math.round(section.answers.length / section.total) * 100}% Completion -->
+			<!-- 	</h3> -->
+			<!-- </div> -->
```
ADSmith-0: Currently broken, need to figure out a better way of doing it, currently the numbers aren't correct at all

ui/packages/comhairle/src/lib/utils/errorHandling.ts
```diff
@@ -3,7 +3,7 @@ type Ok<K extends string, T> = Record<K, T> & { err: null };
 type Err<K extends string, E extends ErrorType> = Record<K, null> & { err: E };
 export type Result<K extends string, T, E extends ErrorType> = Ok<K, T> | Err<K, E>;
 
-export async function tryCatchAsync<E extends string, T>(
+export async function tryCatchAsync<T, E extends ErrorType>(
```
ADSmith-0, line 6: Turns out you can't ignore a generic if you've already provided one (which is why I put error first originally) so changed it around because it makes more sense for it to match the `Result` type

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts
```diff
@@ -3,33 +3,41 @@ import type { PageLoad } from './$types';
...
-			return await prioritizationInsightsLoader(api, step_id);
+	const response = await tryCatchAsync(async () => {
+		switch (toolConfig?.type) {
+			case 'polis':
+				return await polisInsightsLoader(step_id);
+			case 'prioritization':
+				return await prioritizationInsightsLoader(step_id);
+			case 'thinkingspace':
+				return await thinkingSpaceInsightsLoader(step_id);
+			case 'heyform':
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: I got rid of the `api` bit here because I was a bit confused, it seems like we have 2 api instances? One that gets passed through the `+page.ts` files, and then `apiClient`? I thought it would make more sense to only use one, and it seems to all still be working ok

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/+layout.ts
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/+layout.ts:1]: Hope it's alright, I deleted this because the data is coming through from a higher `+layout.ts` file anyway, so I think this just blocks loading, in the docs they say to avoid waterfalls like this if possible: https://svelte.dev/tutorial/kit/await-parent, I checked and all the data is still coming through

Image

ADSmith-0:
Need to spin out another ticket to handle group (questions that are nested together) as the data structure for that is really weird and it seemed like it was going to be really hard to do it

Also need to add in the responses bit, it looks like we might need some special data from the server to work it out

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: Oh really? This broke for me as the `api` that is passed around in the `+page.ts` / `+layout.ts` files expects to be run server side whereas the `apiClient` that is imported from `@crown-shy/api-client` expects to be run client side. 

This caused issues for me that were quite confusing for a while. When I switched to passing the `api` from the `+page.ts` it fixed. 

Could you create a prioritization tool, create some proposal responses and check out whether this breaks the prioritization step insights pages?

Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: Maybe the `api` can be both actually https://github.com/crownshy/comhairle/blob/7af741b18f23e092aa88672a622a4ce4c2492382/ui/packages/comhairle/src/routes/+layout.ts?plain=1#L9, but it was definitely breaking for me using the `apiClient` as it always I think expects to be client side and so does forward the auth cookie. See:
https://github.com/crownshy/comhairle/blob/7af741b18f23e092aa88672a622a4ce4c2492382/ui/packages/api-client/src/client.ts?plain=1#L34-L38
https://github.com/crownshy/comhairle/blob/7af741b18f23e092aa88672a622a4ce4c2492382/ui/packages/api-client/src/client.ts?plain=1#L16-L27

Comment (ADSmith-0) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: Yeah that's a really good point actually, `+page.ts`/`+layout.ts` can run client-side or server-side, so idk what happens when they try to run server side, maybe that would cause issues

It seems to be working for me, but idk, it might be worth changing it back just in case?

image

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationInsights.svelte
rosswilson2306: Ah, the eternal struggle with typescript, null vs undefined.

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: That's strange that it's working for you. Yeah, it might be worth changing it back just to be safe. I expect the insights endpoints will need to have some sort of check on user access, eg https://github.com/crownshy/comhairle/blob/0a0a24179e3f102b9dce01008d31e4093ea7929f/api/src/tools/prioritization.rs?plain=1#L720. It might change but I would think we'd have some sort of authentication on those endpoints, which means we need the auth-token cookie. So it might be safer to ensure it's the `api` from the root `+layout.ts` which can run on both server and client.

Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/design/step/[step_id]/insights/+page.ts:26]: I should have really documented why I passed the api in as an argument, sorry.

ui/packages/comhairle/src/lib/tools/heyform/guards.ts
```diff
@@ -0,0 +1,27 @@
...
+	HEYFORM_FIELD_KIND,
+	HEYFORM_NON_CHOICE_FIELD_KIND,
+	HEYFORM_OTHER_FIELD_KIND,
+	type HeyFormChoiceFieldKind,
+	type HeyFormFieldKind,
+	type HeyFormNonChoiceFieldKind,
+	type HeyFormOtherFieldKind
+} from './utils';
+
+export const isHeyFormFieldKind = (str: string | HeyFormFieldKind): str is HeyFormFieldKind =>
```
alpinegeorge, line 12: This appears to be unused

ui/packages/comhairle/src/lib/tools/heyform/guards.ts
```diff
@@ -0,0 +1,27 @@
...
+	str: string | HeyFormChoiceFieldKind
+): str is HeyFormChoiceFieldKind =>
+	HEYFORM_CHOICE_FIELD_KIND.includes(str as HeyFormChoiceFieldKind);
+
+export const isHeyFormNonChoiceFieldKind = (
+	str: string | HeyFormNonChoiceFieldKind
+): str is HeyFormNonChoiceFieldKind =>
+	HEYFORM_NON_CHOICE_FIELD_KIND.includes(str as HeyFormNonChoiceFieldKind);
+
+export const isHeyFormOtherFieldKind = (
```
alpinegeorge, line 25: This also appears to be unused

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
alpinegeorge: How so?

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -0,0 +1,78 @@
...
+		isHeyFormOtherFieldKind
+	} from '$lib/tools/heyform/guards';
+
+	interface Props {
+		data: SurveyQuestion[];
+	}
+
+	let { data }: Props = $props();
+
+	const isValidQuestion = (section: SurveyQuestion): boolean =>
```
alpinegeorge, line 19: All of these consts are defined but appear unused

rosswilson2306 (commented):
All looks good. The only thing is the api client in the insights loaders. I think it might be worth reverting that just to be on the safe side. Maybe we can revisit it but it seems like this stuff is going to be demoed soon so probably better being safe.

ADSmith-0:
@rosswilson2306 Yeah I agree! I've changed the API stuff back now

ui/packages/comhairle/src/lib/tools/heyform/guards.ts
ADSmith-0: Yeah I added this in just for the sake of completeness in case we need it in the future, I can get rid of it if we want though!

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
ADSmith-0: They're all used for the rendering logic below: https://github.com/crownshy/comhairle/blob/d4bf077b0e532b353e4e77b8f30550da6c44d5b3/ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte#L59

ui/packages/comhairle/src/lib/tools/heyform/guards.ts
ADSmith-0: It's used in the `insights-loader` here: https://github.com/crownshy/comhairle/blob/789baae159efcea62df8e5cc5f6ca0181cd23f53/ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts#L45

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
ADSmith-0: It's very weird, but it seems like the `total` is just the total number of responses, and that includes skipped responses

So skipped responses aren't a lack of answer which is what I was hoping they'd be, they're instead a falsy value, so like if it was a `short_text` field it would be `{ value: "" }`, so total is kind of useful for working out the percentage, but we have to rely on manually calculating the number of responses from the answers which is fine

The difficulty comes in with things like choice questions, because they don't follow the 1 answer = 1 submission model, they instead do `{ choice: "a", count: 5 }`, so that doesn't give us the correct info for how many responses there have been as it could be a field that allows the user to select multiple choices

Thinking about it now I think the way to do it would probably be to add a field on the choice question type that returns the number of non-empty submissions? Something like that would probably work, but then I'm not sure what a skipped choice value looks like I'll need to check

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -32,47 +32,51 @@
...
-				{#if section.answers.length <= 3}
-					<Doughnut data={section.answers} key="label" value="count" />
-				{:else}
-					<SurveyBarChart
-						data={section.answers}
-						x="label"
-						y="count"
-						kind={section.kind}
-					/>
+{#if data.length <= 0}
```
ADSmith-0, line 35: Added this just now:

Image

rosswilson2306 (approved):

alpinegeorge (approved):

PR #769, opened 2026-07-28
668/prioritization insights

rosswilson2306:
Actions:

- ui for scatter plots, kde plots and other UI used for prioritization insights
- storybook documentation

Motiviation:

- prioritization tool insights


ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
ADSmith-0, line 1: For this did we want to use the `ScatterChart` from LayerChart? https://www.layerchart.com/docs/components/ScatterChart

ui/packages/comhairle/src/lib/reports/polis/ConsensusContinuum.svelte
ADSmith-0, line 1: Also I was going to ask for this stuff, did we want to colocate the `reports` folder inside of the insights page, or do we think this is going to be used for other reports stuff elsewhere?

rosswilson2306: @ADSmith-0 I think we're going to have other kinds of reports, like the ones shown to the public, so I think keeping it in lib makes sense.

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
rosswilson2306: @ADSmith-0 I'll have another look at the layerchart ones. I had a quick look through and couldn't see any examples of the sort of negative + positive axes that we need for this case (see pictures below for the design) but I didn't have that hard a look to be honest.

screenshot_2026-07-28_15-59-05

ui/packages/comhairle/src/lib/reports/polis/ConsensusContinuum.svelte
ADSmith-0: Oh ok yeah that makes sense!

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
ADSmith-0: Oh yeah I see what you mean it doesn't look like they allow for having the axes in the center, that's really weird

rosswilson2306: @ADSmith-0 I've managed to get something working with the layerchart ScatterChart component.

ADSmith-0 (dismissed):

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationScatterPlot.svelte
```diff
@@ -56,19 +63,35 @@
...
+		<ScatterPlot
+			points={scatterPoints}
+			xAxisLabel={xQuestion.text}
+			yAxisLabel={yQuestion.text}
+			{xDomain}
+			{yDomain}
+		/>
+	{:else}
+		<div class="flex flex-col items-center justify-center gap-2">
+			<span>Somethign went wrong gathering data.</span>
```
dtodorowa, line 93: lil typo tehee

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
```diff
@@ -17,68 +18,108 @@
...
-		return [-maxAbs, maxAbs] as const;
-	});
+	let flipAxes = $state(false);
+	function handleFlipAxes(e: Event) {
+		const target = e.target as HTMLSelectElement;
 
-	function toLeftPercent(x: number, [min, max]: readonly [number, number]) {
-		return ((x - min) / (max - min)) * 100;
+		if (target.name === 'yAxis') {
+			flipAxes = target.value === xAxisLabel ? true : false;
```
dtodorowa: isn't the comparison already a boolean?
`flipAxes = target.value === xAxisLabel`

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
```diff
@@ -17,68 +18,108 @@
...
-						<div
-							class="border-primary absolute h-4 w-4 -translate-x-1/2 -translate-y-1/2 rounded-full border-2 bg-transparent"
-							style="left: {toLeftPercent(point.x, xDomain)}%; top: {toTopPercent(
-								point.y,
-								yDomain
-							)}%"
-						></div>
-					{/each}
-				</div>
+				<select
```
dtodorowa, line 60: not a biggie, but could use shadcn select

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationProposalResults.svelte
```diff
@@ -0,0 +1,280 @@
...
+					}
+					return r.question_id === questionId;
+				})
+			)
+			.flat();
+
+		return data;
+	}
+
+	function countResponses(
```
dtodorowa: doesn't this do the same section/question filtering as `extractQuestionResponses`?

I wonder if we can do something like `extractQuestionResponses(...).length` to avoid duplication?

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
rosswilson2306: @dtodorowa The shadcn select passes `value: string` as a handler. I wanted to use the same handler across different selects by using `e.target.name` so needed the full event coming through in the change handler.

ui/packages/comhairle/src/lib/reports/ScatterPlot.svelte
rosswilson2306: Yeah, right enough. I'll change that.

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationScatterPlot.svelte
rosswilson2306: Good spot!

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationProposalResults.svelte
```diff
@@ -0,0 +1,280 @@
...
+		<KdePlot
+			minLabel={question.type.minLabel}
+			maxLabel={question.type.maxLabel}
+			category={question.text}
+			rawData={{
+				[question.text]: extractQuestionResponses(proposal, question.id, section?.id).map(
+					(entry) => entry.value
+				)
+			} as Record<string, number[]>}
+			maxX={question.type.maxValue}
```
dtodorowa, line 131: not super sure about this one, but since we don't pass min, would any nonzero-min question render against the wrong axis?

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
```diff
@@ -0,0 +1,141 @@
...
+						]}
+					>
+						{#snippet children({ gradient })}
+							<Area
+								data={series.values}
+								y0={() => 0}
+								y1={(d) => zScale(d.value)}
+								curve={curveBasis}
+								fill={gradient}
+								class=""
```
dtodorowa: can drop this prob (empty class)

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
```diff
@@ -0,0 +1,141 @@
...
+		minLabel,
+		maxLabel,
+		category,
+		rawData,
+		maxX = 10,
+		options = { densityLine: true, outline: false }
+	}: Props = $props();
+
+	let overlap = $state(4);
+	let height = $state(200);
```
dtodorowa: do we need state here? i think they are never reassigned

dtodorowa: and height too?

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
```diff
@@ -0,0 +1,141 @@
...
+
+	// Kernel density estimator
+	function kde(kernel: (v: number) => number, thresholds: number[], data: number[]) {
+		return thresholds.map((t) => ({
+			x: t,
+			value: mean(data, (d) => kernel(t - d)) ?? 0
+		}));
+	}
+
+	const N = categories.length;
```
dtodorowa: these consts (`N`, `thresholds`, `seriesData`, `maxDensity`) are computed from the `rawData` prop but aren't `$derived`, so they won't recompute if the prop changes. Do we need derived or do we not expect changes? prob not? just checking

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationProposalResults.svelte
rosswilson2306: Yeah, you're right. I'll update.

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
```diff
@@ -0,0 +1,141 @@
...
+	const basePadding = { top: 0, bottom: 0, left: 0, right: 0 };
+	const thresholds = range(0, maxX + 1, 1);
```
dtodorowa: can continuous questions have a nonzero minValue?

if yes, this (and `averagePercentage`) assume that the scale starts at 0. The curve will prob get clipped or shifted?

Not super sure about this, but wanna double check

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationProposalResults.svelte
rosswilson2306: I think this graph will really only work if based off the xAxis. I needed an upper threshold for the xAxis but the yAxis is completely dependent on the data. So I don't think there's any danger of anything rendering against the wrong axis.

I never thought about the min, however. I'll add an optional prop for the min value.

screenshot_2026-07-29_09-44-59

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
rosswilson2306: No, we don't. That's left behind from the example I used. I'll remove.

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
rosswilson2306: Probably not, but I'll update to use derived anyway.

ui/packages/comhairle/src/lib/reports/KdePlot.svelte
rosswilson2306: I don't think at the moment it can go below 0 but I guess it could. I've made some updates with the new `minX` prop to account for this (I think).

dtodorowa (approved):

PR #770, opened 2026-07-29
Add actions concept to permissions model

alpinegeorge:
Motivations:

 - Modifying the permissions model to check for allowed actions instead of assigned roles allowa for much easier extraction-based verification when multiple overlapping roles can perform the same actions.
 - This introduction can be performed without impacting existing databases.
 - In future, we could extend this to allow custom roles, should the need arise.

Actions:

 - Added the concept of actions.
 - Added actions enums for existing system and conversation resources, and associated said actions with the existing roles.
 - Restructured permissions model code to improve maintainability.
 - Update permissions caching strategy to cache all user/org roles.

### DIFF
--- a/api/src/models/permissions.rs
+++ b/api/src/models/permissions.rs
@@ -1,16 +1,47 @@
+//! # Permissions model
+//!
+//! This module defines the permissions model for the Comhairle application, including resource types, roles, actions, and permission handling.
+//!
+//! ## Definitions:
+//! - Resource Type      : A string that identifies a category of resources in the system (e.g., "system", "conversation").
+//! - Resource ID        : A UUID that uniquely identifies a specific resource within its category.
+//! - Role               : An alias for a group of permissions that can be assigned to a user or organization on a specific resource.
+//! - Action             : A specific operation that can be performed on a resource, such as reading, updating, or deleting it.
+//! - Policy             : Defines which actions are allowed for a specific resource type on a per-role basis.
+//! - Permission Triplet : A combination of resource type, resource ID, and role name that uniquely identifies a permission assignment.
+//!
+//! ## Adding New Resource Types, Roles, and Actions
+//! Resource types, roles, and actions are each a single enum ([`ResourceType`],
+//! [`Role`], [`Action`]). To extend the model:
+//! 1. Add a variant to the relevant enum, preserving any persisted string value via
+//!    `#[strum(serialize = "...")]` / `#[serde(rename = "...")]`.
+//! 2. Map new roles to their resource type in [`Role::resource_type`] and their allowed
+//!    actions in [`Role::actions`]; map new actions to their resource type in
+//!    [`Action::resource_type`].
+//! 3. If the resource is addressed by a path, add an extractor struct via
+//!    `define_owned_resource!` / `define_unowned_resource!` so [`crate::routes::auth::authorize`]
+//!    can resolve its id and owner.
+//!
+//! ## Helper Macros
+//! - `define_owned_resource!`   : Defines a resource struct with an owner and implements the `ExtractResourceId` and `OwnedResource` traits for it.
+//! - `define_unowned_resource!` : Defines a resource struct without an owner and implements the `ExtractResourceId` and `OwnedResource` traits for it.
+
 use std::sync::Arc;
 
 use crate::models::users::UserIden;
 use crate::redis_connection::RedisConnection;
+use aide::OperationIo;
 use axum::extract::{FromRequestParts, Path};
 use chrono::{DateTime, Utc};
 use schemars::JsonSchema;
+use sea_query::JoinType;
 use sea_query::{Expr, OnConflict, PostgresQueryBuilder, Query, enum_def};
-use sea_query::{IdenStatic, JoinType};
 use sea_query_binder::SqlxBinder;
 use serde::{Deserialize, Serialize};
 use sqlx::prelude::FromRow;
-use sqlx::{PgPool, Row, query_as_with};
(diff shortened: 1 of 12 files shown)


api/src/models/permissions.rs
```diff
@@ -21,123 +32,217 @@ use crate::models::{
...
-pub struct SystemAdminRole;
+// ---------- //
+// * MACROS * //
+// ---------- //
 
-impl NamedRole for SystemAdminRole {
-    fn name() -> &'static str {
-        "admin"
-    }
+macro_rules! define_unoowned_resource {
```
rosswilson2306: Is `unoowned` a typo for `unowned`?

api/src/models/permissions.rs
```diff
@@ -494,84 +735,84 @@ pub async fn has_resource_permission(
...
+/// Check whether a user, or their organization, can perform an action on a resource.
+pub async fn can_perform_resource_action<Action: PermittedAction>(
+    state: &Arc<ComhairleState>,
+    resource_id: &Uuid,
+    action: Action,
+    user_id: &Uuid,
+    organization_id: Option<&Uuid>,
+) -> Result<bool, ComhairleError> {
+    // System admins are globally authorized for all actions.
+    if has_resource_permission(
```
rosswilson2306, line 873: Maybe I've misunderstood what the `SystemAdmin` role was. I thought this was simply a replacement for the previous admin_users that allowed them to access the admin interface. Is this instead a super admin role? 

If someone who is on the `ADMIN_USERS` list in the helm_charts signs up for the first time do they receive the SystemAdmin role?

rosswilson2306: This is the bit I'm talking about https://github.com/crownshy/comhairle/blob/2997801e0b8f9896ef8577c947b75adfaf86e26c/api/src/routes/auth.rs?plain=1#L322-L335, it looks like it grants the SystemAdmin role to anyone in the `ADMIN_USERS` config. 

Because all the servers share the same `ADMIN_USERS` list, this could mean that someone from maine signs up to the scot gov server and they are given super admin access.

Unless I'm misunderstanding.

api/src/routes/conversations.rs
```diff
@@ -192,9 +189,10 @@ async fn get_conversation(
     if !original_conversation.is_live {
         if let Some(user) = &user {
             if user.id != original_conversation.owner_id
-                && !has_resource_permission(
+                && !can_perform_resource_action(
                     &state,
-                    ConversationContentEditorRole::make_triplet(&original_conversation.id),
+                    &original_conversation.id,
+                    ConversationAction::Read,
```
rosswilson2306: I take it this doesn't block non-admin users on the public site?

api/src/models/permissions.rs
alpinegeorge: Yes, good spot

api/src/models/permissions.rs
alpinegeorge: I may have misunderstood the current admin model as well.
It's meant as a super admin role just now, but we can change that.
I've assigned permissions to the SystemAdminRole here, but really there's no need to since it short circuits permissions checks to preserve the previous admin user behaviour. I'll remove the confusing permissions assignments and add a note.

alpinegeorge: I should also just use is_admin_user here instead.

api/src/routes/conversations.rs
alpinegeorge: If the user has the content_editor role just now, then it won't block. Otherwise, I don't think there's a way for non-admin users to see a non-live conversation, right? Either way, it shouldn't be a change in behaviour from the current setup.

api/src/models/permissions.rs
alpinegeorge: Reworked to separate the admin and super admin role types. Permissions can now only be edited by super admins, and system admin now matches the old admin behaviour.
Also simplified model to three enums to reduce work to extend resource types/roles/actions.

dtodorowa (dismissed):

dtodorowa:
approved if this needs to go to fix the issue with moderation permissions. (unless, you are gonna do a targeted fix for that now,?)

api/src/models/permissions.rs
```diff
@@ -21,175 +52,414 @@ use crate::models::{
...
-        PermissionTriplet(SYSTEM_RESOURCE_TYPE, &SYSTEM_RESOURCE_ID, Self::name())
+    /// The actions this role is permitted to perform (the permission policy).
+    pub fn actions(self) -> &'static [Action] {
+        match self {
+            Role::SuperAdmin => &[
+                Action::ListPermission,
+                Action::GrantPermission,
+                Action::RevokePermission,
+            ],
+            Role::Admin => &[],
```
rosswilson2306, line 350: @alpinegeorge Will this prevent conversation owners (admin but not super admin) from sharing conversations with other admin users?

Maybe that's what we want? But if so superadmins would need to be able to view all conversations so that they can grant sharing permissions on behalf of other users (you might have implemented this but I haven't gotten that far).

alpinegeorge: If said admin is the conversation owner, then they can edit permissions, but other admins won't be able to.
We could add a new role which allows that type of moderation though.

rosswilson2306: Perfect, just wanted to check.

rosswilson2306 (dismissed):

alpinegeorge:
Allowed resource owners to list users with permissions on that resource, and removed debug logging.

stuartlynn (approved):

Issue #750, opened 2026-07-27
Prioritisation Tool: We should make it clear when required fields are missing before moving on


stuartlynn:
It seems like you cant submit feedback on a proposal if the comment section is blank. We should either make that optional or we should show some kind of feedback when a participant tried to submit the feedback without having all the required fields filled out

PR #773, opened 2026-07-30
Prioritization: neutral slider default + clear answered/unanswered states

dtodorowa:
Follow-up to the required-fields work (#750). That change surfaced *when* a rating is unanswered; this one fixes the slider itself so "unanswered" reads clearly and answering the neutral value isn't a fiddle.

## Problem

The continuous (slider) question had two rough edges:
- The thumb sat at the **minimum** but the value was actually `null` until dragged, so it looked answered while counting as unanswered.
- To pick a value at the resting position, you had to drag the thumb away and back, a click that landed on it registered nothing.

## Change

**Slider now has an explicit "not answered yet" state.** The thumb rests in the **middle** as a neutral starting position, but with **no track fill** and a hollow grey thumb. That absence of fill is the "you haven't answered" cue, so:
- it doesn't anchor participants toward either end the way defaulting-to-min did, and
- the value stays genuinely unanswered until they interact, so required-field validation still fires.

**On touch:** the track fills, the thumb fills in with the theme's primary colour, and the leaning-side label (e.g. "Public") highlights.

**Middle-tap fixed:** a single tap on the resting thumb now commits the middle value instead of doing nothing, handled with a capture-phase gesture flag + pointerup so drags and clicks-elsewhere aren't affected.

**Cursor:** proper `grab`/`grabbing` on the thumb, `pointer` on the track.

**Submit button:** muted (`secondary`) while the proposal is incomplete, solid (`default`) once complete. Still always clickable, so clicking while incomplete triggers the required-field highlights.

https://github.com/user-attachments/assets/90efbbe0-939a-42bf-9524-ef7df4fde246

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
@@ -540,7 +540,11 @@
 					<Button onclick={onDone}>Finish</Button>
 				{/if}
 			{:else}
-				<Button onclick={submitAndAdvance} disabled={submitting}>
+				<Button
+					variant={isComplete ? 'default' : 'secondary'}
+					onclick={submitAndAdvance}
+					disabled={submitting}
+				>
 					{#if submitting}
 						<LoaderCircle class="mr-2 h-4 w-4 animate-spin" />
 					{/if}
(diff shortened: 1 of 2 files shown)


dtodorowa:
update: the slider looks better now (improved contrast, so it doesn't look disabled anymore)
image

alpinegeorge (approved):

PR #777, opened 2026-07-30
751 prioritisation completion feedback

### LINKED ISSUE #751: Prioritisation Tool: We should give visible feedback that a user has compleated a proposal
Currently when a participant completes a proposal and submits (not the full submission at the end but their feedback on a single proposal), we dont show any visual feedback and the form just resets. 

We should 

1. Show some kind of indication that the submission was successful 
2. Scroll back to the top of the proposal element so that it's clear they have started the next proposal.

dtodorowa:
Right now when you submit your feedback on a single proposal, nothing happens visually. The form just resets and it's not obvious the submission worked or that you've moved to the next proposal.

This adds:

- **A short "Response submitted" interstitial** a card with a checkmark fades in over the screen for about a second after each submit, so it's a clear, deliberate moment instead of a silent reset.
- **Auto-scroll to the top of the next proposal** once the interstitial clears, we advance and scroll the new proposal into view, so it's obvious you're starting a fresh one.

Built on the existing `Portal` overlay pattern (same one `LoadingOverlay` uses) and standard shadcn tokens, so no new deps.

**Notes for review**
- The interstitial fires on every submit, including the last one right before the summary view. Happy to skip it on the final proposal if that feels like one beat too many.
- Timing is a single constant (`SUBMITTED_INTERSTITIAL_MS`, 1100ms) if we want it snappier/slower.

https://github.com/user-attachments/assets/f2f25372-8e24-436f-8ae1-bd5816f89e02

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
@@ -57,7 +57,11 @@
 
 	async function scrollToProposalTop() {
 		await tick();
-		proposalTopEl?.scrollIntoView({ behavior: 'smooth', block: 'start' });
+		const reduceMotion = window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+		proposalTopEl?.scrollIntoView({
+			behavior: reduceMotion ? 'auto' : 'smooth',
+			block: 'start'
+		});
 	}
 
 	/** Brief success interstitial shown between proposals so a submit registers as a
@@ -596,12 +600,12 @@
 {#if showSubmitted}
 	<Portal>
 		<div
-			class="bg-background/70 animate-in fade-in-0 fixed inset-0 z-100 flex items-center justify-center backdrop-blur-sm duration-200"
+			class="bg-background/70 animate-in fade-in-0 fixed inset-0 z-100 flex items-center justify-center backdrop-blur-sm duration-200 motion-reduce:animate-none"
 			role="status"
 			aria-live="polite"
 		>
 			<div
-				class="bg-card border-border animate-in zoom-in-95 fade-in-0 flex flex-col items-center gap-3 rounded-2xl border px-10 py-8 text-center shadow-lg duration-200"
+				class="bg-card border-border animate-in zoom-in-95 fade-in-0 flex flex-col items-center gap-3 rounded-2xl border px-10 py-8 text-center shadow-lg duration-200 motion-reduce:animate-none"
 			>
 				<CheckCircle2 class="text-primary size-12" />
 				<span class="text-foreground text-lg font-semibold">Response submitted</span>
--- a/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+	import { tick } from 'svelte';
+	import { Portal } from 'bits-ui';
 	import { Button } from '$lib/components/ui/button';
 	import * as Card from '$lib/components/ui/card';
 	import { Progress } from '$lib/components/ui/progress';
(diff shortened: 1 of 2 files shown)


alpinegeorge (dismissed):

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -563,3 +592,20 @@
 		{/if}
 	</div>
 {/if}
+
+{#if showSubmitted}
+	<Portal>
+		<div
+			class="bg-background/70 animate-in fade-in-0 fixed inset-0 z-100 flex items-center justify-center backdrop-blur-sm duration-200"
```
alpinegeorge: What about prefers-reduced-motion? We shouldn't animate in that case, right?

ADSmith-0: I think from what I know of accessibility small subtle animations are usually ok? It's more like the big movement, stuff like parallax scrolling or scaling or moving an element by a medium-large amount you need `prefers-reduced-motion` for, but I could be wrong

dtodorowa: goo catch! Latest commit should address it  

https://github.com/user-attachments/assets/b8af600d-2c69-4dff-b683-4295fd2baeae

dtodorowa: > I think from what I know of accessibility small subtle animations are usually ok? It's more like the big movement, stuff like parallax scrolling or scaling or moving an element by a medium-large amount, but I could be wrong

you might be onto something here bc my OS would only block big animations when I have it enabled, hmm. need to learn more. And I was surprised this wasn't blocked automatically

dtodorowa: but maybe scroll animations are triggering hm, so makes sense to remove?

ADSmith-0: I'm not sure the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion) say "Animations such as scaling or panning large objects can be vestibular motion triggers." So I would think anything outside of that should usually be ok?

But now I'm second guessing a bit because for the media library I added a small translate on hover

https://github.com/user-attachments/assets/c9e5d1dc-6b7c-425e-9985-a8fc9de7b9c0

And now I'm wondering if that counts and we need to put that behind a flag, idk what "large object" counts as, but I imagine if the pan is very small then it should probably be ok?

ADSmith-0: @dtodorowa 

Fairly short read that's helpful too: https://www.a11yproject.com/posts/understanding-vestibular-disorders/, from that page:

"Note that rather than setting animation: none and transition: none, we instead set the animation-duration and transition-duration properties to a short enough duration that the motion itself is no longer noticable. This is to prevent issues in any case there is a dependency on the animation to run (such as when listening to the [animationend event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event))."

alpinegeorge (dismissed):

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
dtodorowa: tehee this was a good rabbit hole, thanks for digging into it! think we're fine here: the interstitial hides itself on a `setTimeout`, not an animationend listener, so `motion-reduce:animate-none` just drops the motion and the timer still advances things normally. I think the 0.001s thing from the article only matters when something is waiting on the animation to finish

I pushed a [note to the style guide](https://github.com/crownshy/comhairle/pull/777/changes#diff-476f8d3866159e359d47fe8844c04fa4fc9e658d9e31f771b73701474081c488R288) (new "Motion and reduced motion" section) trying to capture this: gate big motion, small stuff is usually fine, and the shorten-vs-remove rule for the waiting-on-animationend case. Lmk if it makes sense or if we should put it somewhere else.

ADSmith-0: Yeah that looks great! Cheers for writing it up! :smile:

ADSmith-0 (approved):

Issue #780, opened 2026-07-30
Ground the learning assistant in the current learn-step content


dtodorowa:
Split out from #583 (which is scoped to visual/design consistency only).

When a user is on a learn step, the learning assistant > both the inline one at the bottom of the learn page and the "Find out more" sidebar  should receive the step's **content** as context, so questions like "explain this page in simpler words" work against the actual text on screen instead of relying on knowledge-base retrieval.

### Today
- Learn page (inline `LearningAssistant`): sends only the step **title**, as a `[Reading "<title>"]` prefix.
- Sidebar assistant: sends no context at all.

### Want
- Pass the step **content** (not just the title) as grounding context.
- Applies to both the inline assistant and the sidebar when the user is on a learn step.
- Needs live step content plumbed into the layout-level sidebar.

### Scope
- Learn steps only, other step types have no "page content" and need no context.
- Do we want to do other steps too?

dtodorowa:
This came up again. It originally was discussed here #591, so going to link the two

Issue #782, opened 2026-07-30
Improve tab strip design on the 'Find out more' sidebar


dtodorowa:
## Problem

On the "Find out more" drawer/sidebar, the close arrow sits in the top-left corner *above* the tab strip, which looks awkward and disconnected from the tabs.

The close button is `absolute top-0 left-0 p-3` while the drawer content has `py-12` padding, so the arrow floats alone in the empty space above the tabs (Learning assistant / FAQs / Privacy Policy).

## What we want

Rework the header so the close control and the tab strip feel like one coherent row rather than the arrow hovering on its own line above the tabs. A few directions to consider:

- Put the close arrow inline with the tab strip (e.g. arrow on the left, tabs filling the rest of the row).
- Or move/restyle the close control so it reads as part of the drawer chrome (top-right, or aligned with the tabs baseline).
- Reduce the large top padding gap that currently separates the arrow from the tabs.

## Where

`ui/packages/comhairle/src/lib/components/ConversationSupportSidebar.svelte` the `Drawer.Close` + `Tabs.Root` block.

<img width="753" height="510" alt="Image" src="https://github.com/user-attachments/assets/247d2474-64af-451e-8dc4-4749662e4f10" />

## Notes

- Keep it accessible (the close control must stay a focusable, labelled button).
- Tabs currently use `text-sm`; worth checking against the text-base floor while we are in here.

dtodorowa:
Built a throwaway prototype: the chevron sits on the same row as the tabs, underline tabs feel lighter than the chunky dark pill, and it kills the empty row above the tabs that started this.

<img width="1464" height="883" alt="Image" src="https://github.com/user-attachments/assets/f527310d-3413-4588-8a6b-b3ba8145ad6a" />

PR #785, opened 2026-07-30
Align the learning assistant across the learn page and sidebar

### LINKED ISSUE #583: Align the design for the learning assistant in both the learning page and sidebar
We should reuse the Q&A structure that the learning pages use in the side bar for the learning assistant.

Should show the same content as the session on the learning pages

dtodorowa:
The "Find out more" sidebar used a separate chat-bubble component while the learn page used the collapsible Q&A one. They already shared the same session under the hood, so this just unifies the presentation: the sidebar now renders the same learning assistant as the learn page.

**What changed**
- Moved `LearningAssistant` (+ its skeleton) out of `tools/learn` into `lib/components/` since it's now shared.
- Added a `variant="sidebar"` mode that hides only the redundant uppercase heading (the drawer tab already says "Learning assistant") and fills/scrolls inside the drawer. Same intro copy, same Q&A, same behaviour otherwise.
- Wired it into `ConversationSupportSidebar`, dropped the now-unused `user` prop.
- Deleted the old `ChatBot`, the dead `CollapsibleChat`, and their `types.ts`.

Passing step content into the assistant came to my mind again while looking at this, but tracking it separately in #780.

We should probably do a design pass of that sidebar.. i will try to prototype some


ADSmith-0:
@dtodorowa 

Just a general question what is that component bit at the top of the screenshot?

image

Because I think we use it across a bunch of places in the app it would probably be good to have it as a component (it's in the thinking space pages, and also in the designs for the survey reports stuff)

I was thinking maybe it would be called a `Rail`/`Selector` or something like that? What do you think?

ADSmith-0 (approved):

dtodorowa:
> @dtodorowa
> 
> Just a general question what is that component bit at the top of the screenshot?
> 
> image
> Because I think we use it across a bunch of places in the app it would probably be good to have it as a component (it's in the thinking space pages, and also in the designs for the survey reports stuff)
> 
> I was thinking maybe it would be called a `Rail`/`Selector` or something like that? What do you think?

Yeeh good point. And while we are here, I am not sure I love the look of this anyway, it seems to be hardcoded into dark mode + dk how i feel about the 2-row collapse/navigation. I'm not fully satisfied and I think it would benefit from some design love, so [created a ticket](https://github.com/crownshy/comhairle/issues/782) to revisit the look. We should create another one to create the component and reuse it

ADSmith-0:
> > @dtodorowa
> > Just a general question what is that component bit at the top of the screenshot?
> > image
> > Because I think we use it across a bunch of places in the app it would probably be good to have it as a component (it's in the thinking space pages, and also in the designs for the survey reports stuff)
> > I was thinking maybe it would be called a `Rail`/`Selector` or something like that? What do you think?
> 
> Yeeh good point. And while we are here, I am not sure I love the look of this anyway, it seems to be hardcoded into dark mode + dk how i feel about the 2-row collapse/navigation. I'm not fully satisfied and I think it would benefit from some design love, so [created a ticket](https://github.com/crownshy/comhairle/issues/782) to revisit the look. We should create another one to create the component and reuse it

Sounds great! :slightly_smiling_face:

PR #800, opened 2026-08-05
fix: scope Polis vote progress per step instead of per poll

### LINKED ISSUE #799: Two Polis steps share vote progress: min-votes gating collides (localStorage keyed by poll id, not step)
## Summary

When a conversation has two Polis steps, the minimum-votes (required votes) gating gets confused: vote progress and the "you can continue" flag from one step bleed into the other.

## What I first thought it was

My first read was that this is purely a storage-keying bug. The `required_votes` **threshold value** is fine (stored per-step, server-side, in each `workflow_step`'s `tool_config` JSONB). But the vote **progress** (the running count toward the threshold) lives in `localStorage`, and it was keyed by the Polis poll id, not the workflow step id:

```ts
// polisVoteStore.ts (before)
function storageKey(userId: string, polisId: string): string {
	return `${STORAGE_PREFIX}-${userId}-${polisId}`;
}
```

So if two Polis steps resolved to the same poll id, they shared one localStorage entry: meeting the threshold in step 1 wrote `hasMetThreshold: true`, and step 2 loaded the same key on mount and treated you as already done.

## But that framing is off

Two Polis steps ending up on the **same poll id isn't a normal state**. Each step mints its own poll at setup (`polis_setup` -> `create_poll`), and preview -> live promotion (`clone_tool`/`launch`) mints another fresh poll. `PolisToolSetup` has no `poll_id` field, so there's no admin path to point a step at an existing poll. You only get a shared poll id if someone deliberately wires it that way (hand-editing config/DB). So the "two steps, same poll" repro isn't a state the product actually produces on its own.

## The deeper cause

The Polis embed was the **only** tool on the step page not wrapped in `{#key workflowStep.id}`:

```svelte
{#if toolConfig?.type === Polis.TOOL_NAME}
    <Polis.UserUI ... />   <!-- no {#key workflowStep.id} -->
{/if}
```

SvelteKit reuses the page component on param-only navigation, so hopping from one Polis step straight to the next reused the same `<PolisEmbed>` instance. The state that gates "continue" is captured once at mount:

- `const initialData = getVoteData(...)` runs once
- `let hasMetThreshold = $state(initialData.hasMetThreshold)` seeded once
- `const polis = new PolisApi(...)` constructed once, keeps talking to step 1's poll

So two **adjacent** Polis steps shared vote count, the continue flag, and even the live Polis conversation, **regardless of whether their poll ids matched**. Re-keying the store alone would not have fixed this case, because those captured values never re-read. The poll-id storage key was the second-order symptom; the missing remount was the primary cause.

## The fix

Both lines of defense, since they compose:

1. **`{#key workflowStep.id}` around the Polis block** so it remounts per step (fixes the adjacent case and rebuilds `PolisApi` against the right poll).
2. **Vote store keyed per step + preview/live** (`voteScopeKey = ${preview|live}-${stepId}`) instead of by poll id, so progress stays separate across reloads, same-poll steps, and preview vs live.

Covered by unit tests in `polisVoteStore.test.ts`.

## Still worth doing (follow-up)

This is still per-device localStorage. The real long-term fix is a **server-side source of truth**: an endpoint returning how many votes/comments a user has submitted for a step, and gating on that. It survives a cleared cache, works across devices, and removes the shared-storage foot-gun entirely. Tracking separately.

## Repro

1. Create a conversation with two Polis steps.
2. Vote past the required-votes threshold on the first step.
3. Open the second step: it showed the threshold met / continue unlocked without voting.

dtodorowa:
## What

Fixes two Polis steps bleeding vote progress into each other: hitting the required-votes threshold on one step unlocked "continue" on another without voting.

## The actual cause

First read was a storage-keying bug: vote progress in `localStorage` was keyed by the Polis poll id, not the step. True, but the deeper issue is that the Polis embed was the **only** tool on the step page not wrapped in `{#key workflowStep.id}`. SvelteKit reuses the page component on param-only nav, so two adjacent Polis steps shared one `<PolisEmbed>` instance (vote count, threshold flag, and even the live `PolisApi` conversation) regardless of poll id. Re-keying the store alone wouldn't have fixed that, since those values are captured once at mount.

Also worth noting: two steps landing on the same poll id isn't a normal state anyway. Each step mints its own poll at setup, so this only happens in edge configs.

## Changes

- Wrap the Polis embed in `{#key workflowStep.id}` so it remounts per step (fixes the adjacent case, rebuilds `PolisApi` against the right poll).
- Key the vote store per step + preview/live (`voteScopeKey`) instead of per poll id.
- Add unit tests for the store scoping.

## Follow-up

This is still per-device `localStorage`. The real long-term fix is a server-side source of truth (an endpoint returning the participant's vote count for a step, gated on that instead of a local counter). I'll open a separate issue/PR for that endpoint.

After fix:

https://github.com/user-attachments/assets/27ac7614-f3a4-4a09-88d0-6e9111c0046c

Before Fix:

https://github.com/user-attachments/assets/99754b3f-340f-4e04-9e7d-b4f01df9d38e

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/polis/PolisEmbed.svelte
+++ b/ui/packages/comhairle/src/lib/tools/polis/PolisEmbed.svelte
@@ -26,6 +26,7 @@
 		onDone: () => void | Promise<void>;
 		requiredVotes?: number;
 		workflowStepId?: string;
+		isPreview?: boolean;
 		showRemainingStatementCount?: boolean;
 		onCanContinueChange?: (canContinue: boolean) => void;
 	};
@@ -37,11 +38,16 @@
 		onDone,
 		requiredVotes = 10,
 		workflowStepId = polis_id,
+		isPreview = false,
 		onCanContinueChange,
 		showRemainingStatementCount
 	}: Props = $props();
 
 	const stepId = workflowStepId;
+
+	// Vote progress is stored per step (and separately for preview vs live) so two
+	// Polis steps, even ones sharing a poll, never share their threshold state.
+	const voteScopeKey = `${isPreview ? 'preview' : 'live'}-${stepId}`;
 
 	let polisCurrentStatement = $state<PolisStatement | undefined>(undefined);
 	let polisLoading = $state(false);
@@ -74,7 +80,7 @@
 
 	type Screen = 'voting' | 'add-opinion' | 'continue-prompt' | 'completed';
 
-	const initialData = getVoteData(user_id, polis_id);
+	const initialData = getVoteData(user_id, voteScopeKey);
 	let totalVotes = $state(initialData.totalVotes);
 	let hasMetThreshold = $state(initialData.hasMetThreshold);
 	let screen = $state<Screen>('voting');
@@ -166,7 +172,7 @@
 			anchoredRemaining--;
 		}
 
-		const data = incrementVotes(user_id, polis_id, safeRequiredVotes);
+		const data = incrementVotes(user_id, voteScopeKey, safeRequiredVotes);
 		hasMetThreshold = data.hasMetThreshold;
 
 		if (data.totalVotes === safeRequiredVotes) {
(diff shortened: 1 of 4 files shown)


alpinegeorge (approved):

dtodorowa:
i need to add a follow up to make the polis voting transitions less jarring, jesus

PR #804, opened 2026-08-05
Organizations admin interface

alpinegeorge:
Motivations:

 - We need to track organization contact emails to support a new use
   case in CivicOS.
 - We would like an easy way to be able to manage organizations via the
   admin interface.

Actions:

 - Add contact email as an optional field to the organizations data
   table and create request model.
 - Added a functional interface for now which we can improve if we want
   to make this self-serve later.

### DIFF
--- a/api/src/models/organization.rs
+++ b/api/src/models/organization.rs
@@ -33,6 +33,7 @@ pub struct Organization {
     #[partially(omit)]
     pub mission: TextContentId,
     pub org_type: OrganizationType,
+    pub contact_email: Option<String>,
     pub external_url: Option<String>,
     pub regions: Vec<Uuid>,
     #[partially(omit)]
@@ -74,12 +75,13 @@ impl std::fmt::Display for OrganizationType {
     }
 }
 
-const DEFAULT_COLUMNS: [OrganizationIden; 9] = [
+const DEFAULT_COLUMNS: [OrganizationIden; 10] = [
     OrganizationIden::Id,
     OrganizationIden::Name,
     OrganizationIden::Description,
     OrganizationIden::Mission,
     OrganizationIden::OrgType,
+    OrganizationIden::ContactEmail,
     OrganizationIden::ExternalUrl,
     OrganizationIden::Regions,
     OrganizationIden::CreatedAt,
@@ -92,6 +94,7 @@ pub struct CreateOrganization {
     pub description: String,
     pub mission: String,
     pub org_type: OrganizationType,
+    pub contact_email: Option<String>,
     pub external_url: Option<String>,
     pub regions: Option<Vec<Uuid>>,
 }
@@ -100,6 +103,9 @@ impl CreateOrganization {
     fn columns(&self) -> Vec<OrganizationIden> {
         let mut columns = vec![OrganizationIden::Name, OrganizationIden::OrgType];
 
+        if self.contact_email.is_some() {
+            columns.push(OrganizationIden::ContactEmail);
+        }
         if self.external_url.is_some() {
             columns.push(OrganizationIden::ExternalUrl);
         }
@@ -113,6 +119,9 @@ impl CreateOrganization {
     fn values(&self) -> Vec<sea_query::SimpleExpr> {
         let mut values = vec![(*self.name).into(), self.org_type.clone().into()];
 
+        if let Some(value) = &self.contact_email {
+            values.push(value.clone().into());
+        }
         if let Some(value) = &self.external_url {
             values.push(value.clone().into());
         }
@@ -164,6 +173,9 @@ impl PartialOrganization {
         }
         if let Some(value) = &self.org_type {
             values.push((OrganizationIden::OrgType, value.clone().into()));
+        }
+        if let Some(value) = &self.contact_email {
+            values.push((OrganizationIden::ContactEmail, value.clone().into()));
         }
         if let Some(value) = &self.external_url {
             values.push((OrganizationIden::ExternalUrl, value.clone().into()));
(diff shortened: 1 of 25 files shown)


api/src/models/organization.rs
```diff
@@ -156,6 +193,138 @@ pub async fn create(
...
+    for user_email in user_emails {
+        let trimmed = user_email.trim();
+        if trimmed.is_empty() {
+            continue;
+        }
+
+        match users::get_user_by_email(trimmed, db).await {
+            Ok(user) => {
+                let update_request = users::UpdateUserRequest {
+                    organization_id: Some(*organization_id),
```
rosswilson2306, line 211: This might be a rare case but if someone is already attached to an organisation and someone else tries to add them to another organization, will this override the first organization and mean they are no longer a part of it?

api/src/routes/organizations.rs
```diff
@@ -66,40 +306,367 @@ async fn get(
...
+async fn add_member(
+    State(state): State<Arc<ComhairleState>>,
+    Path(organization_id): Path<Uuid>,
+    RequiredUser(user): RequiredUser,
+    resource: OrganizationResource,
+    Json(payload): Json<UpsertOrganizationUserBody>,
+) -> Result<(StatusCode, Json<UpsertOrganizationUserResponseDto>), ComhairleError> {
+    authorize(&state, &user, Action::OrganizationUpdate, &resource).await?;
+
+    let allow_create_user = payload.allow_create_user.unwrap_or(false);
```
rosswilson2306, line 362: Probably not something to worry about here but is this something that should be a permission action for the user?

api/src/routes/organizations.rs
```diff
@@ -66,40 +306,367 @@ async fn get(
...
+        existing
+    } else {
+        organization::update(&state.db, &organization_id, &partial).await?
+    }
+    .into();
 
     Ok((StatusCode::OK, Json(organization)))
 }
 
+async fn update_localized_text_content(
```
rosswilson2306, line 624: This looks like it might be a useful method elsewhere so maybe should go in the `crate::models::translations` as `upsert_localized_text_content`. What do you think?

api/src/models/users.rs
```diff
@@ -455,6 +493,42 @@ pub async fn get_user_by_username(username: &str, db: &PgPool) -> Result<User, C
...
+        .build_sqlx(PostgresQueryBuilder);
+
+    sqlx::query_as_with::<_, User, _>(&sql, values)
+        .fetch_all(db)
+        .await
+        .map_err(ComhairleError::DatabaseError)
+}
+
+/// Set or clear the organization membership for a user.
+pub async fn set_user_organization_id(
```
stuartlynn, line 514: We might want users to be able to join multiple organizations in the future but I think this is fine just now.

ui/packages/comhairle/src/routes/(admin)/admin/organizations/[organization_id]/+page.ts
```diff
@@ -0,0 +1,63 @@
...
+
+		if (access?.canManageTeam) {
+			const teamResponse = await fetch(`/api/organizations/${params.organization_id}/team`);
+			if (teamResponse.ok) {
+				team = await teamResponse.json();
+			}
+		}
+
+		let regions: { id: string; name: string }[] = [];
+		const regionsResponse = await fetch('/api/regions?limit=500');
```
Comment (rosswilson2306) [inline ui/packages/comhairle/src/routes/(admin)/admin/organizations/[organization_id]/+page.ts]: Do the api client methods not work for these requests?

ui/packages/comhairle/src/routes/(admin)/admin/organizations/new/+page.svelte
```diff
@@ -0,0 +1,191 @@
...
+	let orgType: OrganizationType = $state('non_profit');
+	let contactEmail = $state('');
+	let externalUrl = $state('');
+	let allRegions = $state<{ id: string; name: string }[]>([]);
+	let selectedRegions = $state<string[]>([]);
+	let userEmails = $state('');
+	let organizationAdminEmails = $state('');
+
+	onMount(async () => {
+		const response = await fetch('/api/regions?limit=500');
```
rosswilson2306: Same here, could the api client method be used?

stuartlynn (dismissed):

stuartlynn:
This looks good to me. Just a quick comment that in future we may want to allow users to be part of multiple organizations but going to merge this in so we can make progress

ui/packages/comhairle/src/routes/(admin)/admin/organizations/new/+page.svelte
```diff
@@ -0,0 +1,191 @@
...
+	let name = $state('');
+	let description = $state('');
+	let mission = $state('');
+	let orgType: OrganizationType = $state('non_profit');
+	let contactEmail = $state('');
+	let externalUrl = $state('');
+	let allRegions = $state<{ id: string; name: string }[]>([]);
+	let selectedRegions = $state<string[]>([]);
+	let userEmails = $state('');
+	let organizationAdminEmails = $state('');
```
rosswilson2306: If this is a form to create a new organization we should probably use superform with a schema if possible to keep it consistent with other forms.

stuartlynn (dismissed):

api/src/models/organization.rs
alpinegeorge: Rare for now, but we will have to face it soon. I think switching to a many-many users to organizations model would be the best solution here.

api/src/routes/organizations.rs
alpinegeorge: Sounds good

rosswilson2306 (approved):

Issue #815, opened 2026-08-06
Glossary: auto-tooltip defined terms in Learn step rich text


dtodorowa:
## What

Give conversation builders a **glossary**: a shared list of defined terms, each with a short explanation. Any occurrence of a glossary term in a Learn step's rich text gets a hover tooltip automatically, so builders define a term once instead of marking up every occurrence by hand.

Example entry (shape from the discussion):

```json
{ "text": ["bus", "autobus"], "tooltip": "A vehicle that carries multiple people" }
```

Any occurrence of `bus` or `autobus` in the content renders as a term with a tooltip showing the explanation. The `text` array holds the term plus its synonyms / translated forms.

## Why

Participants hit unfamiliar or jargon terms in Learn content (e.g. "referral"). Rather than tooltip each occurrence manually in the editor, we want a single glossary as the source of truth that auto-populates tooltips across the content. HeyForm is already handled separately, so this ticket is **Learn steps only** (the tiptap rich-text path).

## Approach

1. **Glossary data model + storage** — a list of `{ text: string[], tooltip: string }` entries. Recommend scoping it at the **conversation level** (shared across all steps), since the same terms recur across Learn pages. Open for discussion (see below).
2. **Builder UI** — a small management surface for conversation builders to add / edit / remove glossary entries (term + synonyms + explanation).
3. **Auto-tooltip on render** — when Learn rich text is rendered to participants, match glossary terms in the text and wrap each match so it shows the explanation on hover.

## Where this plugs in (from codebase research)

Learn content is authored **in Comhairle** via the tiptap editor and rendered through a shared SSR-safe path, so this is very doable on the frontend:

- **Editor / schema**: `ui/packages/comhairle/src/lib/components/RichTextEditor/editorConfig.ts` — `getBaseExtensions({ mode })` is the single schema shared by editor and renderer.
- **Custom-node pattern to copy**: `.../RichTextEditor/extensions/sourceDocument.ts` — a custom node that renders differently in `editor` vs `renderer` mode. Good template if we go the tiptap-node route.
- **Render path (display side)**: `.../lib/utils/renderRichText.ts` (`renderRichTextToHtml`) via `@tiptap/static-renderer`, consumed by `.../RichTextEditor/ContentRenderer/ContentRenderer.svelte`. Auto-matching most likely lives here so it applies to already-authored content without re-editing.
[...]

PR #821, opened 2026-08-07
Glossary 2/4: read from metadata + show in Learn steps

dtodorowa:
### Stack (merge bottom to top)

1. #820 — rich-text rendering engine
2. **#821 — read from metadata + show in Learn steps ← you are here**
3. #822 — CSV import + AI translation helpers
4. #823 — admin Glossary tab

---

**Stack 2 of 4** for the Learn-step glossary tooltips (#815). Targets stack 1 (`815-glossary-1-render-engine`); review/merge that one first.

### What this adds
Wires the engine from stack 1 to real conversation data for participants.

- `lib/glossary/parseGlossary.ts` — coerces the untyped conversation `metadata` blob into a valid `Glossary`, degrading malformed data to "no tooltips" instead of a render crash. Exports `GLOSSARY_METADATA_KEY`.
- `lib/glossary/localizedGlossary.ts` — holds terms and explanations per locale and resolves down to the participant's current locale, falling back to the conversation's primary locale
- `LearnUI.svelte` — reads the glossary off `conversation.metadata` and passes it to `ContentRenderer`

### Effect
After this, a glossary stored on a conversation shows tooltips in that conversation's Learn steps, in the participant's language. There's no editor yet, so the glossary would have to be written into `metadata` by hand until stack 4.

### Tests
Parse + localized-resolve unit tests pass (branch runs 56 total).

Shout if you want more detail on the locale fallback logic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ui/packages/comhairle/src/lib/glossary/parseGlossary.test.ts
+++ b/ui/packages/comhairle/src/lib/glossary/parseGlossary.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect } from 'vitest';
+import { parseGlossary, glossaryFromMetadata } from './parseGlossary';
+
+describe('parseGlossary', () => {
+	it('keeps well-formed entries', () => {
+		expect(parseGlossary([{ text: ['bus', 'autobus'], tooltip: 'A vehicle' }])).toEqual([
+			{ text: ['bus', 'autobus'], tooltip: 'A vehicle' }
+		]);
+	});
+
+	it('trims terms and tooltip and drops blank terms', () => {
+		expect(parseGlossary([{ text: [' bus ', '', '  '], tooltip: '  A vehicle  ' }])).toEqual([
+			{ text: ['bus'], tooltip: 'A vehicle' }
+		]);
+	});
+
+	it('drops entries with no usable term or no tooltip', () => {
+		expect(
+			parseGlossary([
+				{ text: [], tooltip: 'orphan' },
+				{ text: ['bus'], tooltip: '' },
+				{ text: ['bus'], tooltip: 'ok' }
+			])
+		).toEqual([{ text: ['bus'], tooltip: 'ok' }]);
+	});
+
+	it('ignores non-string terms and malformed rows', () => {
+		expect(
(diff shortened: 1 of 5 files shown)


alpinegeorge (approved):

ui/packages/comhairle/src/lib/glossary/parseGlossary.test.ts
```diff
@@ -0,0 +1,50 @@
...
+});
+
+describe('glossaryFromMetadata', () => {
+	it('reads the glossary key out of a metadata object', () => {
+		const metadata = { glossary: [{ text: ['bus'], tooltip: 'A vehicle' }], other: 1 };
+		expect(glossaryFromMetadata(metadata)).toEqual([{ text: ['bus'], tooltip: 'A vehicle' }]);
+	});
+
+	it.each([null, undefined, 'x', 5, {}, { glossary: null }])(
+		'returns [] when there is no valid glossary (%p)',
```
ADSmith-0, line 45: Just curious what does the `%p` mean here? I know in `printf` languages you can do that kind of stuff, but I didn't know it was a thing in JS?

dtodorowa: yehh exacto, not a JS thing, it's from `it.each` (vitest/jest). the title string gets `printf`-style formatting, so `%p` = pretty-print each value into the test name. that way each row shows up as its own test like `...({ glossary: null })` instead of them all having the same name.

others exist too (`%s` string, `%i` number, `%#` index) but `%p` is the nice default since it handles objects + primitives

more examples here: https://vitest.dev/api/test#test-each

ADSmith-0: Oh awesome! I had no idea! Thank you for explaining! :smile:

dtodorowa: np np thanks for asking ^^

PR #822, opened 2026-08-07
Glossary 3/4: CSV import + AI translation helpers

dtodorowa:
### Stack (merge bottom to top)

1. #820 — rich-text rendering engine
2. #821 — read from metadata + show in Learn steps
3. **#822 — CSV import + AI translation helpers ← you are here**
4. #823 — admin Glossary tab

---

**Stack 3 of 4** for the Learn-step glossary tooltips (#815). Targets stack 2 (`815-glossary-2-learn-integration`).

### What this adds
Two pure, unit-tested helpers that the admin editor (stack 4) builds on. No UI or wiring here.

- `lib/glossary/glossaryCsv.ts` — a small RFC 4180 CSV parser plus two-column glossary import (term + synonyms in the first column, explanation in the second)
- `lib/glossary/translateGlossary.ts` — fills a target locale's missing terms and explanations by translating the primary-locale text; best-effort and non-destructive (never overwrites existing translations)

### Tests
CSV + translation unit tests pass (branch runs 73 total).

These have no consumer until stack 4 — I split them out to keep that PR focused on the Svelte UI. Fine to fold them together if you'd rather review three PRs than four; let me know.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
+++ b/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
@@ -12,8 +12,7 @@ export function parseCsvRows(text: string): string[][] {
 	let field = '';
 	let inQuotes = false;
 
-	// Skip a UTF-8 BOM if a spreadsheet exported one.
-	for (let i = text.charCodeAt(0) === 0xfeff ? 1 : 0; i < text.length; i++) {
+	for (let i = 0; i < text.length; i++) {
 		const char = text[i];
 
 		if (inQuotes) {
--- a/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
+++ b/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
@@ -13,9 +13,7 @@ export function parseCsvRows(text: string): string[][] {
 	let inQuotes = false;
 
 	// Skip a UTF-8 BOM if a spreadsheet exported one.
-	let i = text.charCodeAt(0) === 0xfeff ? 1 : 0;
-
-	for (; i < text.length; i++) {
+	for (let i = text.charCodeAt(0) === 0xfeff ? 1 : 0; i < text.length; i++) {
 		const char = text[i];
 
 		if (inQuotes) {
--- a/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
+++ b/ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
@@ -1,0 +1,88 @@
+import type { Glossary } from './types';
+import { parseGlossary } from './parseGlossary';
+
+/**
+ * Parses CSV text into rows of string cells (RFC 4180: quoted fields may contain commas,
+ * newlines, and "" escaped quotes). Hand-rolled because there's no CSV dep in the repo and
+ * the glossary's two-column shape doesn't warrant adding one.
+ */
+export function parseCsvRows(text: string): string[][] {
+	const rows: string[][] = [];
+	let row: string[] = [];
+	let field = '';
+	let inQuotes = false;
+
+	// Skip a UTF-8 BOM if a spreadsheet exported one.
+	let i = text.charCodeAt(0) === 0xfeff ? 1 : 0;
+
+	for (; i < text.length; i++) {
+		const char = text[i];
+
+		if (inQuotes) {
+			if (char === '"') {
+				if (text[i + 1] === '"') {
+					field += '"';
+					i++;
+				} else {
+					inQuotes = false;
+				}
+			} else {
+				field += char;
+			}
+			continue;
+		}
+
+		if (char === '"') {
+			inQuotes = true;
+		} else if (char === ',') {
+			row.push(field);
+			field = '';
+		} else if (char === '\n') {
+			row.push(field);
+			rows.push(row);
+			row = [];
+			field = '';
+		} else if (char !== '\r') {
+			// A lone or CRLF '\r' is dropped; '\n' handles the row break.
+			field += char;
+		}
+	}
+
+	row.push(field);
+	rows.push(row);
+
+	// Drop blank lines (a trailing newline leaves an empty final row).
+	return rows.filter((cells) => !(cells.length === 1 && cells[0].trim() === ''));
+}
+
+const HEADER_TERM_NAMES = ['term', 'terms', 'word', 'words', 'phrase'];
+const HEADER_DEFINITION_NAMES = ['tooltip', 'definition', 'explanation', 'meaning', 'description'];
+
+function looksLikeHeader(cells: string[]): boolean {
+	const first = cells[0]?.trim().toLowerCase() ?? '';
+	const second = cells[1]?.trim().toLowerCase() ?? '';
+	return HEADER_TERM_NAMES.includes(first) || HEADER_DEFINITION_NAMES.includes(second);
+}
+
+/**
+ * Parses a CSV glossary. Two columns: the first holds the term plus any synonyms separated
+ * by `;` or `|` (comma is the CSV field separator, so it can't double as the synonym one);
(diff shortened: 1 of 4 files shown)


ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
```diff
@@ -0,0 +1,88 @@
...
+export function parseCsvRows(text: string): string[][] {
+	const rows: string[][] = [];
+	let row: string[] = [];
+	let field = '';
+	let inQuotes = false;
+
+	// Skip a UTF-8 BOM if a spreadsheet exported one.
+	let i = text.charCodeAt(0) === 0xfeff ? 1 : 0;
+
+	for (; i < text.length; i++) {
```
ADSmith-0: I think you're do this in a 1 line
```ts
	for (let i = text.charCodeAt(0) === 0xfeff ? 1 : 0; i < text.length; i++) {
```

But not a big deal either way!

ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
```diff
@@ -0,0 +1,88 @@
...
+ * the glossary's two-column shape doesn't warrant adding one.
+ */
+export function parseCsvRows(text: string): string[][] {
+	const rows: string[][] = [];
+	let row: string[] = [];
+	let field = '';
+	let inQuotes = false;
+
+	// Skip a UTF-8 BOM if a spreadsheet exported one.
+	let i = text.charCodeAt(0) === 0xfeff ? 1 : 0;
```
alpinegeorge: Interesting that it chose to include this. Byte order shouldn't make a difference for utf-8.
And it's generally not recommended to include it.

alpinegeorge (dismissed):

ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	row.push(field);
+	rows.push(row);
+
+	// Drop blank lines (a trailing newline leaves an empty final row).
+	return rows.filter((cells) => !(cells.length === 1 && cells[0].trim() === ''));
+}
+
+const HEADER_TERM_NAMES = ['term', 'terms', 'word', 'words', 'phrase'];
+const HEADER_DEFINITION_NAMES = ['tooltip', 'definition', 'explanation', 'meaning', 'description'];
```
ADSmith-0, line 56: Just wondering if it's possible that some definitions might include these terms?

Like "Term" can be used in political contexts as well I think, would that mean that it gets stripped out?

ui/packages/comhairle/src/lib/glossary/translateGlossary.ts
```diff
@@ -0,0 +1,60 @@
...
+		target: string,
+		primary: string
+	) => Promise<string> = aiTranslateContent
+): Promise<LocalizedGlossary> {
+	if (targetLocale === primaryLocale) return glossary;
+
+	const result: LocalizedGlossary = [];
+	for (const entry of glossary) {
+		const text = { ...entry.text };
+		const tooltip = { ...entry.tooltip };
```
ADSmith-0, line 28: I think the more efficient way to clone an object is now 
```ts
structuredClone(obj);
```

But again it's not a big deal 🙂

ui/packages/comhairle/src/lib/glossary/translateGlossary.ts
```diff
@@ -0,0 +1,60 @@
...
+
+		const sourceTerms = entry.text[primaryLocale];
+		if (sourceTerms?.length && !entry.text[targetLocale]?.length) {
+			try {
+				const translated = await translate(
+					sourceTerms.join(', '),
+					targetLocale,
+					primaryLocale
+				);
```
ADSmith-0: Can use `tryCatchAsync` for these `try...catch`s if we want, but not a huge deal either way

ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
dtodorowa: shouldn't happen but good to double check. it only looks at the first row, and it's an exact match on the whole cell, not a substring. so "a political term for..." won't trip it since the cell isn't literally just "term". only weird case is if someone's very first row is a term that IS the word "term", which is pretty unlikely. 

can make it stricter (need both columns to match) if you'd rather? i would leave that as a follow up ticket either way tho 🤭

ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
dtodorowa: nice, done

ui/packages/comhairle/src/lib/glossary/translateGlossary.ts
dtodorowa: yeh, gp! swapped both over

ui/packages/comhairle/src/lib/glossary/translateGlossary.ts
dtodorowa: maybe i'm wrong but i think shallow spread's fine here actually

we only ever add a new locale key, never touch the existing arrays, so nothing shared gets mutated 🤔
`structuredClone` would deep copy everything for no reason? idk? left it as is but lmk if you disagree, maybe there is a gap in my understanding too

ui/packages/comhairle/src/lib/glossary/glossaryCsv.ts
dtodorowa: ok i looked into this, you're right.

the reason was apparently that a BOM **could** break header detection, but JS trim() treats the BOM as whitespace, and everything downstream (header check, term + tooltip extraction) already trims, so a stray BOM will get neutralized on its own.

dropped itt and left a test that imports a BOM'd "CSV UTF-8" export end to end to verify. good catch

alpinegeorge (approved):

ui/packages/comhairle/src/lib/glossary/translateGlossary.ts
ADSmith-0: Yeah in terms of the data shallow merging is totally fine here, it was more from a performance/memory perspective, because the spread operator (`...`) can be a bit bad performance-wise [link 1](https://jonlinnell.co.uk/articles/spread-operator-performance) [link 2](https://4markdown.com/be-careful-with-the-spread-operator-in-javascript/)

Basically what it's doing under the hood is it's just running a `for` loop, so when you spread what it's doing is:
```ts
const copiedObj = {}
for (const key of obj) {
  copiedObj[key] = obj[key];
}
return copiedObj
```
Same with array spreading like `const newArr = [ ...arr ];` - it's ok for small objects, but the bigger the objects is the worse the performance will get

If you use the inbuilt methods instead like
```ts
const newArr = arr.slice();
```
then I think it translates to native code under-the-hood for the copying so it ends up usually being a faster and more memory efficient

Although to be clear It's probably not noticeable unless you have a really big object or do it lots across the codebase but it's just one of the things where it's a free performance win :slightly_smiling_face: 

_Although_ looking into it more I'm actually wrong about `structuredClone` it seems as though it doesn't actually translate and run natively and is the slowest at cloning an object :sob: It seems as though the performance issues are more to do with arrays, so it's totally correct to do spread for objects, ignore me haha [link](https://benchmarklab.azurewebsites.net/Benchmarks/Show/19868/0/structuredclone-vs-spread-vs-objectassign#latest_results_block), but with `slice()` it is more efficient [link](https://measurethat.net/Benchmarks/Show/2667/0/arrayprototypeslice-vs-spread-operator#latest_results_block)

But yeah I'll leave the comment anyway in case it's of interest :slightly_smiling_face:

Issue #591, opened 2026-07-08
Learning Assistant: give access to Learn Step page content


dtodorowa:
Learning Assistant and Learn Step editor currently share the same Knowledge Base (uploaded docs), but the Assistant has no visibility into the actual content authored in the Learn Step editor itself.

**Problem**:
Participants are likely to assume the Learning Assistant knows at least as much as they do about the material they're reading, since a human learner reads the learn page content, it's inconsistent for the assistant not to "know" it too. This also blocks features like "explain this in simpler words," which require the assistant to know what "this" refers to.

**Proposed solution:**
Add Learn Step page content as an input source to the Learning Assistant's knowledge base, in addition to the existing uploaded Knowledge Base docs.

shuyanglin:
Related to this, we also use Knowledge base to provide context for thinking space (I assume) 

On reflection, I think sometimes we see knowledge base as a place for organisers to add context to the topic. While in practice they could only think of adding one or two supporting documents in to the database, instead of thinking knowledge base as the place that actually should provide enough knowledge.

PR #827, opened 2026-08-07
780 ground the learning assistant in the current learn step content

### LINKED ISSUE #780: Ground the learning assistant in the current learn-step content

dtodorowa:
### What & why

The participant-facing Learning Assistant (both the inline one on a learn page and the "Find out more" support sidebar) previously only answered from admin-uploaded knowledge-base documents. It had no idea what was actually written on the learn steps, so asking it about the page in front of you didn't work well.

This grounds the assistant in the conversation's own learn-step content by syncing that content into the knowledge base as a single reserved-name document, retrieved through the existing RAG `{knowledge}` path. We went with the KB-sync approach (not per-request page injection) so it reuses the infra #591 already wanted. Full rationale and the rejected options are in ADR-0010.

### Changes

**Backend** 
- New `POST /conversation/{id}/documents/sync_learning_content` endpoint (admin-only).
- `build_learn_content_document` concatenates every learn step's authored text (published `tool_config` only) into one markdown doc with per-step headings and `---` separators.
- `resolve_page_content` handles both modern `text_content` pages and legacy inline-markdown pages, in the conversation's primary locale.
- Sync = delete the old reserved-name doc (`learn-step-content.md`) then re-upload and re-parse via the existing `DocumentJob` worker, which also reconnects the chat bot once parsing finishes (per ADR-0007).
- No-op (no job created) when there's nothing to sync.

**Prompt**
- System prompt now tells the model to ask a short clarifying question on vague requests ("explain this") instead of guessing.

**Frontend**
- Admin learning-assistant settings page gets a **"Sync learn content"** button with copy stating changes aren't picked up automatically (manual by design).
- The support sidebar now passes the current learn step's title as a retrieval hint, matching what the inline assistant already sent. Non-learn steps pass no hint. 

### Known limits (by design, see ADR-0010)
- Edits to a learn step are invisible to the assistant until someone clicks Sync.
- "Explain *this* page" retrieves from parsed chunks, not the whole page, so it can be imperfect and may trigger the clarifying question. This is the accepted good-to-have; grounded Q&A is the must.
- First sync carries parse latency before the assistant can answer from the content.

*another one I would like to point your attention to:*
Stale citations after re-sync. Re-syncing rebuilds the learn-content document under a new id and deletes the old one. Citations in assistant answers from before the sync point at the deleted document, so opening those sources fails. For now this is handled with an error message, not prevention: the PDF viewer detects the missing-source case and shows a plain-language note ("the learning materials have been updated since this answer was written... ask again for up-to-date sources") instead of a raw HTTP error. Asking the question again returns fresh citations against the new document. A fuller fix (stable document ids across syncs, or invalidating/annotating prior answers) is out of scope here and can follow if it bites in real usage.

### Manual test coverage

**Admin sync**
- [x] Sync button appears on the conversation's Learning Assistant settings page
- [ ] Clicking Sync shows the "sync started... ready once parsing finishes" notification
- [x] Button shows spinner + "Syncing..." and is disabled while in flight
- [x] After parsing completes, a `Learning Material.pdf` document is present in the KB
- [x] Re-syncing replaces the old doc (only one, not duplicates)
- [x] Sync on a conversation with no learn steps succeeds and creates no job (no-op)

**Assistant answering (inline, on a learn page)**
- [x] Ask a question answerable from the current step → grounded answer
- [ ] Ask "explain this page in simpler words" → reasonable answer for that step
- [x] Ask something not in any learn content or KB → "not found in the dataset" behaviour
- [x] Vague prompt with no context ("explain this") → assistant asks a clarifying question instead of guessing

**Assistant answering (support sidebar / "Find out more")**
- [x] Sidebar assistant answers against learn content the same way the inline one does
- [x] On a learn step, the current step title is passed as a hint (retrieval biases toward that step)
- [x] On a non-learn step, no hint is passed and the assistant behaves as before

**Staleness / lifecycle**
- [x] Edit a learn step without syncing → assistant still answers from the old content (expected)
- [x] Sync again → assistant picks up the edited content after parse completes

image

image

image


ui/packages/comhairle/src/lib/components/PdfViewer/PdfViewer.svelte
```diff
@@ -284,6 +290,20 @@
 			})
 			.catch((e) => {
 				if (cancelled) return;
+				// A missing document comes back as 404 (pdf.js MissingPDFException); the
+				// RAGFlow-proxied "document not found" path can also surface as 500
```
ADSmith-0, line 295: Could potentially do this as `tryCatchAsync`, but the logic looks pretty difficult so happy to keep this as is 🙂

ui/packages/comhairle/src/lib/components/PdfViewer/PdfViewer.svelte
```diff
@@ -284,6 +290,20 @@
...
+				// typically because the learn content was re-synced after this answer.
+				const status =
+					typeof (e as { status?: unknown })?.status === 'number'
+						? (e as { status: number }).status
+						: undefined;
+				const name = e instanceof Error ? e.name : '';
+				sourceUnavailable =
+					name === 'MissingPDFException' ||
+					name === 'UnexpectedResponseException' ||
+					status === 404 ||
```
ADSmith-0: Btw I added a `HttpStatus` constant like we have for the backend in case you want to use it https://github.com/crownshy/comhairle/blob/staging/ui/packages/comhairle/src/lib/utils/constants.ts, again I'm happy either way though 🙂

ui/packages/comhairle/src/lib/utils/previewKind.ts
ADSmith-0, line 1: Just wondering if we need any additional context for the file name? If I read `lib/utils/previewKind` I'm not sure I would know what it's in reference to, but that could just be me lacking context knowledge, if this is a common thing we use in different situations?

api/src/routes/documents.rs
```diff
@@ -615,6 +856,91 @@ mod tests {
...
+                                    ..Default::default()
+                                },
+                            ))
+                        })
+                    });
+            })
+            .await?;
+
+        let boundary = "test-boundary";
+        let body = format!(
```
ADSmith-0: I think you can use the `MultipartBodyBuilder` from the `test_helpers` here if you want

```rust
let body = MultipartBodyBuilder("test-boundary")
  .add_file("Learning material.pdf", "application/pdf", "%PDF-1.7 fake pdf bytes")
  .build();
```

But up to you 🙂

ui/packages/comhairle/src/lib/components/PdfViewer/TextViewer.svelte
```diff
@@ -0,0 +1,53 @@
...
+	let failed = $state(false);
+	// A 404 means the source is gone (e.g. the learning material was re-synced and
+	// this answer's document was replaced), not a transient load failure - worth a
+	// clearer, less alarming message than a generic error.
+	let gone = $state(false);
+
+	// Rendered as escaped text (not {@html}) so an uploaded text file can never
+	// inject markup. Synced learn content is markdown, which reads fine this way.
+	onMount(async () => {
+		try {
```
ADSmith-0: Could maybe do `tryFetch` here if you want to?

ui/packages/comhairle/src/lib/utils/previewKind.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Map a filename to the viewer that should render it, or `null` when we have no
+ * in-page viewer for that type (the caller should fall back to a plain download).
+ * Callers that always want a viewer (e.g. RAGFlow source chunks, which are only
+ * ever PDFs or uploaded docs) can default an unknown type with `?? 'pdf'`.
+ */
+export function getPreviewKind(fileName: string): PreviewKind | null {
+	const lower = fileName.toLowerCase();
+	if (lower.endsWith('.pdf')) return 'pdf';
+	if (IMAGE_EXTENSIONS.some((ext) => lower.endsWith(ext))) return 'image';
```
ADSmith-0, line 23: Could maybe do this as a switch statement? It depends if you're wanting to export the extensions to do checks elsewhere, but I don't mind either way tbh!

Could also use `Media.getExtension(filename)` here I think, where it will return the extension, you can then just do a `.includes()` like:
```ts
const extension = Media.getExtension(fileName.toLowerCase());
if (IMAGE_EXTENSIONS.includes(extension)) return "image";
// or
switch (extension) {
  case ".jpg":
  case ".jpeg":
  ...
      return "image";
  ...
}
```

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte
```diff
@@ -65,6 +83,116 @@
...
+	// Fetch an image URL and rasterise it to a PNG data URL via a canvas. Going through a canvas
+	// (rather than embedding the bytes directly) normalises any source format - including webp /
+	// avif that pdfmake can't embed - to PNG, and lets us downscale huge images. Returns null on
+	// any failure (network, or a CORS-tainted canvas): a missing image is skipped, never fatal.
+	function imageUrlToPngDataUrl(url: string): Promise<string | null> {
+		return new Promise((resolve) => {
+			const img = new Image();
+			img.crossOrigin = 'anonymous';
+			img.onload = () => {
+				try {
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte:101]: Can do `tryCatchAsync` here if we want to, but idm either way

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/LearnSyncStatus.svelte
```diff
@@ -0,0 +1,76 @@
...
+	);
+	const isDone = $derived(!!doc && (doc.parse_status === 'DONE' || doc.parse_progress >= 1));
+	const isFailed = $derived(doc?.parse_status === 'FAIL');
+	const percent = $derived(doc ? Math.round(doc.parse_progress * 100) : 0);
+
+	async function poll() {
+		if (!doc) {
+			return;
+		}
+		try {
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/LearnSyncStatus.svelte]: Can do `tryCatchAsync` here if we want but no worries either way!

api/src/routes/documents.rs
```diff
@@ -165,16 +178,28 @@ async fn download_document(
 
     let knowledge_base_id = get_knowledge_base_id(&state, &conversation_id).await?;
     let download_stream = bot_service
-        .download_document(document_id, knowledge_base_id)
+        .download_document(document_id.clone(), knowledge_base_id)
```
rosswilson2306: @dtodorowa Why did this need to change to a clone? I can't see anywhere other than in error messages where it is used and those should reference automatically.

api/src/routes/documents.rs
```diff
@@ -263,6 +288,205 @@ async fn get_knowledge_base_id(
...
+    job_id: Option<Uuid>,
+    document: Option<ComhairleDocument>,
+}
+
+/// Replace the conversation's learn-content knowledge-base document with a
+/// freshly-generated PDF and kick off a re-parse.
+///
+/// The PDF is built client-side from the learn steps (a text-bearing document
+/// with real tables) and posted here as multipart `file`; the backend owns the
+/// RAGFlow dance. The reserved-name document is deleted first (RAGFlow re-chunks
```
rosswilson2306: @dtodorowa What is a `RAGFLOW dance`? Sounds funky.

Can I be a pain and ask you to replace references to RAGFlow with "bot service"? We're trying to keep the services provider agnostic so if we change providers comments like this would still make sense. You can leave it in the annoying bit about Ragflow errors cause that is a specific issue with Ragflow error handling.

api/src/routes/documents.rs
```diff
@@ -22,12 +22,25 @@ use crate::{
...
 };
 
+/// Reserved knowledge-base document name for the auto-synced learn-step content.
+///
+/// A conversation has at most one of these in its knowledge base; a re-sync
+/// replaces it. The frontend builds a text-bearing PDF from the learn steps and
+/// uploads it under this name, so RAGFlow parses it natively (clean text +
+/// per-chunk highlight positions) and the existing PDF viewer displays it. See
+/// the "learn content as PDF" spec and ADR-0014.
+pub const LEARN_CONTENT_DOCUMENT_NAME: &str = "Learning material.pdf";
```
rosswilson2306: Maybe we should use a name a little less likely to be created by a human by accident. Something like `comhairle_learning_step_material.pdf` ? I feel like most users would use spaces over underscores when naming files so that might add a little more safety.

api/src/routes/documents.rs
```diff
@@ -263,6 +288,205 @@ async fn get_knowledge_base_id(
...
+#[instrument(err(Debug), skip(state))]
+async fn learn_content(
+    State(state): State<Arc<ComhairleState>>,
+    Path(conversation_id): Path<Uuid>,
+    user: OptionalUser,
+) -> Result<(StatusCode, Json<LearnContentResponse>), ComhairleError> {
+    require_conversation_document_access(&state, &user, &conversation_id).await?;
+
+    let conversation = conversation::get_by_id(&state.db, &conversation_id).await?;
+    let locale = &conversation.primary_locale;
```
rosswilson2306, line 455: Shouldn't this come from the `LocaleExtractor` so that it works across different languages?

api/src/routes/documents.rs
```diff
@@ -263,6 +288,205 @@ async fn get_knowledge_base_id(
...
+}
+
+/// Return the conversation's learn-step content as raw, renderable pages (grouped by
+/// step), so the assistant's source viewer can show the real learn pages via the
+/// rich-text renderer instead of the flattened knowledge-base document.
+///
+/// Reads the draft (`preview_tool_config`), matching what sync ingests, so the viewer
+/// shows the same content the assistant answers from.
+#[instrument(err(Debug), skip(state))]
+async fn learn_content(
```
rosswilson2306, line 447: This is probably fine for now, but I feel like these routes are very specific to the learn tool so should maybe go on that router. Don't worry about it for now though, I'm a little unsure as the documents router is also very tied to that tool.

ui/packages/comhairle/src/lib/utils/constants.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Reserved knowledge-base document name for the auto-synced learn-step content.
+ * Mirrors `LEARN_CONTENT_DOCUMENT_NAME` in the API (`api/src/routes/documents.rs`).
+ */
+export const LEARN_CONTENT_DOCUMENT_NAME = 'Learning material.pdf';
```
rosswilson2306: Like I mentioned in the backend, I feel like a less human likely name would be a bit safer.

rosswilson2306 (commented):
Left a few comments / questions.

ui/packages/comhairle/src/lib/components/PdfViewer/PdfViewer.svelte
dtodorowa: will keep this one for now, I think the cancel-guard + pdf.js .then().catch() doesn't fit tryCatchAsync nicely;
on that note tho: would be nice to add some tech debt to go around and asses all of our tryCatches and turn them into tryCatchAsync

ui/packages/comhairle/src/lib/components/PdfViewer/PdfViewer.svelte
dtodorowa: niiiiice, done, using HttpStatus.NotFound/InternalServerError now

ui/packages/comhairle/src/lib/utils/previewKind.ts
dtodorowa: air, added a doc comment up top explaining it picks the viewer for a file. kept the name since it ties to PreviewKind, lmk if you'd still rather rename!

api/src/routes/documents.rs
dtodorowa: good call, swapped it in, way cleaner!

ui/packages/comhairle/src/lib/components/PdfViewer/TextViewer.svelte
dtodorowa: donnne, uses tryFetch now

ui/packages/comhairle/src/lib/utils/previewKind.ts
dtodorowa: left the .endsWith version, Media drags EasyForm in and this is meant to be a tiny leaf util. happy to revisit if you feel strongly tho, we can create a follow-up

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte
Comment (dtodorowa) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/+page.svelte:101]: keeping this one also, the promise never rejects (always resolves null on fail) so there's nothing for `tryCatchAsync` to catch I think?

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/LearnSyncStatus.svelte
Comment (dtodorowa) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/learning-assistant/LearnSyncStatus.svelte]: dunnnnee 🥳

api/src/routes/documents.rs
dtodorowa: yeh this makes sense, I think `download_dovument` didn't need to own the id. switched `download_document` to take `&str` so it now bothers it. And we don't need a clone now and the id still around for the 404 message

api/src/routes/documents.rs
dtodorowa: LMAOO this is not original thought, it's my agent trying to be funny 💀 cringee

makes total sense re bot service! Will add it to the style guide also.

How about the Ragflow error-handling bits? They are provider-specific, are we ok with leaving those?

api/src/routes/documents.rs
dtodorowa: good call, renamed it to `comhairle_learning_step_material.pdf` (underscores  will defo make the collision more unlikely)

updated both the API and frontend side

ui/packages/comhairle/src/lib/utils/constants.ts
dtodorowa: noice, resolved that also on the earlier comment

api/src/routes/documents.rs
dtodorowa: agree that these are learn concerns. only reason they're here is they share the Knowledge Base plumbing (kb-id lookup, access check, parse job) with the generic doc routes

don't want to bloat this PR tho, do you think it's ok to add a follow-up and split into a learn router? 

Added this: https://github.com/crownshy/comhairle/issues/862

api/src/routes/documents.rs
dtodorowa: sooo, this only feeds the admin sync, which builds one shared PDF for the whole convo's knowledge base, not a per-viewer thing.

I think so `primary_locale` iis more stable since it's the language it was authored in. `LocaleExtractor` would tie it to whichever admin's cookie synced and it felt a bit random (but maybe I am accounting for issues we won't ever have)

real limitation tho is the single language pdf :/ I think if we want to do this properly, we need a synced per locale. 
But that felt kind of overwhelming to deal with before tomorrow, so I didn't look into it. 

just logged it as [#863](https://github.com/crownshy/comhairle/issues/863) and noted it in ADR-0010, so it doesn't get lost. We probably need to do it before the Gaelic engagement or Afghanistan

ui/packages/comhairle/src/lib/utils/previewKind.ts
ADSmith-0: No Media is just the wider interface, cause we work with files so much throughout the codebase I wanted to make a single interface to do all this kind of stuff, but yeah it's a static method so you don't need to instantiate it, you should just be able to copy-paste the line above and this is all it will do
```ts
	static getExtension(filename: string): string | undefined {
		const index = filename.lastIndexOf('.');
		if (index < 0) return undefined;
		return filename.slice(index);
	}
```

But yeah I agree it's a small thing so we can do it in another ticket if we want! :slightly_smiling_face: :

api/src/routes/documents.rs
rosswilson2306: > How about the Ragflow error-handling bits? They are provider-specific, are we ok with leaving those?

Yeah, exactly. That's specific to ragflow so fine to leave that. I did setup something in the client to handle these annoying errors but maybe it's not being used here because it's coming from a method that doesn't necessarily return json https://github.com/crownshy/comhairle/blob/460d2117184eb02a5665503ebe2260a7b8e41bd4/adaptors/ragflow/src/client.rs?plain=1#L39-L61.

api/src/routes/documents.rs
rosswilson2306: Okay, that makes sense. But yeah, we might need to think about it before the gaelic / afghan stuff.

api/src/routes/documents.rs
rosswilson2306: Ha, I assumed we had a router for the learn tool but looks like it's empty. Thanks for making the ticket. We can have a think about it when someone picks it up.

rosswilson2306 (approved):

api/src/routes/documents.rs
dtodorowa: oh coool! added this https://github.com/crownshy/comhairle/issues/865

PR #830, opened 2026-08-07
783 Learning Assistant: open sources in the PDF viewer, with the passage highlighted

### LINKED ISSUE #783: Learning Assistant: show sources in the PDF viewer (and highlight the passage)
When the Learning Assistant answers a question, the retrieved sources appear as filename badges under the answer text. Clicking a badge today opens a plain text modal showing the retrieved excerpt (`activeChunk.content`, HTML stripped) in `LearningAssistant.svelte`. The actual document is never shown.

We want to open the real PDF in the existing viewer instead, and ideally highlight the passage the answer came from.

The good news: the viewer already exists and this is mostly wiring.

## Phase 1: open the PDF in the viewer (quick win)

Replace the text-excerpt modal with the existing PDF viewer.

- Viewer already built on pdfjs-dist: `src/lib/components/PdfViewer/PdfViewer.svelte` + modal wrapper `PdfDocumentDialog.svelte`.
- Already wired for a separate feature (rich-text article source badges) in `RichTextEditor/ContentRenderer/ContentRenderer.svelte`, so there is a working pattern to copy.
- Each source chunk already carries `document_id` (`ReferenceChunk` in `src/lib/api/chatClient.svelte.ts`), and PDFs are served at `/api/conversation/{conversationId}/documents/{document_id}/download`.
- Work: point the badge click at `PdfDocumentDialog` with `src` = that download URL instead of setting `activeChunk`. Small, contained change.

## Phase 2: highlight the passage in the PDF (stretch, pending backend check)

- `ReferenceChunk` has a `positions?: number[][]` field (RAGFlow-style bounding boxes, roughly `[page, x0, x1, top, bottom]`) that is present in the type but unused today.
- Work: (a) confirm the backend actually populates `positions` for these chunks before building on it; (b) `PdfViewer` currently renders canvas + text layer only, with no highlight overlay, so add jump-to-page plus a highlight rectangle scaled to the viewport.
- Risk: if `positions` is not populated, phase 2 needs backend work first.

## Acceptance
- Phase 1: clicking a Learning Assistant source badge opens the document in the PDF viewer rather than the text modal.
- Phase 2: the viewer opens on the right page with the relevant passage highlighted (dependent on `positions` being available).

## Key files
- `src/lib/components/LearningAssistant/LearningAssistant.svelte`
- `src/lib/api/chatClient.svelte.ts` (`ReferenceChunk`, `positions`)
- `src/lib/components/PdfViewer/PdfViewer.svelte`, `PdfDocumentDialog.svelte`
- `src/lib/components/RichTextEditor/ContentRenderer/ContentRenderer.svelte` (reference wiring)

dtodorowa:
Source citations in the learning assistant now open the real document in the PDF viewer, scrolled to the passage and highlighted, instead of a stripped-text modal. Works for fresh and reloaded answers alike.

## What

- **Source badges** under an answer open the real PDF in the existing viewer, on the right page with an amber highlight over the passage.
- **Inline ⓘ citations** are clickable too, each opening its own exact chunk's passage (more precise than the per-document badge).
- The **hover popover** now shows a short preview plus an "Open in document" action, replacing the old "See more" that just expanded raw extraction text.
- Highlighting works on **reload as well as live**, including answers created before this change.

## How the reload part works

Live answers carry passage coordinates (`positions`) over the stream, but RAGFlow's stored session history strips them, which is why reloaded answers couldn't highlight. Positions are a static property of a chunk (where it sits in the document, keyed by its stable id), not of the question, so on history load the backend re-fetches them from RAGFlow's chunk store and re-attaches them by chunk id. That fixes every old answer retroactively, not just new ones. It's best-effort: if the chunk lookup fails, you just get no highlight, never a broken load. Cost is one RAGFlow call per cited document on reload (paginated, capped).

## Main pieces

- **Viewer**: `PdfViewer` gains a highlight overlay + jump-to-page; `PdfDocumentDialog` threads `highlights`/`page` through; new `highlights.ts` converts RAGFlow `positions` to rectangles (with unit tests).
- **Citations**: `LearningAssistant` opens the viewer from badges and inline citations; `MessageWithReferences` / `ReferencePopoverContent` make citations clickable.
- **Reload positions**: adaptor `list_chunks` + backend `enrich_reference_positions` re-attach positions on history load; `positions` added to the reference type; frontend carries it through.

## Notes

- Page numbers are treated as 1-based (matches RAGFlow's payloads). Worth an eyeball, but it's a one-line flip in the converter if ever off.
- Adds `positions` to the reference type, so the regenerated `@crownshy/api-client` (`api.ts`) is included in this PR. `open-api-spec.json` is gitignored, so it's not part of the diff.
- Enrichment is best-effort by design: a RAGFlow chunk-lookup failure degrades to "no highlight," never a failed history load.

Old:
image

New (actually use the pdf viewer)

image

### DIFF
--- a/adaptors/ragflow/src/document.rs
+++ b/adaptors/ragflow/src/document.rs
@@ -109,6 +109,44 @@ pub async fn stop_parse(
     client.delete(&path, Some(&body), None).await
 }
 
+/// List a document's chunks. RAGFlow keeps each chunk's `positions` here (the
+/// coordinates of the passage in the source PDF) even though the stored session
+/// history strips them, so this is how we re-attach highlight coordinates on
+/// reload. See issue #783.
+pub async fn list_chunks(
+    client: &RagflowClient,
+    dataset_id: &str,
+    document_id: &str,
+    query: Option<GetQueryParams>,
+) -> Result<(StatusCode, ChunkList)> {
+    let path = format!("/datasets/{dataset_id}/documents/{document_id}/chunks");
+    let (status, value) = client.get(&path, query.as_ref(), None).await?;
+
+    let json: GetChunksResponse = serde_json::from_value(value)?;
+
+    Ok((status, json.data))
+}
+
+#[derive(Deserialize)]
+pub struct GetChunksResponse {
+    pub data: ChunkList,
+}
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct ChunkList {
+    #[serde(default)]
+    pub chunks: Vec<DocumentChunk>,
+}
+
+/// A document chunk from the chunk-listing endpoint. Only the fields we need to
+/// map a citation back to its passage; RAGFlow returns more.
+#[derive(Serialize, Deserialize, Default, Clone)]
+pub struct DocumentChunk {
+    pub id: String,
+    #[serde(default)]
+    pub positions: Option<Vec<Vec<f64>>>,
+}
+
 #[derive(Serialize, Deserialize)]
 pub struct GetDocumentsResponse {
     pub code: i32,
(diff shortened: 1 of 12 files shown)


adaptors/ragflow/src/document.rs
```diff
@@ -109,6 +109,44 @@ pub async fn stop_parse(
...
+    pub chunks: Vec<DocumentChunk>,
+}
+
+/// A document chunk from the chunk-listing endpoint. Only the fields we need to
+/// map a citation back to its passage; RAGFlow returns more.
+#[derive(Serialize, Deserialize, Default, Clone)]
+pub struct DocumentChunk {
+    pub id: String,
+    #[serde(default)]
+    pub positions: Option<Vec<Vec<f64>>>,
```
alpinegeorge, line 147: Should this be an `Option<Vec<Vec<{integer}>>>` to avoid conversion errors, or is it truly a float value?
https://github.com/infiniflow/ragflow/blob/8e37f5543af860e9de99a68cea7ea7c315d2206a/api/apps/restful_apis/chunk_api.py#L111

dtodorowa: ran this by the gpts and claudios and the reasoning was: they're ints on ragflow's side, but ragflow passes it through with no cast, so if it ever sent a float back an int type would blow up on parse and we'd lose the highlight. f64 reads both fine, and coords are tiny + the highlight math is fractional anyway so no real downside. 

that said im cool with changing to i32 if you'd rather have the stricter type? not super suree?

alpinegeorge: If they're guaranteed to be small offsets, then that's cool

alpinegeorge (approved):

PR #854, opened 2026-08-11
Civic os places migration

alpinegeorge:
Motivation:

- We would like to migrate our regions.ts data to the database soon in
       civic_os and we will need appropriate data fields to handle the new
       data.

Actions:

- Created regions_area table which will hold zipcode areas for now, but may be migrated to use generic PostGIS geometry in the future.
- Added metadata fields to regions, organizations and events to support extra civic_os use cases.

### DIFF
--- a/api/src/models/event.rs
+++ b/api/src/models/event.rs
@@ -177,6 +177,7 @@ pub struct Event {
     pub format: EventFormat,
     #[partially(transparent)]
     pub location: Option<EventLocation>,
+    pub metadata: Option<serde_json::Value>,
     #[partially(omit)]
     pub created_at: DateTime<Utc>,
     #[partially(omit)]
@@ -324,7 +325,7 @@ impl LocalizedEvent {
     }
 }
 
-const DEFAULT_COLUMNS: [EventIden; 16] = [
+const DEFAULT_COLUMNS: [EventIden; 17] = [
     EventIden::Id,
     EventIden::Name,
     EventIden::Description,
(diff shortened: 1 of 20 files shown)


api/src/routes/regions.rs
```diff
@@ -98,8 +135,62 @@ async fn delete(
     Ok((StatusCode::OK, Json(region)))
 }
 
+#[instrument(err(Debug), skip(state))]
```
alpinegeorge: This can be removed, since this will belong in metadata.

api/src/models/event.rs
```diff
@@ -510,6 +515,55 @@ pub async fn update(
...
+    let (metadata,) = sqlx::query_as_with::<_, (Option<serde_json::Value>,), _>(&sql, values)
+        .fetch_one(db)
+        .await?;
+
+    Ok(metadata)
+}
+
+/// Merge the supplied object into the event's `metadata` jsonb column at the
+/// top level. Existing keys are overwritten by the patch.
+pub async fn patch_metadata(
```
rosswilson2306, line 541: Can you slap `#[instrument(err(Debug))]` on these methods?

rosswilson2306 (dismissed):
I left a comment about instrumentation but it seems like some of this might be temporary solutions until we can proper spatial data with postgis so feel free to ignore. I've approved.

api/src/models/event.rs
alpinegeorge: Done

rosswilson2306 (approved):

PR #864, opened 2026-08-12
671/exp with translations

rosswilson2306:
Motivation:

- allow translations within nested JSON structures, eg workflow step tool configs, to allow frontend facing config to be translatable (ie prioritization tool questions and section questions)

Actions:

- convert prioritization tool text fields to TextContentIds
- new TranslatableJson macro which provides localization functionality for public site on fields within any nested structure marker as `#[translatable]`, as well as sending colocated translations for admin UI
- traits and helper functions for resolving translations from nested json structures
- trait for building translations from string values (used in setup of prioritization tool)
- admin UI refactor to integrate colocated translations for prioritization tool


dtodorowa:
got some weird sporadic behaviour with translations (towards the end), will try to debug soon

https://github.com/user-attachments/assets/75a9a0c2-6a42-4355-aeae-9c0d5017ca4c

rosswilson2306:
@dtodorowa 
> got some weird sporadic behaviour with translations (towards the end), will try to debug soon

I can't seem to recreate that. Maybe we can jump on a call quickly to try and recreate?

dtodorowa (approved):
works nicely i think, let's merge and test on staging

One cool follow up could be: if we have a prioritization step and have prepared all the text, adding a new language should auto-translate all the fields, so then translators mostly have to click on "approve" rather than have to click the "Translate" button a few times rather than click the button themselves

#906 added it here, but this is low priority IMO

dtodorowa:
Would consider this as a follow-up too (or just accept it as a limitation) because it's a super edge case and i found it super randomly

If admin has their language set to Spanish and the source language of a conversation is English (+Spanish supported lang), then trying to add proposals doesn't work

I think from what I can derive so far, source language and UI language have to be the same 

https://github.com/user-attachments/assets/4438d4c7-aa9d-44eb-8711-ddedb8c16a94


Created a ticket with more info to capture it
https://github.com/crownshy/comhairle/issues/905

alpinegeorge (approved):

Issue #889, opened 2026-08-13
Extract a pluralize helper for the '=== 1 ? "" : "s"' idiom


dtodorowa:
### What

We repeat the same inline pluralization all over the frontend:

```ts
`${n} statement${n === 1 ? '' : 's'}`
```

There are ~10 call sites and no shared helper. A small util (e.g. `pluralize(count, singular, plural?)` in `src/lib/utils/`) would remove the duplication and read better at the call site.

### Known call sites (non-exhaustive)

- `src/lib/tools/polis/PolisModeration.svelte` (2x)
- `src/lib/tools/polis/polis-moderation/StatementModerationRow.svelte`
- `src/lib/tools/polis/polis-moderation/SplitStatementDialog.svelte`
- `src/lib/tools/polis/polis-moderation/AddSeedStatementsDialog.svelte`
- `src/lib/tools/thinking_space/QuestionFlow.svelte`
- `src/lib/components/LiveEvent/CreateBreakoutDialog.svelte`
- `src/routes/(admin)/admin/conversations/[conversation_id]/configure/GlossaryEditor.svelte` (2x)
- `src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventRecordings.svelte`

Grep `? '' : 's'` for the full list.

### Why it's its own ticket

Surfaced during review of the statement-splitting PR (#806). Left out of that PR deliberately: it's a codebase-wide idiom, so touching all sites belongs in its own change rather than diverging just two of them.

### Notes

- Check for an existing helper first before adding one.
- Some sites use irregular plurals or trailing context, so the helper should take an optional explicit plural form.

Issue #834, opened 2026-08-07
Optional reason field when rejecting a statement?


dtodorowa:
Spun out of the #806 split exploration. Right now when an admin rejects a statement in Polis moderation there's nowhere to note why.

The backend already supports it: `polis_statement_aux.moderation_reason` exists and the moderate endpoint already takes an optional `moderation_reason`. What's missing is only a UI to type one on a manual reject.

**Question for the team:** do we want to let admins add an optional reason when rejecting?

Things to decide:

- Keep it optional and unobtrusive. We don't want to force a reason on every reject and slow moderation down. (Small optional field? Popover on reject?)
- Where does the reason show: mod log / audit only, or ever surfaced to participants? (probably not participants)
- Per-statement reject vs bulk reject: does bulk get a shared reason, or skip it?

Not blocking the #806 split work; that flow sets the reason automatically.

PR #890, opened 2026-08-13
806 Split / reword composite statements without turning them into seeds

### LINKED ISSUE #806: Spike: splitting composite Polis statements without them becoming seeds
Part of the Polis moderation line of work (epic #792). **Spike / exploration - no design yet.**

## The problem

When a policy team moderates a Polis poll, sometimes a participant submits a **composite statement** - one statement that bundles two separate ideas. Standard Polis moderation practice is to **split it into two** clean, separately-votable statements. We have no way to do this today, and the naive way of doing it has a nasty side effect.

The only "admin adds a statement" path we have is `post_seed_comment`, which hard-codes `is_seed: true` when it POSTs to Polis (`api/src/wiki_poll_service/polis_service.rs:612`). So if an admin adds the two split statements through the current tooling, they land as **seed statements** - misrepresenting them as host-authored, when they actually came from a participant. That is the thing we want to avoid.

## What we already know (so the spike doesn't re-derive it)

- Admin/seed statements post server-side with `"pid": "mypid"`, which Polis resolves to the poll's admin participant. There is **no** "admin = pid 0" convention in our code; the `pid ?? 0` in `PolisApi.ts:152` is just a client-side JS fallback, not meaningful.
- Adding a statement today forces `is_seed: true` (`polis_service.rs:612`). Whether Polis lets a host post a comment with `is_seed: false` is untested.
- The aux table (`polis_statement_aux`) already stores a "which statement was this in reference to" pointer: `visible_statement_when_submitted` (`api/src/models/polis_statement_aux.rs:34`). There is no edit/split lineage column (no `original_statement_id`).
- There is currently no path to edit a statement's text either. `update_statement_aux` (`polis_statement_aux.rs:195`) rewrites only the local aux row and never propagates to Polis; no UI calls it.
-  Does Polis even allow a statement's text to be edited? No, "split" means **add new statement(s)**, not edit in place.
## Open questions

**Backend / Polis mechanics**

- [ ] When an admin posts a statement, what `pid` and `is_seed` does it actually get in Polis? Can we post a host-authored statement that is **not** flagged `is_seed`? (I think pid 0, and is_seed - tru)
- [ ] For a split, what happens to the original composite statement - reject it, keep it, mark it superseded? What do participants who already voted on the original see, and does splitting break vote continuity / opinion groups?
- [ ] Do we want to record lineage - e.g. an `original_statement_id` (and/or author pid) on `polis_statement_aux` so a split statement points back at the participant statement it came from, so it is never misread as a host seed?

**Admin-end UX**

- [ ] What does splitting a composite statement look like in the moderation interface? (e.g. a "split" action on a row that opens the original text and lets the admin author 2+ replacements.)
- [ ] How is a split/derived statement labelled vs a real host seed vs a raw participant statement, both in the moderation list and to participants?
- [ ] Do we want to allow plain **statement editing** at all as a separate feature, or is split the only case we care about? (Open product question.)

## Deliverable

- A recommended direction: what "split" does in the backend (Polis interaction + any `polis_statement_aux` change) and what it looks like on the admin end. Throwaway prototype/sketch is fine.
- A follow-up build ticket.

## Out of scope

- Building the feature.
- General moderation rework (that is the #792 line).

## Related

- #792 open question: "Are host preview statements classed as seed statements by Polis?"
- #797 (stale aux rows at launch), ADR-0003 (native Polis admin proxy).

dtodorowa:
Exploratory spike PR off #806. Gives moderators a way to split a composite participant statement ("cats and dogs are great") into separately votable parts, or reword an ambiguous one, while keeping the replacements honestly attributed as participant-derived rather than host seeds. Design rationale lives in ADR-0011 (see docs)

### Why

Today the only "admin adds a statement" path hardcodes `is_seed: true`, so the naive split would misrepresent participant-derived content as host-authored, a provenance lie. Pol.is has no edit-in-place, so a reword is unavoidably a new statement plus a rejection of the old one.

### What it does

A split (one operation, one or many replacements) runs server-side in this order:

1. **Posts** each replacement with `is_seed: false` (real, votable, never a host seed).
2. **Auto-accepts** each one so it's immediately votable instead of littering the mod queue, the admin already approved it by authoring it.
3. **Rejects** the original composite last, recording a system note in `moderation_reason`.
4. **Records lineage** via a new self-referential `original_statement_id` column. Discriminator: `is_seed: false` + `original_statement_id` set ⇒ derived; `is_seed: true` ⇒ seed; neither ⇒ raw participant.

Ordering is fail-safe: replacements exist and are accepted before the original is rejected, so a partial failure never destroys a statement. Local aux bookkeeping is a single DB transaction; the cross-system Pol.is sequence can't be (no transactions there).

### Changes

- **Migration** adds `original_statement_id UUID REFERENCES polis_statement_aux(id)` (`ON DELETE SET NULL`).
- **Model** (`polis_statement_aux.rs`): `CreateDerivedStatement` + `record_split` (atomic local write); sync preserves `original_statement_id` as a comhairle-only field, like `themes` / `moderation_reason`.
- **Endpoint** (`tools/polis.rs`, `polis_service.rs`): the split writes the derived aux rows itself from the tids Pol.is returns, rather than post-then-resync.
- **UI**: `SplitStatementDialog` (author replacements against the original + its viewed context), plus an "Edited" badge and "Edited from" / "Replaced by" lineage labels in the moderation table.

### Notes / provisional

- Admin-facing labelling ("Edited" badge, original ↔ replacements linking) is provisional per the ADR and may move.
- The separate question of an optional reason on a plain reject is tracked in #834.
- Follow-up cleanup (pluralize helper) filed as #889.

Admin editing:

https://github.com/user-attachments/assets/4a9d7f4e-e833-42b2-8c43-2a015ab8436a

Then statements show as non-seeds on the participant end:

https://github.com/user-attachments/assets/2895b66f-5eff-4c40-acf7-244e03db56bd


api/src/models/polis_statement_aux.rs
```diff
@@ -131,6 +136,121 @@ pub async fn create(
     Ok(aux)
 }
 
+/// A derived statement to insert as part of a split: an admin-authored
+/// replacement (`is_seed = false`, auto-accepted) that points back at the
+/// original participant statement via `original_statement_id`. `user_id` is
+/// left NULL: the admin authored it, so it is not attributed to a participant.
```
rosswilson2306, line 142: @dtodorowa Just curious why you wouldn't carry over the user_id from the original statement / comment if they were the source? Is the `zid` more important for what participant created the polis statement?

api/src/wiki_poll_service.rs
```diff
@@ -62,6 +62,19 @@ pub trait WikiPollService: Send + Sync {
         auth_cookies: &str,
     ) -> Result<String, WikiPollServiceError>;
 
+    /// Post a host-authored statement, choosing whether it is a seed.
+    /// `is_seed = true` matches `post_seed_comment` (auto-approved by Polis,
+    /// `mod: 1`); `is_seed = false` posts a genuine non-seed statement, which
+    /// Polis leaves pending (`mod: 0`) until moderated. Returns the new `tid`
```
rosswilson2306: Maybe take out the stuff about `Polis` from here as this is the provider agnostic part (or attempts to be at least!)

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
```diff
@@ -179,6 +180,46 @@
...
+	const lineage = $derived.by(() => {
+		const byId = new Map(statements.map((s) => [s.id, s]));
+		const replacementsByOriginal: Record<string, string[]> = {};
+		for (const s of statements) {
+			if (s.original_statement_id) {
+				(replacementsByOriginal[s.original_statement_id] ??= []).push(s.statement_text);
+			}
+		}
+		const result: Record<string, { editedFrom?: string; replacedBy?: string[] }> = {};
+		for (const s of statements) {
```
rosswilson2306, line 202: @dtodorowa Couldn't these two for loops be combined into one?

api/src/models/polis_statement_aux.rs
dtodorowa: so i think this came from trying to capture what Stuart was saying (but there is a potential that I got it wrong, and it's good to double check)

the reason is that this new statement(s) is the moderator's words and attributing that to the participant would be a bit shady. We point back to their original statement, but we don't want to put words in their mouth (bad way of saying it but i hope you get the point)


(also, adding another thing, if anyone is looking at this conversation and wondering what zid is > the `zid` column is a bit of a misnomer because it stores the Polis participant id (pid), not a Polis conversation id (the conversation lives in polis_conversation_id))

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
dtodorowa: i don't think that the the two for loops can fully merge bc the second reads `replacementsByOriginal[s.id]`, which isn't complete until the first has seen every statement (a replacement can appear before or after its original, maybe this is bad design hm)

`byId` is a needless extra pass tho, so i can fold it in the first loop. Also, lmk if you would prefer to have it structured differently, i'm open to ideas

api/src/wiki_poll_service.rs
dtodorowa: good point! reworded

ui/packages/comhairle/src/lib/tools/polis/PolisModeration.svelte
rosswilson2306: Right, fair enough. I missed that. All good!

rosswilson2306 (approved):

PR #895, opened 2026-08-13
[Survey Insights] Should show the number of responses on each question

### LINKED ISSUE #887: [Survey Insights] Should show the number of responses on each question
### Description
Currently the questions don't show the number of responses on them, this is because based on the question type, calculating the number of responses can be difficult, so it was spun out from the main ticket #766 

### Acceptance criteria
- [x] Questions should accurately show the number of responses

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/887

### Testing
image

### DIFF
--- a/adaptors/heyform-rust-sdk/src/types.rs
+++ b/adaptors/heyform-rust-sdk/src/types.rs
@@ -219,8 +219,8 @@ pub struct FormReportResponse {
     pub id: String,
     pub kind: Option<String>,
     pub title: Option<String>,
-    pub total: i32,
-    pub count: i32,
+    pub total: u32,
+    pub count: u32,
     pub average: f64,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub chooses: Option<Vec<serde_json::Value>>,
@@ -289,7 +289,7 @@ pub struct Submission {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Submissions {
-    pub total: i32,
+    pub total: u32,
     pub submissions: Vec<Submission>,
 }
 
--- a/api/src/tools/heyform.rs
+++ b/api/src/tools/heyform.rs
@@ -727,11 +727,7 @@ pub fn build_survey_insights(
             // Choice-based questions already have a full aggregate breakdown
             // via `choices`; for everything else (short_text, opinion_scale,
             // etc.) the per-answer values are the only meaningful content.
-            let submissions = if choices.is_none() {
-                submissions_by_field.get(&response.id).cloned()
-            } else {
-                None
-            };
+            let submissions = submissions_by_field.get(&response.id).cloned();
 
             let kind = field
                 .map(|f| f.kind.clone())
--- a/api/src/tools/heyform.rs
+++ b/api/src/tools/heyform.rs
@@ -452,7 +452,7 @@ pub async fn submissions(
     Ok((
         StatusCode::OK,
         Json(Submissions {
-            total: submissions.len() as i32,
+            total: submissions.len() as u32,
             submissions,
         }),
     ))
@@ -505,9 +505,9 @@ pub struct InsightQuestion {
     /// Falls back to the field ID when the schema contains no plain-text title.
     pub title: String,
     /// Number of submissions that have been sent through.
-    pub count: u64,
+    pub count: u32,
     /// Total number of responses recorded for this question.
-    pub total: i32,
+    pub total: u32,
     /// Field properties
     #[serde(skip_serializing_if = "Option::is_none")]
     pub properties: Option<HashMap<String, serde_json::Value>>,
@@ -729,7 +729,7 @@ pub fn build_survey_insights(
                 .get(&response.id)
                 .unwrap_or(&Vec::new())
                 .iter()
-                .fold(0u64, |acc, submission| {
+                .fold(0u32, |acc, submission| {
                     let val = submission.value.to_string();
                     // Empty values from HeyForm
                     if val != "\"\"" && val != "{\"value\":[]}" {
--- a/api/src/tools/heyform.rs
+++ b/api/src/tools/heyform.rs
@@ -504,6 +504,8 @@ pub struct InsightQuestion {
     /// Human-readable question title extracted from the form schema.
     /// Falls back to the field ID when the schema contains no plain-text title.
     pub title: String,
+    /// Number of submissions that have been sent through.
+    pub count: u64,
     /// Total number of responses recorded for this question.
     pub total: i32,
     /// Field properties
@@ -642,6 +644,7 @@ pub fn build_survey_insights(
             if let Some(field_id) = answer.get("id").and_then(|v| v.as_str()) {
                 if let Some(val) = answer.get("value") {
                     if !val.is_null() {
+                        let str_val = &val.to_string();
                         let key = (field_id.to_string(), sub_id.clone());
                         if seen_answers.insert(key) {
                             submissions_by_field
@@ -723,11 +726,29 @@ pub fn build_survey_insights(
 
             let choices = resolve_choices(response, &field_choices_by_id);
 
+            let count = submissions_by_field
+                .get(&response.id)
+                .unwrap_or(&Vec::new())
+                .iter()
+                .fold(0u64, |acc, submission| {
+                    let val = submission.value.to_string();
+                    // Empty values from HeyForm
+                    if val != "\"\"" && val != "{\"value\":[]}" {
+                        return acc + 1;
+                    }
+
+                    acc
+                });
+
             // Attach individual submission answers for non-choice questions.
             // Choice-based questions already have a full aggregate breakdown
             // via `choices`; for everything else (short_text, opinion_scale,
             // etc.) the per-answer values are the only meaningful content.
-            let submissions = submissions_by_field.get(&response.id).cloned();
+            let submissions = if choices.is_none() {
+                submissions_by_field.get(&response.id).cloned()
+            } else {
+                None
+            };
 
             let kind = field
                 .map(|f| f.kind.clone())
@@ -739,6 +760,7 @@ pub fn build_survey_insights(
                 id: response.id.clone(),
                 kind,
                 title,
+                count,
                 total: response.total,
                 properties,
                 choices,
--- a/api/src/tools/heyform.rs
+++ b/api/src/tools/heyform.rs
@@ -504,8 +504,8 @@ pub struct InsightQuestion {
     /// Human-readable question title extracted from the form schema.
     /// Falls back to the field ID when the schema contains no plain-text title.
     pub title: String,
-    /// Number of submissions that have been sent through.
-    pub count: u32,
+    /// Number of times this question was answered.
+    pub answered: u32,
     /// Total number of responses recorded for this question.
     pub total: u32,
     /// Field properties
@@ -725,7 +725,7 @@ pub fn build_survey_insights(
 
             let choices = resolve_choices(response, &field_choices_by_id);
 
-            let count = submissions_by_field
+            let answered = submissions_by_field
                 .get(&response.id)
                 .unwrap_or(&Vec::new())
                 .iter()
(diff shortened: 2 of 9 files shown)


ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
ADSmith-0, line 1: Refactored this a bit to make it hopefully a bit easier to follow

ui/packages/api-client/src/template.hbs
```diff
@@ -75,6 +75,47 @@ const endpoints = makeApi([
 {{/each}}
 ] as const satisfies ZodiosEndpointDefinitions);
 
+// Axios error for failing request
```
ADSmith-0, line 78: Added this type here because without it when you do `pnpm run generateAPI` it will delete this

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -39,14 +39,14 @@
...
-				<!-- 		· -->
-				<!-- 		{Math.round(section.answers.length / section.total) * 100}% Completion -->
-				<!-- 	</h3> -->
-				<!-- </div> -->
+				<div class="flex flex-row">
+					<h3 class="text-muted-foreground mr-10 text-sm">
+						{section.count}
+						{section.count === 1 ? 'response' : 'responses'}
+						·
+						{Math.round((section.count / section.total) * 100)}% Completion
```
dtodorowa: if total is 0 here, would this render `NaN% Completion`? we can maybe add a guard hiding this when there are no responses?

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -39,14 +39,14 @@
...
-				<!-- 	<h3 class="text-muted-foreground mr-10 text-sm"> -->
-				<!-- 		{section.answers.length} -->
-				<!-- 		{section.answers.length === 1 ? 'response' : 'responses'} -->
-				<!-- 		· -->
-				<!-- 		{Math.round(section.answers.length / section.total) * 100}% Completion -->
-				<!-- 	</h3> -->
-				<!-- </div> -->
+				<div class="flex flex-row">
+					<h3 class="text-muted-foreground mr-10 text-sm">
+						{section.count}
```
dtodorowa: is `count` always meant to match the answers we chart below? noticed a rating value of 0 counts on the server but gets dropped by the `!!s` filter, so the header could read one more than the plot. (ignore if this was intentional, just trying to understand better)

api/src/tools/heyform.rs
```diff
@@ -723,6 +725,20 @@ pub fn build_survey_insights(
...
             let choices = resolve_choices(response, &field_choices_by_id);
 
+            let count = submissions_by_field
+                .get(&response.id)
+                .unwrap_or(&Vec::new())
+                .iter()
+                .fold(0u64, |acc, submission| {
+                    let val = submission.value.to_string();
+                    // Empty values from HeyForm
+                    if val != "\"\"" && val != "{\"value\":[]}" {
```
dtodorowa, line 735: curious how brittle matching on the stringified JSON is vs checking the parsed value. e.g. if a new empty shape shows up would this silently over-count

but probably not so important, just flagging because brain went in that direction

api/src/tools/heyform.rs
```diff
@@ -723,6 +725,20 @@ pub fn build_survey_insights(
 
             let choices = resolve_choices(response, &field_choices_by_id);
 
+            let count = submissions_by_field
+                .get(&response.id)
+                .unwrap_or(&Vec::new())
```
dtodorowa, line 730: does this allocate a throwaway `Vec` on every question even when the key exists?

`.into_iter().flatten()` might be able to do the same job without the temp? but need advice from the rust pros, i don't know rust too well

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
ADSmith-0: This is what it looks like with no responses 
image

I guess some of the info is redundant we can hide some of it if we want?

api/src/tools/heyform.rs
```diff
@@ -504,6 +504,8 @@ pub struct InsightQuestion {
     /// Human-readable question title extracted from the form schema.
     /// Falls back to the field ID when the schema contains no plain-text title.
     pub title: String,
+    /// Number of submissions that have been sent through.
+    pub count: u64,
```
dtodorowa: any reason count is `u64` while total right below is `i32`?

dtodorowa: nitpick and u can ignore, but naming this `answered` might read clearer, since it's the "non-empty answers" rather than a raw count?

ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
```diff
@@ -159,18 +147,22 @@ function transform(insight: InsightQuestion): SurveyQuestion | undefined {
 
 			for (const submission of insight.submissions) {
 				const value = submission.value as HeyFormLegalTermsValue;
-				answers[Number(value)].count += 1;
+				// value `true` = 1
+				// 1 + 1 % 2 = 0
+				// index 0 of answers = "yes"
+				answers[(Number(value) + 1) % 2].count += 1;
```
dtodorowa: the middle line reads `1 + 1 % 2 = 0`, but doesn't `%` bind tighter so it's actually 2? wondering if a one-liner on why yes/no is flipped would land better than tracing the maths? wdyt? but i'm cool either way

ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
```diff
@@ -181,100 +173,98 @@ function transform(insight: InsightQuestion): SurveyQuestion | undefined {
...
-							const fullName = s.value as HeyFormFullNameValue;
-							return `${fullName.firstName} ${fullName.lastName}`;
-						})
-						.filter((s) => !!s.trim()) ?? []
-			});
+			return (
+				insight.submissions
+					?.map((s) => {
+						const fullName = s.value as HeyFormFullNameValue;
+						if (fullName === '') return '';
```
dtodorowa: this `=== '' return ''` guard shows up in full_name, address, date_range and file_upload

we could pull it into a small helper? or do you prefer them inline per-case for readability?

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
ADSmith-0: Oh yeah that's a really good catch! I think in my head I was thinking it's just for `undefined` values (I think I wrote this before realising what HeyForm picks as empty values), but yeah you're totally right this would target any falsy value which means 0 would be dropped, changed it, thank you! :smile:

dtodorowa: so fast tehee! nice

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
dtodorowa: sweeetoo! thanks for confirming and my bad for commenting before checking the UI

api/src/tools/heyform.rs
ADSmith-0: Yeah I was thinking about parsing it, but I think it would be quite hard with serde because you'd have to try and parse it as a string and if that fails then you'd have to parse it as an object, then you'd have to grab the key and then parse the value as an array, so I thought this might be simpler, because I think the "empty" types for HeyForm are fairly consistent

I think that these values should only ever show up if the user skips a question?

dtodorowa (dismissed):
would love to test on staging and get people's eyes on it

api/src/tools/heyform.rs
dtodorowa: yehh makes sense

api/src/tools/heyform.rs
ADSmith-0: No I think this is only when it fails so, if it returns from the `get` it's a `Vec`, but if it's a `None`, then it will create a `Vec` I think it makes sure that the type for the next part of the chain is consistent as a `Vec`, as otherwise I don't know the `iter()` could call as it would be an Option<Vec> if that makes sense?

I'm the same I don't know Rust super well would probably be good to have a Rust pro weight in :joy:

api/src/tools/heyform.rs
ADSmith-0: Yeah so I thought from the serde docs you could either have `u64`, `i64` or `i32`, but I think that might be wrong so it might be worth doing it as a `u32`? But yeah basically just because the count can't ever be negative so I think it makes more sense for it to be a u-int rather than an i-int, it might be worth changing over total as well?

ADSmith-0: Yeah that's a really good point actually! I was doing it to try to stick to HeyForm convention, but I think you're right I think `answered` would be a lot nicer I can change that over it's a quick change :slightly_smiling_face:

adaptors/heyform-rust-sdk/src/types.rs
```diff
@@ -219,8 +219,8 @@ pub struct FormReportResponse {
     pub id: String,
     pub kind: Option<String>,
     pub title: Option<String>,
-    pub total: i32,
-    pub count: i32,
+    pub total: u32,
+    pub count: u32,
```
ADSmith-0, line 223: Changed so all `total` and `count` are `u32` as it shouldn't be possible to get a negative number?

ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
ADSmith-0: Yeah that's a fair point I wasn't sure I could explain it simply, but I've given it a go, let me know what you think!

dtodorowa (dismissed):

ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
ADSmith-0: Yeah that's a good point! I think it would probably make it more obvious what's happening if we put it in a helper, just changed it! :slightly_smiling_face:

dtodorowa:
looks good to me, would be nice to have some rust pros (@rosswilson2306 @alpinegeorge) have a look at it, but i'm happy to merge and test a bit on staging

api/src/tools/heyform.rs
dtodorowa: ohh gotcha!

ADSmith-0:
@dtodorowa Sorry I just realised I could do the boolean logic bit way simpler so that invalidated the approval, but also happy to wait until Ross or George take a look before merging it :slightly_smiling_face:

alpinegeorge (approved):

api/src/tools/heyform.rs
rosswilson2306: I think this is fine but you would typically use `.unwrap_or_default()`, which is definitely not eagerly evaluated.

api/src/tools/heyform.rs
rosswilson2306: You couldn't use an enum to deserialize it into a strict type? Eg https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b1a52c07f2aaf8d7e86ab1dcb987d5d6

api/src/tools/heyform.rs
ADSmith-0: When I tried to use it it gave me this error:
image

And I wasn't super sure how to fix it

rosswilson2306: Probably not a big deal but I think:
```
let answered = submissions_by_field.entry(response.id).or_default()...
```
is probably the way to do it. But I doubt it makes much difference.

rosswilson2306 (approved):

api/src/tools/heyform.rs
ADSmith-0: I tried to give this a go, but for some reason it's not deserialising the objects correctly, I might leave it as is then I can come back to it at another point if that's alright

PR #907, opened 2026-08-17
[Waves] 856 Configurable minimum proposals to review before continuing

### LINKED ISSUE #856: Prioritization: skippable step with configurable minimum required votes
## Summary

Let the prioritization step be **skippable**, with a configurable **minimum number of principles a participant must vote on** before they can continue to the next step. Roughly the Polis / thinking-space model: you don't have to answer all 9, but you do need to engage with at least N.

## Desired behaviour

- The step is skippable at the workflow level.
- Add a config value (tool config) for **minimum required votes** — i.e. how many proposals a participant must submit before "Continue"/"Finish" to the next step unlocks.
- Once the participant has voted on at least that many principles, they can progress to the next step (default target: at least 1 principle, tunable).
- Below the threshold, the affordance to move on should make the requirement clear (e.g. "Vote on at least N proposals to continue").

## Where it lives

`ui/packages/comhairle/src/lib/tools/prioritization/` — config in `types.ts` / `resolveToolConfig`, gating logic in `PrioritizationUser.svelte` (currently `allDone` requires every proposal submitted; there's no "min N" concept).

## Open questions

- Is the threshold a count (N proposals) or a percentage?
- Does "skippable + min votes" interact with the existing per-proposal required questions (likert/continuous)? Presumably a *submitted* proposal still needs its required questions answered; the new gate is just how many submissions are needed to advance.
- Where does the admin set this in the Manage UI?

Design/product to confirm the exact numbers; building for a configurable minimum with a sensible default of 1.

dtodorowa:
Implements #856. The prioritization step becomes skippable, gated on a configurable minimum number of proposals a participant must review before they can continue.

## What changed

**Config.** New `required_reviews: Option<i32>` on `PrioritizationToolConfig`, with a matching admin field in the Manage UI. Debounced save, bounded by `min`/`max` to the proposals that actually exist, and the description tells the admin that a number above that count means all of them.

**The gate.** Lives in `reviewGate.ts` as two pure functions with unit tests rather than inline in the component, because it is the whole contract between the admin's config and the step's Next button:

- `requiredReviewCount(configured, proposalCount)` resolves the admin setting against reality. Unset means every proposal. A set value is clamped to the proposals that exist, so an over-set minimum can never strand a participant on a step they cannot finish.
- `canLeaveStep({ reviewedCount, requiredReviews, proposalsLoaded })` is the gate. `proposalsLoaded` is load-bearing: before proposals arrive the required count is 0, so a plain `reviewed >= required` would open the gate on mount and let a fast participant click straight past the step.

**Default.** Unset means *all* proposals, not 1. This inverts what #856 proposed. Reasoning is in ADR-0015: defaulting to 1 would silently loosen every prioritization step that already exists. The cost is that Waves now has to set 1 by hand, which is worth a product confirmation before this merges.

## Split out of this PR

Two things had been committed onto this branch that were not part of #856:

- The workflow-step double-navigation fix, now #911 below this in the stack.
- The scroll-to-top UX change, now #912 above it.

## Follow-ups, not addressed here

- #913, a pre-existing `$effect` race in the shared step page that can leave Next stuck disabled. Affects every gated tool, so it was left out rather than widening this PR.
- #914, the disabled Next giving no reason when the gate is unmet. The requirement copy exists but sits in the proposal card rather than next to the button, and moving it is a design call on shared `StepHeader`.

## Stack

Merge bottom to top. Use rebase or merge-commit for the lower two: squashing rewrites SHAs and will make the PRs above show duplicated commits until they are rebased.

| | PR | Branch |
|---|---|---|
| 3 | #912 Scroll to page top after each proposal submit | `prioritization-scroll-ux` |
| 2 | **This PR** | `856/prioritization-min-review-gate` |
| 1 | #911 Fix double navigation on step completion | `fix/workflow-step-double-navigation` |

The stack sits on `671/exp-with-translations` (#864).

### DIFF
--- a/api/src/tools/prioritization.rs
+++ b/api/src/tools/prioritization.rs
@@ -46,6 +46,13 @@ pub struct PrioritizationToolConfig {
     pub section_questions: Vec<Question>,
     pub randomize_order: bool,
     pub alignment_question_id: Option<Uuid>,
+    /// Minimum proposals a participant must review before they can continue to
+    /// the next step. `None` means every proposal must be reviewed, which is the
+    /// default; an admin sets a number only to loosen that. Non-positive values
+    /// are normalised back to `None` on read, and the participant UI clamps the
+    /// value to the proposal count, so the gate is always satisfiable.
+    #[serde(default)]
+    pub required_reviews: Option<i32>,
 }
 
 #[derive(Serialize, Deserialize, Debug, JsonSchema, PartialEq, Clone, TranslatableJson)]
@@ -207,6 +214,9 @@ impl ToolConfigSanitize for PrioritizationToolConfig {
             section_questions: self.section_questions.clone(),
             randomize_order: self.randomize_order,
             alignment_question_id: self.alignment_question_id,
+            // A stored zero or negative would make the gate meaningless, so treat
+            // it as unset (every proposal must be reviewed).
+            required_reviews: self.required_reviews.filter(|v| *v >= 1),
         }
     }
 }
@@ -365,6 +375,7 @@ async fn prioritization_setup(
         section_questions: vec![],
         randomize_order: false,
         alignment_question_id,
+        required_reviews: None,
     })
 }
 
--- a/api/src/tools/prioritization.rs
+++ b/api/src/tools/prioritization.rs
@@ -49,8 +49,9 @@ pub struct PrioritizationToolConfig {
     /// Minimum proposals a participant must review before they can continue to
     /// the next step. `None` means every proposal must be reviewed, which is the
     /// default; an admin sets a number only to loosen that. Non-positive values
-    /// are normalised back to `None` on read, and the participant UI clamps the
-    /// value to the proposal count, so the gate is always satisfiable.
+    /// are normalised back to `None` by `ToolConfigSanitize` on save, not on
+    /// every read, and the participant UI clamps the value to the proposal
+    /// count, so the gate is always satisfiable.
     #[serde(default)]
     pub required_reviews: Option<i32>,
 }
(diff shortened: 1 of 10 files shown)


api/src/tools/prioritization.rs
```diff
@@ -207,6 +215,9 @@ impl ToolConfigSanitize for PrioritizationToolConfig {
             section_questions: self.section_questions.clone(),
             randomize_order: self.randomize_order,
             alignment_question_id: self.alignment_question_id,
+            // A stored zero or negative would make the gate meaningless, so treat
+            // it as unset (every proposal must be reviewed).
+            required_reviews: self.required_reviews.filter(|v| *v >= 1),
```
ADSmith-0: Does this return a number? I would have thought that you'd have to do `.len()` or something like that?

api/src/tools/prioritization.rs
```diff
@@ -46,6 +46,14 @@ pub struct PrioritizationToolConfig {
...
     pub randomize_order: bool,
     pub alignment_question_id: Option<Uuid>,
+    /// Minimum proposals a participant must review before they can continue to
+    /// the next step. `None` means every proposal must be reviewed, which is the
+    /// default; an admin sets a number only to loosen that. Non-positive values
+    /// are normalised back to `None` by `ToolConfigSanitize` on save, not on
+    /// every read, and the participant UI clamps the value to the proposal
+    /// count, so the gate is always satisfiable.
+    #[serde(default)]
+    pub required_reviews: Option<i32>,
```
ADSmith-0, line 56: Just a small thing but might be better to do `u32` here? Because the `required_reviews` can't ever be negative? But I'm happy either way!

rosswilson2306 (approved):

PR #917, opened 2026-08-17
Fix: HeyForm iframe height breaks on page refresh at survey step

### LINKED ISSUE #853: HeyForm iframe height breaks on page refresh at survey step
## Summary

The HeyForm survey iframe renders at the wrong height after a **page refresh**. It's fine on the happy path (join the conversation, walk through to the survey step), but if you reload the page while on the survey step, the iframe height is broken.

## Steps to reproduce

1. Join a conversation and get to a HeyForm survey step. Confirm the form sizes correctly.
2. Refresh the page (hard reload) while on that step.
3. The iframe height is messed up.

## Expected

After a refresh, the iframe should auto-size to the active question's content just like it does on first arrival.

## Where it lives

`ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte`

Auto-height is a `postMessage` handshake: our HeyForm fork measures its active question and posts `{ source: 'HEYFORM', eventName: 'FORM_RESIZE', height }`; the embed listens and sets `measuredHeight`. Until a message arrives, `measuredHeight` stays `null` and the iframe falls back to the `min-h-110` fixed height.

## Likely cause (to confirm)

On a fresh navigation the form is loaded uncached and the `message` listener (registered in the `$effect`) is in place before the fork emits, so it works. On refresh the timing shifts, and the resize message appears to be lost:

- The `message` listener is attached in an `$effect`, which runs after mount. On a cached refresh the iframe can fire `load` (and the fork can post `FORM_RESIZE`) before the listener exists, so the height is never applied and it sticks at the `min-h-110` fallback.
- Related: if the fork only emits `FORM_RESIZE` on question change rather than on initial render, a refresh that lands mid-survey never gets an initial measurement.

Worth verifying with a `console.log` in `onFrameMessage` on refresh vs. fresh load to see whether the message is arriving at all.

https://github.com/user-attachments/assets/0cccea28-fab5-49be-bbad-5fad07441b64

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/853

### Testing

https://github.com/user-attachments/assets/6e2ad7e9-d85a-4f2b-a674-5c8a17d4208c

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
+++ b/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
@@ -181,8 +181,8 @@
 			<HeyFormEmbedSkeleton />
 		</div>
 	{/if}
-	<div class="mx-auto mt-1 w-full max-w-2xl overflow-hidden rounded-xl [grid-area:1/1]">
-		{#if browser}
+	{#if browser}
+		<div class="mx-auto mt-1 w-full max-w-2xl overflow-hidden rounded-xl [grid-area:1/1]">
 			<iframe
 				bind:this={iframeEl}
 				src={fullUrl}
@@ -196,6 +196,6 @@
 					? 'opacity-100'
 					: 'opacity-0'}"
 			></iframe>
-		{/if}
-	</div>
+		</div>
+	{/if}
 </div>
--- a/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
+++ b/ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { fade } from 'svelte/transition';
 	import HeyFormEmbedSkeleton from './HeyFormEmbedSkeleton.svelte';
+	import { browser } from '$app/environment';
 
 	type Props = {
 		onDone: () => void;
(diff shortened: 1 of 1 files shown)


ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
```diff
@@ -10,7 +11,7 @@
 		userId: string;
 		extraSurveyParams?: Record<string, string>;
 	};
-	let { onDone, surveyId, userId, surveyURL, serverURL, extraSurveyParams }: Props = $props();
+	let { onDone, surveyId, userId, serverURL, extraSurveyParams }: Props = $props();
```
ADSmith-0, line 14: Says it was unused as a variable

ui/packages/comhairle/src/lib/tools/heyform/HeyFormEmbed.svelte
```diff
@@ -56,6 +57,9 @@
 		pingTimer = setInterval(() => {
 			elapsed += RESIZE_PING_INTERVAL_MS;
 			if (measuredHeight !== null || elapsed >= RESIZE_PING_TIMEOUT_MS) {
+				if (elapsed >= RESIZE_PING_TIMEOUT_MS) {
+					console.warn('Resize timeout reached');
+				}
```
ADSmith-0, line 62: Thought it might be helpful to have a console log if the timeout is reached? Happy to remove it though!

dtodorowa (dismissed):

dtodorowa (approved):
awesome! thanks Adam

alpinegeorge (approved):

PR #918, opened 2026-08-17
752/preassigned breakout rooms

rosswilson2306:
Motivation:

- ensure pre-assigned breakout rooms are maintained in live meetings
- fix bug where editing breakout rooms in admin resulted in empty list of rooms

Actions:

- expose pre-assigned breakout room plan as `initialAssignements` in `CreateBreakoutDialog` for live meetings
- expose attendees / participants to `CreateBreakoutDialog` in admin

### DIFF
--- a/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
+++ b/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
@@ -4,14 +4,14 @@
 	import { Button } from '$lib/components/ui/button';
 	import { notifications } from '$lib/notifications.svelte';
 	import { apiClient } from '@crownshy/api-client/client';
-	import type { BreakoutPlanDto, EventAttendanceDto } from '@crownshy/api-client/api';
+	import type { BreakoutPlanDto, EventAttendanceEtx } from '@crownshy/api-client/api';
 	import CreateBreakoutDialog from '$lib/components/LiveEvent/CreateBreakoutDialog.svelte';
 	import type { VideoCallParticipant } from '$lib/services/videoCallService.svelte';
 
 	type Props = {
 		conversation_id: string;
 		event_id: string;
-		attendees: (EventAttendanceDto & { email?: string })[];
+		attendees: EventAttendanceEtx[];
 	};
 
 	let { conversation_id, event_id, attendees }: Props = $props();
--- a/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
+++ b/ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
@@ -4,16 +4,17 @@
 	import { Button } from '$lib/components/ui/button';
 	import { notifications } from '$lib/notifications.svelte';
 	import { apiClient } from '@crownshy/api-client/client';
-	import type { BreakoutPlanDto } from '@crownshy/api-client/api';
+	import type { BreakoutPlanDto, EventAttendanceDto } from '@crownshy/api-client/api';
 	import CreateBreakoutDialog from '$lib/components/LiveEvent/CreateBreakoutDialog.svelte';
 	import type { VideoCallParticipant } from '$lib/services/videoCallService.svelte';
 
 	type Props = {
 		conversation_id: string;
 		event_id: string;
+		attendees: (EventAttendanceDto & { email?: string })[];
 	};
 
-	let { conversation_id, event_id }: Props = $props();
+	let { conversation_id, event_id, attendees }: Props = $props();
 
 	let plan = $state<BreakoutPlanDto | null>(null);
 	let loading = $state(true);
@@ -180,7 +181,11 @@
 
 <CreateBreakoutDialog
 	bind:open={dialogOpen}
-	participants={[]}
+	participants={attendees.map((a) => ({
+		user_id: a.userId,
+		username: a.email ?? a.userId,
+		role: a.role
+	}))}
 	initialAssignments={dialogAssignments}
 	moderatorIds={dialogModeratorIds}
 	enableModerators={true}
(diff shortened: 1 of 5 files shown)


alpinegeorge (dismissed):

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
```diff
@@ -180,7 +181,11 @@
 
 <CreateBreakoutDialog
 	bind:open={dialogOpen}
-	participants={[]}
+	participants={attendees.map((a) => ({
+		user_id: a.userId,
+		username: a.email ?? a.userId,
```
Comment (dtodorowa) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte:186]: Where does `a.email` come from? I can't find it on `EventAttendanceDto`. Does that mean that it falls back to the raw UUID?

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
```diff
@@ -239,6 +230,32 @@
...
+		participants.forEach((p) => {
+			const roomIndex = roomIndexByUserId.get(p.user_id);
+
+			if (roomIndex !== undefined) {
+				roomAssignments[roomIndex].push(p);
+			}
+		});
+
+		return roomAssignments;
+	}
```
Comment (dtodorowa) [inline ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:257]: What should happen to someone who isn't in any pre-assigned room? Looks like they get
skipped here and then land in room 0 via the `?? 0` (line 375)

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
```diff
@@ -1170,7 +1108,8 @@
 <!-- Dialogs -->
 <CreateBreakoutDialog
 	bind:open={showCreateBreakout}
-	participants={inCallParticipants}
+	participants={allParticipants}
+	initialAssignments={preassignedBreakoutRooms}
```
Comment (dtodorowa) [inline ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:1112]: this keeps empty rooms, but the backend filters them out before saving
`(!room.participants.is_empty() in video_call.rs)`

not sure but it might be shifting the indices so
`getUserBreakoutRoom` sends people to the wrong room? worth confirming?

Comment (dtodorowa) [inline ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:1112]: wonder if we need to pass moderatorIds / enableModerators like the admin dialog does?

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte:186]: So the endpoint this comes from actually returns `EventAttendanceExt`, which is the attendance record with the email as well. https://github.com/crownshy/comhairle/blob/527e242e13bb235502584ff036bb7b67b6ba229a/api/src/routes/event_attendances.rs?plain=1#L231

I didn't think there was one generated by the apiClient but looks like there is, must have just been my lsp playing up. I'll change it.

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:257]: I guess so. I'll test that out. 

What should be the desired behaviour for that case do you think?

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:1112]: > this keeps empty rooms, but the backend filters them out before saving

I'm not sure I follow what you mean? `preassignedBreakoutRooms` is generated from the list breakout rooms `plan` that comes from the backend so where would the empty rooms come from?

Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:1112]: Yeah, maybe. That seems to be related to assigning attendees as moderators. Is that something we want to be able to do on the fly during the call? What do you think?

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:257]: @dtodorowa It looks like the backend ensures that anyone who registers after the breakout rooms are assigned is given a room https://github.com/crownshy/comhairle/blob/527e242e13bb235502584ff036bb7b67b6ba229a/api/src/routes/event_attendances.rs?plain=1#L111

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte
Comment (rosswilson2306) [reply ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/EventBreakoutRooms.svelte:186]: > Does that mean that it falls back to the raw UUID?

This is actually what the backend does as well, falls back to the user id when generating the room dtos https://github.com/crownshy/comhairle/blob/527e242e13bb235502584ff036bb7b67b6ba229a/api/src/routes/events/breakout.rs?plain=1#L73

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
Comment (dtodorowa) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:1112]: i think there is a chance for this to come up as a request later, when more people are using the feature. But wouldn't really sweat it now

ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte
Comment (dtodorowa) [reply ui/packages/comhairle/src/routes/(public)/conversations/[conversation_id]/events/[event_id]/live/+page.svelte:257]: noice!

dtodorowa (approved):

Issue #857, opened 2026-08-11
Prioritization: make optional open-text obviously optional (show "(optional)")


dtodorowa:
## Summary

The open-text question on each principle is already optional in logic, but nothing signals that to the participant, so it feels like they must write something for all 9 principles. Make it obviously optional.

## Desired behaviour

- Show **"(optional)"** next to the open-text question label. (Chosen approach — simplest and clearest.)
- Optionally consider only revealing the text box when the participant clicks/taps to respond, but the label treatment is the priority for launch.

## Where it lives

`ui/packages/comhairle/src/lib/tools/prioritization/components/QuestionField.svelte` — the label at `<p class="text-base font-medium">{question.text}</p>`. Text-type questions (`question.type.kind === 'text'`) are the optional ones; likert/continuous are required. `PrioritizationUser.svelte` already treats only non-text questions as required (`requiredQuestions`).

## Acceptance

- Text questions render an "(optional)" affordance on the label.
- Required (likert/continuous) questions are unchanged.

PR #927, opened 2026-08-18
857 prioritization make optional open text obviously optional show optional

dtodorowa:
Two participant-facing changes on top of prioritization-scroll-ux.

Text questions say they are optional

Text answers were already optional, but nothing told the participant. QuestionField now appends "(optional)" to the label of text questions. ([#857](https://github.com/crownshy/comhairle/issues/857))

Reading width and layout

The proposal card ran the full page width while StepHeader sat at 896px, so the two never lined up. The participant view is now wrapped in max-w-200 (800px), which covers the progress bar, card and nav buttons together.

Inside the card:

- Proposal text is left-aligned and fills the box, rather than sitting centred at prose's 65ch default.
- The title is left-aligned. .pile sets place-items: center, so it needed justify-self-start to opt out.
- The section body and its questions always stack. This was a grid lg:grid-cols-2, and lg: is a viewport breakpoint, so at 1440px it would still have split the 800px container in two and put the questions beside the text rather than below it.

Global questions
image

Per-section questions
image

### DIFF
--- a/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
@@ -433,7 +433,7 @@
 									{#if toolConfig.sectionQuestions.length > 0}
 										<div class="space-y-6">
 											{#if section.body}
-												<div class="prose max-w-none text-base">
+												<div class="prose text-base">
 													<ContentRenderer content={section.body} />
 												</div>
 											{/if}
@@ -538,7 +538,7 @@
 					{#each current.sections as section (section.id)}
 						<div class="space-y-6">
 							{#if section.body}
-								<div class="prose max-w-none text-base">
+								<div class="prose text-base">
 									<ContentRenderer content={section.body} />
 								</div>
 							{/if}
--- a/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
+++ b/ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
@@ -40,6 +40,9 @@
 	const toolConfig = $derived(
 		api.resolveToolConfig<string>(workflowStep, conversation.isLive ?? false)
 	);
+	/** Whether the participant can return to this step later — drives the "come
+	 * back to review more" reassurance when they move on before finishing. */
+	const canRevisit = $derived(workflowStep.canRevisit ?? false);
 
 	let proposals = $state<LocalizedProposal[]>([]);
 	let answers = $state<Record<string, Record<string, number | string>>>({}); // proposalId → questionId → value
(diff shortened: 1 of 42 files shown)


ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationUser.svelte
```diff
@@ -368,292 +368,300 @@
...
-{:else if allDone}
-	<div class="space-y-6">
-		<div class="space-y-1 text-center">
-			<!-- <CheckCircle2 class="text-primary mx-auto h-10 w-10" /> -->
-			<h2 class="text-l mt-5 font-semibold">Your answers</h2>
-			<p class="text-foreground text-sm">
-				Tap a proposal to review or adjust your answers. Changes are saved when you
-				continue.
-			</p>
+<div class="mx-auto w-full max-w-200">
```
ADSmith-0, line 371: It might be good to do `prose` here instead of `max-w-200`? Just thinking that doing it based on pixels instead of character width means that on smaller or larger screens the sizing might be off, idk what do you think?

dtodorowa: I think `max-w-200` isn't pixels, it should compile to `calc(var(--spacing) * 200)` which is 50rem, so it already scales with root font size (same as `max-w-4xl` on StepHeader).

Also prose isn't a width utility, it's the full typography stylesheet applied to every child, so it'd restyle the progress bar, radio cards and textarea too.

Maybe I misunderstood and you mean `max-w-prose` = 65ch?

dtodorowa: Just checked with claudio: 65ch is 689px in Montserrat vs 800px here, so not a huge gap.

My concern in general is that ch tracks the font itself, so if we ever swap Montserrat (let's say we have a partner who wants to have comic sans hahah as their font later on) the container will resize on its own. Keeping it rem also keeps it in proportion with the step header maybe? idk

dtodorowa: hmm, actually! the proposal text inside is prose `max-w-none`, so it has no measure cap at all right now. 

Maybe we could cap the text at 65ch and leave the container at 800px. What do you think @ADSmith-0

ADSmith-0: Yeah just basically because we don't have a `max-width: 65ch` class, I'm not sure about what we have available to us from tailwind and all that kind of stuff, so I thought prose would work alright because I checked and it has that built in

Yeah so I think with rem would it would be based on the font height but not the sentence length, which is what we're wanting right?

dtodorowa: good call @ADSmith-0 
added prose. let's merge

ADSmith-0: > Just checked with claudio: 65ch is 689px in Montserrat vs 800px here, so not a huge gap.
> 
> My concern in general is that ch tracks the font itself, so if we ever swap Montserrat (let's say we have a partner who wants to have comic sans hahah as their font later on) the container will resize on its own. Keeping it rem also keeps it in proportion with the step header maybe? idk

Yeah I think that would be the issue right? If we're aiming for like 60ch-80ch and then we swap out the font, then it might be 100ch per line or 40ch per line instead something like that

> hmm, actually! the proposal text inside is prose `max-w-none`, so it has no measure cap at all right now.
> 
> Maybe we could cap the text at 65ch and leave the container at 800px. What do you think @ADSmith-0

Yeah that's a really good idea! I think that would be good! :slightly_smiling_face:

ADSmith-0 (approved):

PR #952, opened 2026-08-19
[Survey Insights] Should be able to handle group questions

### LINKED ISSUE #888: [Survey Insights] Should be able to handle group questions
### Description
Currently group questions (questions inside of a group) are ignored on the survey insights, this is because their formatting is quite odd, so we'll need to look into how to fix this and get it working, this was spun off from #766 

### Acceptance criteria
- [ ] Survey insights should be able to handle group quetstions

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/888

### Testing
Screenshot from 2026-08-20 10-30-59


api/src/tools/heyform.rs
```diff
@@ -729,16 +786,7 @@ pub fn build_survey_insights(
...
                 .unwrap_or(&Vec::new())
-                .iter()
-                .fold(0u32, |acc, submission| {
-                    let val = submission.value.to_string();
-                    // Empty values from HeyForm
-                    if val != "\"\"" && val != "{\"value\":[]}" {
-                        return acc + 1;
-                    }
-
-                    acc
```
ADSmith-0, line 740: Removed this for this check here, https://github.com/crownshy/comhairle/pull/952/changes#diff-9709a9d130c0ce85c9b49cd0c23684ae4039dc420373e2fdff7640e4f3ce028cR575, that way we can do the check while going through the submissions instead of needing another loop

ui/packages/comhairle/src/lib/reports/survey/insights-loader.ts
```diff
@@ -224,9 +230,6 @@ function transformNonChoiceData(insight: InsightQuestion): NonChoiceQuestion['an
 					})
 					.filter((s) => !!s.trim()) ?? []
 			);
-
-		default:
-			return [];
```
ADSmith-0, line 229: Removed this so that we get TS errors if we miss a switch arm

ui/packages/comhairle/src/lib/reports/survey/SurveyInsights.svelte
```diff
@@ -37,7 +38,7 @@
 {:else}
 	{#each data as section (section.id)}
 		{#if isValidQuestion(section)}
-			<div class="py-10">
+			<div class="py-10" {@attach handleNested(section.properties?.parent)}>
```
ADSmith-0, line 41: Just to explain this because I'm not sure if everyone has seen it

Basically I was trying to work out how to do nested questions without modifying the data itself because it felt like that was going to become a real pain, I went with an attachment and I'm pretty happy with how it turned out

Attachments in svelte are essentially component-level lifecycle methods, but it also gives you access to the DOM element, so you can modify it however you want https://svelte.dev/docs/svelte/@attach

So I did an attachment factory here https://github.com/crownshy/comhairle/pull/952/changes#diff-4202fc0c1d1b8e7e178c28be446e31ef6e3fdcd7f48929dcc8ed09a6eff614e8R1, how it works is that it will get the parent section (or create one if it doesn't exist) and then put the element inside of it

alpinegeorge (approved):

PR #953, opened 2026-08-19
671/learning translations

rosswilson2306:
Motiviation:

- allow learning assistant to support extended list of languages

Actions:

- pass `prompt.cross_languages` param to ragflow when creating chat from model creation
- replace hardcoded text with paraglide translations in learning assistant UI


ADSmith-0 (approved):
Just out of interest what do the `cross_languages` do?

rosswilson2306:
> Just out of interest what do the `cross_languages` do?

@ADSmith-0 Sorry if I over-explain this, I don't know how familiar you are with ragflow or RAG systems in general, but basically the chat bots (learning assistants) are all connected to datasets which contain documents that the chat bot can then only answer questions on. If the dataset doesn't contain documents related to the question it won't try and answer and hallucinate incorrect statements. 

The documents will be chunked and parsed in whatever language the document is in. Enabling `cross_languages` basically allows you to ask questions in languages that are different from the parsed documents. I think it translates the question before it does the semantic search. So it means that we can have documents in English and ask questions in pashto, and vice versa have documents in pashto and ask questions in English, and it should still be able to generate responses based solely on the dataset documents.

It seems to work to a degree, but the translated responses do seem to be a little ropey.

ADSmith-0:
> unked and parsed in whatever language the document is in. Enabling `cross_languages` basically allows you to ask questions in languages that are different from the parsed document

Oh awesome, that's really cool! :smile: 

No I know nothing about RAG/ragflow so thank you for taking the time to explain it! :smile:

PR #961, opened 2026-08-19
Separate OpenAPI spec generation into new application

### LINKED ISSUE #960: Cannot generate OpenAPI spec without launching full server
Due to startup order, we currently can't generate our OpenAPI spec without our full server infrastructure in-place.
This is not an issue for local development, however poses an issue when publishing our api client package.

alpinegeorge:
Motivation:

 - We would like to be able to generate the open api spec without having to spin up the full server + database + redis stack.

Actions:

 - Create lightweight api_spec_gen application to generate OpenAPI spec.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -140,12 +140,9 @@ pub struct Args {
 async fn health_check() -> &'static str {
     "OK"
 }
-
-pub async fn setup_server(state: Arc<ComhairleState>) -> Result<Router<()>, ComhairleError> {
-    let args = Args::try_parse().unwrap_or_default();
-
-    tracing::info!("Running with config {:#?}", state.config);
-
+/// Constructs the ApiRouter and extracts the OpenAPI spec.
+/// Note that sub-routers like `routes::auth::router` are async and must be `.await`ed.
+pub async fn build_app_and_spec(state: Arc<ComhairleState>) -> (Router, OpenApi) {
     aide::generate::on_error(|error| {
         println!("{error}");
     });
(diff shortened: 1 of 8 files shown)


api/src/lib.rs
```diff
@@ -322,6 +315,35 @@ pub async fn setup_server(state: Arc<ComhairleState>) -> Result<Router<()>, Comh
...
+    (router, api)
+}
+
+/// Generates the OpenAPI spec without connecting to a live database or instantiating real services.
+pub async fn generate_api_spec() -> Result<OpenApi, ComhairleError> {
+    // Construct a state populated with dummy zeroed values for schema inspection
+    // FIXME: GH 2026-08-19 - The API spec should not depend on runtime state.
+    let dummy_state = Arc::new(unsafe { dummy::create_dummy_state() });
+
+    let (_, api_spec) = build_app_and_spec(dummy_state.clone()).await;
```
rosswilson2306: Would it make sense to move this and the other `unsafe` rust code to the `api_spec_gen` crate? Unsafe rust it's something I'm very familiar with but I guess it's all fine as that's the only place it's being used. I'm just thinking it might be better to keep it all isolated to avoid someone accidentally calling this function elsewhere.

Or is this just a temporary work around?

alpinegeorge: It is a temporary workaround since our API should not depend on runtime state, including config. I do also agree that moving it to the api_spec_gen crate would be best though. I think I can move the dummy module at the same time

rosswilson2306 (approved):

PR #983, opened 2026-08-21
671/thinking space translations

rosswilson2306:
Motivation:

- thinking space should work in supported languages

Actions:

- refactor thinking space to use TranslatableJson macro to support translatable question text fields
- update admin UI
- add paraglide translations for supported languages
- migration binary for thinking space steps


api/src/bin/migrate_prioritization_trans.rs
alpinegeorge, line 1: How do we run this, and the other binary, in production?

ui/packages/comhairle/src/lib/reports/thinking-space/ThinkingSpaceInsights.svelte
```diff
@@ -12,6 +12,8 @@
 
 	let numParticipants = $derived(insights.users.length);
 	let averageNumFollowUps = $derived.by(() => {
+		if (numParticipants === 0) return 0;
```
alpinegeorge, line 15: Good catch

alpinegeorge (approved):
The code looks good.
If checking out the repo and running the binaries in prod works, then I'd say ship it.

api/src/bin/migrate_prioritization_trans.rs
rosswilson2306: They basically just rely on having a `DATABASE_URL` so I was imagining Stuart running against the prod / demo / staging etc servers as he is the only one with access to those dbs. I imagined with an ssh tunnel or something it should be simple to do...

alpinegeorge: That sounds reasonable, ship it :)

PR #1000, opened 2026-08-24
Add custom event links to events

### LINKED ISSUE #988: Event admins should be able to set custom video link
Currently we only default to using the built-in Jitsi video call link.
However, some users may wish to use different software for video conferences.
We should allow them to set a custom value, defaulting to using Jitsi if a custom video call link is not provided.

alpinegeorge:
Motivations:

 - Users do not always want to use Jitsi for online events, but we don't currently support other meeting softwares.

Actions:

 - Add a custom event link field to the events table that, when populated, redirects the user to an alternate meeting URL instead of the default Jitsi meeting.


ADSmith-0 (approved):
Looks good to me! 🙂 Happy if you want to wait for Ross to double-check the backend stuff though as my Rust knowledge isn't the best

rosswilson2306 (approved):

PR #1018, opened 2026-08-26
363 instrumentation

rosswilson2306:
Actions:

+ add instrumentation where missing in routes and models

Motivation:

+ better log coverage

### DIFF
--- a/api/src/models/conversation.rs
+++ b/api/src/models/conversation.rs
@@ -439,7 +439,7 @@ pub async fn delete(
     Ok(conversation)
 }
 
-#[instrument(err(Debug))]
+#[instrument(err(Debug), skip(db))]
 pub async fn get_by_id_or_slug(
     db: &PgPool,
     id_or_slug: &IdOrSlug,
@@ -451,7 +451,7 @@ pub async fn get_by_id_or_slug(
     Ok(conversation)
 }
 
-#[instrument(err(Debug))]
+#[instrument(err(Debug), skip(db))]
 pub async fn get_localised_by_id_or_slug(
     db: &PgPool,
     id_or_slug: &IdOrSlug,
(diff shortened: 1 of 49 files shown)


alpinegeorge (dismissed):
What output does instrumenting `db: &PgPool` result in? I normally skip it, but perhaps I shouldn't?

rosswilson2306:
> What output does instrumenting `db: &PgPool` result in? I normally skip it, but perhaps I shouldn't?

Yeah, right enough. It outputs this:
```
db: Pool { size: 3, num_idle: 3, is_closed: false, options: PoolOptions { max_connections: 10, min_connections: 0, connect_timeout: 30s, max_lifetime: Some(1800s), idle_timeout: Some(600s), test_before_acquire: true } }, 
```
Which isn't incredibly useful to see everywhere and just causes bloat in the logs I guess. I'll remove that in the models.

api/src/models/conversation.rs
```diff
@@ -408,6 +408,7 @@ impl ConversationOrderOptions {
     }
 }
 
+#[instrument(err(Debug), skip(bot_service))]
```
alpinegeorge: skip db + bot_service

api/src/models/conversation.rs
```diff
@@ -711,6 +714,7 @@ impl CreateConversation {
...
```
alpinegeorge: skip db + bot_service

api/src/models/workflow.rs
```diff
@@ -106,6 +106,7 @@ pub async fn get_by_id(db: &PgPool, id: &Uuid) -> Result<Workflow, ComhairleErro
     Ok(conversation)
 }
 
+#[instrument(err(Debug), skip(state))]
 pub async fn launch(
```
alpinegeorge: nit: We're passing database and state here? Odd

alpinegeorge (changes requested):

api/src/models/conversation.rs
rosswilson2306: Good catch!

api/src/models/workflow.rs
rosswilson2306: Yeah, looks like state was added in later and db was kept in. I'll fix.

alpinegeorge (approved):

PR #1030, opened 2026-08-31
Migrate demographics to own table

### LINKED ISSUE #858: Spike: promote demographic/profile questions out of conversation metadata into a reusable data model
## Goal

Timeboxed investigation. Work out how to model demographic / profiling questions as **first-class, reusable data** instead of ad-hoc conversation metadata. Come back with a proposed data-model shape, the scoping + consent questions answered, and a rough estimate. Not a finished feature.

## Where this comes from

From the work with Bloom. We think it generalizes to comhairle: rather than each Conversation stashing "who are you" questions and answers in its own metadata blob, we want a shared model so a user answers a demographic question **once** and we can reuse the answer. Because we expect to reuse it, we don't want to leave it on conversation metadata; we want to promote it to a proper entity.

Note: civic_os has a Participants tab (demographics + recruitment goals) this is what we are referring to

Image

## The pieces to model

1. **Annotated demographic questions.** A question carries a semantic annotation of *what it is about* (age, gender, location, income, ...), not just free text. So "How old are you?" is typed as the `age` attribute. That annotation is what lets us reason across questions and reuse answers.

2. **Org-level profile schema.** Probably define the profile at the **Organization** level: what kind of user profile is this org after, which attributes matter. Open whether attributes/questions are org-scoped, drawn from a shared global taxonomy, or both.

3. **Inheritance / ask-once.** If a given user has already answered an attribute, don't ask again. Answers persist against the user and are inherited across conversations. Big open question: **scope** of inheritance (per-org vs cross-org) and the consent implications of reusing an answer a user gave in one context inside another.

4. **Overwrite vs additional.** When a conversation needs demographics, what are the semantics: reuse the stored answer as-is, overwrite / refresh it, or ask **additional** new questions on top? Precedence and merge rules.

5. **Attribute dictionary.** We may still want a controlled vocabulary / taxonomy that maps question annotations onto canonical attribute keys, so "age", "years old", etc. all resolve to one attribute. That is the "maybe this is just a dictionary" note, scoped to demographics.

## What the spike should answer

- **Data model.** What are the entities and where do they hang? Candidate shape: an attribute **dictionary** (the "what is this question about" taxonomy), an org-level **profile schema** selecting attributes, and a **user-scoped answer store** (`user_id`, `attribute`, `value`, `source`, `captured_at`). Does it live in `data_model/`, `api/`, or is it partly per-tool adaptor?
- **Scope of reuse / inheritance.** Per-org vs cross-org vs global. What is the default, and how does a conversation opt in.
- **Consent & privacy.** Demographic data is sensitive. Reusing an answer given to org A inside org B (or even conversation A into conversation B) needs a consent story: what the user sees / agrees to, whether they can edit or withdraw an answer.
- **Overwrite / merge precedence** rules (from point 4).
- **Relationship to existing tools.** HeyForm already collects arbitrary answers. Is the demographic layer built on top of HeyForm question annotations, or independent?

## Deliverable

A short write-up on this issue: proposed entity model (a sketch or diagram is fine), the inheritance + consent decision, and a rough size. An ADR if we land a direction. No implementation here.

## Related

- #848 (auto-apply glossary to HeyForm) as a shape precedent: define once at a higher level, reuse everywhere.
- #859 (store a per-conversation colour scheme) — sibling spike; the "where do reusable things live, not on conversation metadata" question is shared.

alpinegeorge:
Motivations:

 - We currently have a fixed set of demographics associated with a user profile.
 - We would like to be able to set these demographics for arbitrary categories.

Actions:

 - Add user settings section for user demographics.
 - Migrate the demographics data out of user_profile into new demographics tables.
 - Modify user profile to use new demographics tables.
 - Modify participant report generation to use new demographics tables.

### DIFF
--- a/api/src/routes/demographics.rs
+++ b/api/src/routes/demographics.rs
@@ -267,9 +267,3 @@ pub fn router(state: Arc<ComhairleState>) -> ApiRouter {
         )
         .with_state(state.clone())
 }
-
-#[cfg(test)]
-mod tests {
-    // GH TODO: Tests for the demographics routes would go here.
-    // There isn't much to test here yet, as the routes are primarily thin wrappers around the model functions.
-}
--- a/api/src/routes/demographics.rs
+++ b/api/src/routes/demographics.rs
@@ -1,0 +1,275 @@
+use std::sync::Arc;
+
+use aide::axum::ApiRouter;
+use aide::axum::routing::{delete_with, get_with, post_with, put_with};
+use axum::Json;
+use axum::extract::{Path, Query, State};
+use hyper::StatusCode;
+use tracing::instrument;
+use uuid::Uuid;
+
+use crate::ComhairleState;
+use crate::error::ComhairleError;
+use crate::models::demographics::{
+    self, ConversationDemographics, ConversationDemographicsFilterOptions,
+    CreateConversationDemographics, CreateDemographicsQuestion, CreateDemographicsResponse,
+    DemographicsQuestion, DemographicsQuestionsFilterOptions, DemographicsResponse,
+    DemographicsResponsesFilterOptions, PartialDemographicsQuestion, PartialDemographicsResponse,
+};
+use crate::models::pagination::{PageOptions, PaginatedResults};
+
+// ============================================================================
+// Conversation-demographics associations - CR.D
+// ============================================================================
+
+/// Get all associations between conversations and demographics questions, with optional filters for conversation ID and question slug.
+#[instrument(err(Debug), skip(state))]
+pub async fn get_conversation_demographics(
+    State(state): State<Arc<ComhairleState>>,
+    Query(filters): Query<ConversationDemographicsFilterOptions>,
+    Query(page_options): Query<PageOptions>,
+) -> Result<(StatusCode, Json<PaginatedResults<ConversationDemographics>>), ComhairleError> {
+    demographics::get_conversation_demographics(&state.db, filters, page_options)
+        .await
+        .map(|results| (StatusCode::OK, Json(results)))
+}
+
+/// Create a new association between a conversation and a demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn create_conversation_demographics(
+    State(state): State<Arc<ComhairleState>>,
+    Json(payload): Json<CreateConversationDemographics>,
+) -> Result<(StatusCode, Json<ConversationDemographics>), ComhairleError> {
+    demographics::create_conversation_demographics(&state.db, payload)
+        .await
+        .map(|result| (StatusCode::CREATED, Json(result)))
+}
+
+/// Remove an association between a conversation and a demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn delete_conversation_demographics(
+    State(state): State<Arc<ComhairleState>>,
+    Json((conversation_id, question_slug)): Json<(Uuid, String)>,
+) -> Result<(StatusCode, Json<Option<ConversationDemographics>>), ComhairleError> {
+    demographics::delete_conversation_demographics(&state.db, conversation_id, question_slug)
+        .await
+        .map(|result| (StatusCode::OK, Json(result)))
+}
+
+// ============================================================================
+// Demographics questions - CRUD
+// ============================================================================
+
+/// Get a demographics question by its slug.
+#[instrument(err(Debug), skip(state))]
+pub async fn get_demographics_questions(
+    State(state): State<Arc<ComhairleState>>,
+    Query(filters): Query<DemographicsQuestionsFilterOptions>,
+    Query(page_options): Query<PageOptions>,
+) -> Result<(StatusCode, Json<PaginatedResults<DemographicsQuestion>>), ComhairleError> {
+    demographics::get_demographics_questions(&state.db, filters, page_options)
+        .await
+        .map(|result| (StatusCode::OK, Json(result)))
+}
+
+/// Create a new demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn create_demographics_question(
+    State(state): State<Arc<ComhairleState>>,
+    Json(payload): Json<CreateDemographicsQuestion>,
+) -> Result<(StatusCode, Json<DemographicsQuestion>), ComhairleError> {
+    demographics::create_demographics_question(&state.db, payload)
+        .await
+        .map(|result| (StatusCode::CREATED, Json(result)))
+}
+
+/// Update a demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn update_demographics_question(
+    State(state): State<Arc<ComhairleState>>,
+    Path(question_slug): Path<String>,
+    Json(payload): Json<PartialDemographicsQuestion>,
+) -> Result<(StatusCode, Json<DemographicsQuestion>), ComhairleError> {
+    demographics::update_demographics_question(&state.db, question_slug, payload)
+        .await
+        .map(|result| (StatusCode::OK, Json(result)))
+}
+
+/// Delete a demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn delete_demographics_question(
+    State(state): State<Arc<ComhairleState>>,
+    Path(question_slug): Path<String>,
+) -> Result<(StatusCode, Json<Option<DemographicsQuestion>>), ComhairleError> {
+    demographics::delete_demographics_question(&state.db, question_slug)
+        .await
+        .map(|result| (StatusCode::OK, Json(result)))
+}
+
+// ============================================================================
+// Demographics responses - CRUD
+// ============================================================================
+
+/// Get all responses for a specific demographics question.
+#[instrument(err(Debug), skip(state))]
+pub async fn get_demographics_responses(
+    State(state): State<Arc<ComhairleState>>,
+    Query(filters): Query<DemographicsResponsesFilterOptions>,
+    Query(page_options): Query<PageOptions>,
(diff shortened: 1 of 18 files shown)


api/src/routes/demographics.rs
```diff
@@ -0,0 +1,275 @@
...
+                        .response::<200, Json<Option<DemographicsResponse>>>()
+                }))
+                .with_state(state.clone())
+        )
+        .with_state(state.clone())
+}
+
+#[cfg(test)]
+mod tests {
+    // GH TODO: Tests for the demographics routes would go here.
```
rosswilson2306: @alpinegeorge Are you planning to add before this is merged or after? Or not bother? I agree if there isn't much functionality in the endpoints the model tests cover the main functionality and this is just plumbing.

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
```diff
@@ -0,0 +1,175 @@
...
+			schemaShape[q.slug] = z.string().nullable().optional();
+		}
+
+		return {
+			initialData,
+			schema: z.object(schemaShape)
+		};
+	};
+
+	const config = buildFormConfig();
```
rosswilson2306: @alpinegeorge If this depends on the props, which look like they are invalidated below, should this use `$derive()` or `$derived.by()`?

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
```diff
@@ -0,0 +1,175 @@
...
+			await Promise.all(apiCalls);
+
+			notifications.send({
+				message: 'Demographics updated successfully',
+				priority: 'INFO'
+			});
+			$errors = {};
+
+			// Re-run the page load function to sync fresh responses into the UI state
+			await invalidateAll();
```
rosswilson2306: @alpinegeorge We should probably scope this to the `load` function it depends on rather than use `invalidateAll`.

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
```diff
@@ -0,0 +1,175 @@
...
+
+	let {
+		questions,
+		responses,
+		userId
+	}: {
+		questions: DemographicsQuestion[];
+		responses: DemographicsResponse[];
+		userId: string;
+	} = $props();
```
ADSmith-0: Would it be possible to split the type out to `Props` please? I think it's a bit easier to read ([style_guide](https://github.com/crownshy/comhairle/blob/staging/STYLE_GUIDE.md#naming:~:text=Name%20the%20props%20type))

ui/packages/comhairle/src/routes/(public)/settings/UserDemographicsForm/UserDemographicsForm.svelte
```diff
@@ -0,0 +1,175 @@
...
+
+	const { form: formData, validateForm, errors } = form;
+
+	// Helper to format slugs like "political_party" into "Political Party"
+	const formatSlugLabel = (slug: string) => {
+		return slug
+			.split('_')
+			.map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+			.join(' ');
+	};
```
ADSmith-0: I think this can use the `snakeToSentenceCase()` util from `utils/casingUtils.ts`

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
```diff
@@ -0,0 +1,175 @@
...
+
+		if (!result.valid) {
+			$errors = result.errors;
+			return;
+		}
+
+		try {
+			saving = true;
+
+			// 2. Iterate dynamically over the available questions to dispatch API calls
```
ADSmith-0: I think we said we would prefer `tryCatchAsync` ([style_guide](https://github.com/crownshy/comhairle/blob/staging/STYLE_GUIDE.md#error-handling:~:text=Prefer%20tryCatchAsync)), if you're happy to change it over?

It might also be beneficial to handle the errors individually, like "failed to create demographics", "failed to update demographics", "failed to delete demographics", it might be easier for us if we have to debug in the future? Although I don't know much about this feature so that might be over the top 🙂

ui/packages/comhairle/src/routes/(public)/settings/UserDemographicsForm/UserDemographicsForm.svelte
ADSmith-0, line 1: Do we plan on using this elsewhere btw? Because it might be better to colocate it if we're only using it in the settings ([style_guide](https://github.com/crownshy/comhairle/blob/staging/STYLE_GUIDE.md#keep-files-small-and-single-purpose:~:text=Colocate%20by%20default))

ui/packages/comhairle/src/routes/(public)/settings/+page.svelte
```diff
@@ -3,12 +3,15 @@
...
 	import UserDetailsForm from '$lib/components/UserDetailsForm/UserDetailsForm.svelte';
+	import UserDemographicsForm from '$lib/components/UserDemographicsForm/UserDemographicsForm.svelte';
 	import UpgradeAccountModal from '$lib/components/UpgradeAccountModal/UpgradeAccountModal.svelte';
 	import type { UserDto } from '@crownshy/api-client/api';
 
 	let { data }: PageProps = $props();
 	let participation = $derived(data.participation);
 	let user = $state(data.user) as UserDto;
+	let demographicQuestions = $derived(data.demographicQuestions);
+	let demographicResponses = $derived(data.demographicResponses);
```
ADSmith-0: Not a huge deal but we can do this in a one-liner if we want with [JS destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring#:~:text=const%20%7B%20a%2C%20b%2C%20c%20%7D%20%3D%20obj%3B)

```ts
const { partipication, user, demographicQuestions, demographicResponses }	 = $derived(data);
```

api/src/routes/demographics.rs
alpinegeorge: I just removed the tests module. No point since it's all covered in the models::demographics module anyway.

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: Moved!

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: Done, but it would be good if you could check I've done it correctly.

ui/packages/comhairle/src/routes/(public)/settings/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: I added `displayName` to the question data model, so I'm using that instead now.

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: I've switched to tryCatchAsync. I'll leave the more verbose error handling for when we implement a non-dev-UI for it.

ui/packages/comhairle/src/routes/(public)/settings/+page.svelte
alpinegeorge: Sorted :)

ui/packages/comhairle/src/routes/(public)/settings/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: Moved :)

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
alpinegeorge: I've switched to using `$derived.by()`

ui/packages/comhairle/src/lib/components/UserDemographicsForm/UserDemographicsForm.svelte
ADSmith-0: Yeah that's perfect! :slightly_smiling_face:

ui/packages/comhairle/src/routes/(public)/settings/UserDemographicsForm/UserDemographicsForm.svelte
ADSmith-0: Just wanted to check if we've switched to the `displayName` does that mean we can get rid of the `formatSlugLabel` function?

alpinegeorge: Yes. Thank you. I'll do that now

alpinegeorge: Done

ADSmith-0 (approved):

PR #1031, opened 2026-08-31
947/refresh token

### LINKED ISSUE #947: Refresh JWT
Implement refresh functionality for auth tokens, for both admin and public.

rosswilson2306:
Motivation:

- allow user sessions to be refreshed without having to log in every day when `auth-token` expires

Actions:

- migration to add `refresh_token` table with associated model
- `refresh-token` cookie added to signup / login endpoints
- updates to logout endpoint to revoke and remove refresh-token
- `/refresh` endpoint which rotates refresh tokens and keeps user session alive by issuing new auth-token
- refactor logout functionality to run via api instead of svelte server
- updates to `handleFetch` hook and `createApiClient` to refresh and retry requests failing on 401


api/src/test_helpers.rs
```diff
@@ -330,36 +330,57 @@ impl UserSession {
             username: Some(username.to_owned()),
             password: Some(password.to_owned()),
             email: Some(email.to_owned()),
-            cookie: None,
+            cookies: None,
+        }
+    }
+
+    pub fn cookie_header(&self) -> Option<String> {
```
alpinegeorge, line 337: nit: Could we use [CookieJar](https://docs.rs/axum-extra/latest/axum_extra/extract/cookie/struct.CookieJar.html) for this instead?

api/src/test_helpers.rs
```diff
@@ -728,7 +737,8 @@ impl UserSession {
 
         let user: HashMap<String, Option<Value>> = serde_json::from_value(value)?;
 
-        self.cookie = cookie.clone();
+        // TODO:
```
alpinegeorge, line 740: Leftover TODO

api/src/test_helpers.rs
rosswilson2306: Yeah, I think using `CookieJar` would probably be better. I'm going to make a tech debt ticket to look into this though. I had a quick bash at changing it over and the cookies weren't getting stored across the requests and I don't want to go down a rabbit hole with it right now.

alpinegeorge (approved):

api/src/routes/auth.rs
```diff
@@ -1135,6 +1216,62 @@ pub fn create_session_cookie<'a>(user: &User, state: &Arc<ComhairleState>) -> Co
...
+        .user(user)
+        .secret(&state.config.refresh_jwt_secret)
+        .custom_claims(RefreshClaims {
+            jti: token_record.id,
+        })
+        .duration(chrono::Duration::days(7))
+        .call();
+
+    Cookie::build((REFRESH_KEY, refresh_token))
+        .path("/") // TODO: can this be more tightly scoped
```
alpinegeorge, line 1234: nit: I don't think this should block merge, but have we got an issue to handle this in the near future?

PR #1036, opened 2026-09-01
1033/prioritization insights fixes

rosswilson2306:
Motivation:

- improvements to prioritization insights

Actions:

- change num participants to use num participants who voted on any proposal instead of average who voted on all
- use max and min possible score for ranked proposal bar domain
- fix width of alignment question select in setup tab

### DIFF
--- a/ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationRankedProposalTable.svelte
+++ b/ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationRankedProposalTable.svelte
@@ -5,76 +5,109 @@
 	import type { RankedProposal } from '@crownshy/api-client/api';
 	import Crown from '$lib/components/icons/Crown.svelte';
 	import DivergingProgress from '../DivergingProgress.svelte';
+	import type { ToolConfig } from '$lib/tools/prioritization';
 
 	type Props = {
 		proposals: RankedProposal[];
+		toolConfig: ToolConfig<string>;
 	};
-	let { proposals }: Props = $props();
+	let { proposals, toolConfig }: Props = $props();
 
 	const ROW_ANIMATION_DURATION_MS = 1000;
 	let rowLimit = $state(4);
 	let visibleRows = $derived(proposals.slice(0, rowLimit));
-	let maxRating = $derived(proposals.length > 0 ? proposals[0].alignmentRating : 0);
-	let minRating = $derived(
-		proposals.length > 0 ? proposals[proposals.length - 1].alignmentRating : 0
+	let alignmentQuestion = $derived(
+		toolConfig.questions.find((question) => question.id === toolConfig.alignmentQuestionId)
 	);
-	let progressDomain = $derived([minRating, maxRating]);
+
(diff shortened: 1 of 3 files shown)


ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationRankedProposalTable.svelte
```diff
@@ -5,76 +5,109 @@
...
+			return minValue * proposals.length;
+		}
+
+		if (alignmentQuestion.type.kind === 'continuous') {
+			return alignmentQuestion.type.minValue * proposals.length;
+		}
+
+		return 0;
+	});
+	let progressDomain = $derived([minPossibleScore, maxPossibleScore]);
```
ADSmith-0: It looks like we're only using the `minPossibleScore` and `maxPossibleScore` to calculate the `progressDomain`? In which case we could do it all in a oner if we wanted

```ts
const progressDomain = $derived.by(() => {
  if (!alignmentQuestion) {
    return { min: 0, max: 0 };
  }

  switch (alignmentQuestion.type.kind) {
    case "likert": {
      const values = alignmentQuestion.type.categories.map((cat) => cat.value);
      return { min: Math.min(...values), max: Math.max(...values);
    }
    case "continuous":
      return { min: alignmentQuestion.type.minValue * proposals.length, max: alignmentQuestion.type.maxValue * prosposals.length };
    default:
      return { min: 0, max: 0 };
  }
});
```

In fact for

```ts
const values = alignmentQuestion.type.categories.map((cat) => cat.value);
return { min: Math.min(...values), max: Math.max(...values);
```

You could probably then do it in 1 for loop and grab the values that way, but it doesn't matter too much either way

ui/packages/comhairle/src/lib/tools/prioritization/PrioritizationManage.svelte
```diff
@@ -360,15 +360,23 @@
...
 					{/if}
 				</Select.Trigger>
 				<Select.Content>
 					{#each questions as question (question.id)}
-						<Select.Item value={question.id}>{question.text.localized}</Select.Item>
+						<Select.Item value={question.id}
+							>{question.text.localized.length > 40
+								? question.text.localized.slice(0, 40).concat('...')
+								: question.text.localized}</Select.Item
+						>
```
ADSmith-0, line 379: For this hunk and the one above and the  it might be worth adding in a `title` attribute so that when a user hovers over it they can see the full text of the question?

rosswilson2306: Good shout!

ui/packages/comhairle/src/lib/reports/prioritization/PrioritizationRankedProposalTable.svelte
rosswilson2306: Right enough. I'll update.

ADSmith-0 (approved):

alpinegeorge (approved):

PR #1039, opened 2026-09-02
994/user annon code

rosswilson2306:
Motiviation:

- allow email local can be used as username for otp signups if none provided, instead of random code

Actions:

- migration to add annon_code column to user table, remove unique constraint from username and replace with search index and unique index on annon_code
- model and endpoint updates
- frontend updates for annon users
- refactor otp user creation to use email local for username if none provided


rosswilson2306:
Will create a new PR changing annon to guest in the codebase and db.

PR #1041, opened 2026-09-02
994/guest user code

rosswilson2306:
Motivation:

- allow otp users to be created with email local as username if none provided to give more human friendly username than random code
- rename anonymous users to guest

Actions:

- migration to:
- - add guest_code column to comhairle_user table
- - copy anon usernames into guest_code
- - update annon auth_type to guest
- - remove unique index on username and re-apply appropriate indexes
- model, endpoint and test updates
- frontend updates for guest login /signup flow
- frontend updates to rename anonymous to guest
- paraglide translation updates to rename anonymous to guest
- refactor otp signup to use email local for username if none provided


api/migrations/20260902091503_add_guest_code_to_comhairle_user_table.sql
```diff
@@ -0,0 +1,24 @@
...
+SET guest_code = username
+WHERE auth_type = 'annon';
+
+-- Update auth_type = 'annon' to 'guest'
+UPDATE comhairle_user
+SET auth_type = 'guest'
+WHERE auth_type = 'annon';
+
+-- Remove unique index on username column
+DROP INDEX IF EXISTS username_index;
```
alpinegeorge, line 16: Why remove the unique username constraint? Surely we don't want username collisions for non-guest users

rosswilson2306: That's actually the reason for the ticket. When we sign up otp users via event invites, the only user data we have available is their email address. The idea originally was to use the email local as the username, but we couldn't do this with because of the unique index on the username. So we went with the same pseudo-random code for usernames as anonymous users. 

This meant that users invited to events would appear in jitsi meetings with the random code usernames, which just isn't that user friendly for moderators. Using the email local would be an improvement.

The main reason we had the unique index was for the anonymous users to sign in. I think for otp and email passwords we would usually get by `email`. However, I just realised I should remove the `get_user_by_username` method as that won't work. I don't think it's being used after this change.

alpinegeorge: Would this work:
```sql
CREATE UNIQUE INDEX username_index 
ON comhairle_user (username) 
WHERE auth_type != 'guest';
```

rosswilson2306: Yeah, that would work. I'll update.

alpinegeorge (approved):

Issue #991, opened 2026-08-24
Admin interface Optimization:  Prevent premature loading of data


stuartlynn:
There are places in the admin interface where we are loading data where we dont need to / prematurely. This can make the interface more sluggish than it needs to be. We should ensure that data is being fetched in layout.ts or page.ts that is triggered just before it is needed.

I asked claude to audit places where this is happening as a starting point 

----

# Performance Audit: Premature / Eager Data Loading

Cases where a load function (`+layout.ts` / `+page.ts` / `.server.ts`) fetches data
before it is needed, blocking navigation. Focus: admin conversation subtree, where
one layout load blocks *every* child route.

The hotspot is `admin/conversations/[conversation_id]/+layout.ts` — it fetches several
things in a `Promise.all` that only the `configure` page (or nothing) consumes, yet
they block navigation to `design`, `monitor`, `report`, `events`, `invites`,
`moderate`, and `learning-assistant`.

---

### 1. `stats` — fetched, never used (delete)

- **File:** `src/routes/(admin)/admin/conversations/[conversation_id]/+layout.ts:77-79` (return at `:89`)
- **Loads:** `GetConversationWorkflowStats`, blocks page load.
- **Premature because:** `stats` is returned from the layout but consumed by **no**
  child route. The `monitor` page that needs stats already refetches it itself
  (`monitor/+page.ts:9`). Layer-level fetch is pure redundancy.
- **Fix:** Delete the fetch (lines 45-49, 77-79) and drop `stats` from the return
  object (line 89). Removes 1 blocking call from every conversation admin route.

### 2. `media`, `usersWithPermission`, `cohostOrganizations` — layout-wide, configure-only (push down)

- **File:** `src/routes/(admin)/admin/conversations/[conversation_id]/+layout.ts:52-74, 87-96`
- **Loads:**
  - `GetMedia` (`:54`) — when conversation has an image
  - `ListUsersWithPermission` (`:67-73`) — only when user owns conversation
  - `ListConversationCoHostOrganizations` (`:41`, in the parallel fetch)
- **Premature because:** all three are consumed only by the configure page
  (`configure/+page.svelte:65-67`). Every other child route pays for them on nav.
- **Fix:** Move all three fetches into `configure/+page.ts` (which already loads
  `availableDocuments`). Remove from the layout `Promise.all` (`:38-42`) and return
  object (`:87-96`). Removes 3 blocking calls from design/monitor/report/events/
  invites/moderate/learning-assistant.

[...]

PR #1095, opened 2026-09-03
991/5 - Admin interface Optimization: configure/team

ADSmith-0:


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/team/+page.server.ts
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/team/+page.server.ts:1]: Server side check to check if the user is allowed to view the teams page, for example if they typed in the address bar a `/team` we should block it if they don't have permission

It had to be a `server` page because if you did it as a `+page.ts` then you get a temporary blank page every time you navigate to it

The issue is that the `{ api }` we're passing through is only on the `+page.ts` side, which is different to the `+page.server.ts`, so I can't access it

I've had to use fetches for now, but I think it would be nice in the future to make an interface that we can just use anywhere, I've had some ideas about how we could do that so I'll write a ticket about it 🙂

ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/team/+page.svelte
```diff
@@ -0,0 +1,239 @@
...
+				})}
+			{:else}
+				<section class="flex flex-col gap-8 py-8">
+					<h3 class="text-xl font-semibold">Permissions</h3>
+
+					<form method="POST" class="flex flex-col gap-4" use:enhance>
+						<Form.Field
+							form={userEmailPermissionForm}
+							name="email"
+							class="flex flex-col gap-2"
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/team/+page.svelte:138]: Removed the `ResourceRoleForm` and the `TeamManager` as it appears these are the only places in the code base that we're using them, so I just inlined it here to make it a bit simpler

I'm not sure if we're likely to use it elsewhere in the future? But if we do then we can always move it back

PR #1096, opened 2026-09-03
991/6 - Admin interface Optimization: configure cleanup

ADSmith-0:

Removing unused files that were replaced by 991/1 to 991/5


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/InfoHover.svelte
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/configure/InfoHover.svelte:1]: All the changes to this file are from the `ExampleDialog` it was only used in this file and was fairly small so I inlined it

ui/packages/comhairle/playwright.config.ts
```diff
@@ -14,13 +14,13 @@ dotenv.config({ path: '.playwright.env' });
...
 	/* Run tests in files in parallel */
-	fullyParallel: true,
+	fullyParallel: false,
 	/* Fail the build on CI if you accidentally left test.only in the source code. */
 	forbidOnly: !!process.env.CI,
 	/* Retry on CI only */
 	retries: process.env.CI ? 2 : 0,
 	/* Opt out of parallel tests on CI. */
-	workers: process.env.CI ? 1 : undefined,
+	workers: 1,
```
ADSmith-0: Running the tests in parallel messes stuff up because they're accessing the same UI elements so it leads to race conditions

ui/packages/comhairle/src/lib/utils/casingUtils.ts
```diff
@@ -6,12 +6,17 @@ export function camelToSentenceCase(str: string) {
 	return str.replace(/[A-Z]/g, (letter) => ` ${letter.toLowerCase()}`);
 }
 
-export function snakeToSentenceCase(str: string) {
+export function snakeToStartCase(str: string) {
 	return str
 		.replace(/^[-_]*(.)/, (_, c) => c.toUpperCase()) // Initial char (after -/_)
 		.replace(/[-_]+(.)/g, (_, c) => ' ' + c.toUpperCase());
 }
```
ADSmith-0: Changed the name here to more closely match the functionality https://en.wikipedia.org/wiki/Letter_case#:~:text=Start,-case

PR #1109, opened 2026-09-04
991/13 - Admin interface Optimization: event/invites page

ADSmith-0:


ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/invites/external.ts
```diff
@@ -0,0 +1,3 @@
+import InviteLink from '../../../InviteLink.svelte';
+
+export { InviteLink };
```
Comment (ADSmith-0) [inline ui/packages/comhairle/src/routes/(admin)/admin/conversations/[conversation_id]/events/[event_id]/invites/external.ts:3]: Something I was considering here, I thought that "../../../" type imports don't look great and are kind of hard to read so I thought maybe we could do a file called external and then have the parent imports there? But I'm on the fence about it, I'd be interested to hear what people think?

PR #1125, opened 2026-09-08
671 tool config translation migrations

rosswilson2306:
Actions:

+ migration to convert legacy prioritization and thinking space tool_configs from simple text fields to translations system, while keeping snapshots of legacy tool_configs in case of roll backs

Motivation:

+ prevent conversation serialization issues caused by legacy text content not deserializing into UUIDs

### DIFF
--- a/api/migrations/20260908081428_convert_thinking_space_to_translation_system.sql
+++ b/api/migrations/20260908081428_convert_thinking_space_to_translation_system.sql
@@ -1,0 +1,152 @@
+-- Convert legacy inline thinking-space text fields into
+-- text_content/text_translation records, replacing inline strings with
+-- text_content UUIDs inside tool_config / preview_tool_config JSONB.
+
+-- Add temporary legacy backup columns, populated only for rows this
+-- migration will touch, so an incident can be rolled back with a single
+-- UPDATE without needing a proper down-migration.
+ALTER TABLE workflow_step
+    ADD COLUMN IF NOT EXISTS legacy_tool_config jsonb,
+    ADD COLUMN IF NOT EXISTS legacy_preview_tool_config jsonb;
+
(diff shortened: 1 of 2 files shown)


api/migrations/20260908081428_convert_thinking_space_to_translation_system.sql
```diff
@@ -0,0 +1,152 @@
...
+    v_text_tc_id uuid;
+    v_intent_tc_id uuid;
+    v_root_questions jsonb := '[]'::jsonb;
+BEGIN
+    v_topic_tc_id := pg_temp.ts_new_translation(p_locale, p_config->>'topic', 'plain');
+
+    FOR v_question IN
+        SELECT * FROM jsonb_array_elements(COALESCE(p_config->'root_questions', '[]'::jsonb))
+    LOOP
+        v_topic_tc_id := pg_temp.ts_new_translation(p_locale, v_question->>'text', 'plain');
```
alpinegeorge: I think this should be `v_text_tc_id:= ...`

alpinegeorge (changes requested):

api/migrations/20260908081428_convert_thinking_space_to_translation_system.sql
rosswilson2306: Tremendous spot!

alpinegeorge: Yes, but I missed the repeat a couple of lines after!
Here:
`'text', to_jsonb(v_topic_tc_id),` on line 78

alpinegeorge (approved):

rosswilson2306:
@stuartlynn Did you get a chance to test this against a prod snapshot?

PR #1130, opened 2026-09-08
991/23 - Admin interface Optimization: clean up

ADSmith-0:


ui/packages/comhairle/src/routes/(admin)/admin/+page.ts
```diff
@@ -1,9 +0,0 @@
-import type { PageLoad } from './$types';
-
-export const load: PageLoad = async ({ parent }) => {
-	const parentData = await parent();
-	const ownedConversations = parentData?.ownedConversations ?? { records: [] };
-	const permittedConversations = parentData?.permittedConversations ?? { records: [] };
-
-	return { ownedConversations, permittedConversations };
-};
```
ADSmith-0, line 9: Gets pulled through via parent data we don't need to re-export here

ui/packages/comhairle/src/routes/(public)/+page.ts
```diff
@@ -1,8 +0,0 @@
-import type { PageLoad } from './$types';
-
-export const load: PageLoad = async ({ parent }) => {
-	const parentData = await parent();
-	return {
-		isCommunity: parentData.isCommunity
-	};
-};
```
ADSmith-0, line 8: Same here, the gets pulled through via parent data we don't need to re-export here

PR #1136, opened 2026-09-10
Change string demographics to multiple choice

alpinegeorge:
Motivations:

 - Most of our demographics are multiple choice, and so should not accept any invalid string. Though, we do have some fields like postal codes which should still be able to.

Actions:

 - Allow an admin to define multiple_choice options that define what response values will be accepted.

### DIFF
--- a/api/src/models/demographics.rs
+++ b/api/src/models/demographics.rs
@@ -3,8 +3,8 @@
 // (
 //    slug TEXT PK,                               - Unique identifier for the demographics question
 //    display_name TEXT NOT NULL,                 - Human-readable name for the demographics question
-//    response_type 'string' | 'number' NOT NULL, - The type of response expected for the demographics question
-//    bucket_config JSONB                         - Configuration for how responses should be bucketed (e.g., age ranges)
+//    response_type 'number' | 'string' NOT NULL, - The type of response expected for the demographics question
+//    bucket_config JSONB                         - For 'number': bucket ranges (e.g. age ranges). For 'string': the configurable set of selectable options.
 // )
 // 2. `demographics_response`                     - Represents a user's response to a demographics question
 // (
@@ -30,6 +30,7 @@ use tracing::instrument;
 use uuid::Uuid;
 
 use crate::error::ComhairleError;
+use crate::models::SqlxResultExt;
 use crate::models::pagination::{PageOptions, PaginatedResults};
 
 // ============================================================================
@@ -97,13 +98,13 @@ impl ConversationDemographicsFilterOptions {
     }
 }
 
-/// Represents demographics question response type (either 'string' or 'number').
+/// Represents demographics question response type (either 'number' or 'string').
 #[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 #[cfg_attr(test, derive(PartialEq))]
 pub enum DemographicsQuestionResponseType {
-    String,
     Number,
+    String,
 }
 
 impl Into<SimpleExpr> for DemographicsQuestionResponseType {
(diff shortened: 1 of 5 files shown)


api/src/models/demographics.rs
```diff
@@ -473,6 +527,26 @@ pub async fn delete_conversation_demographics(
...
+    let (sql, values) = sea_query::Query::select()
+        .columns(DEMOGRAPHICS_QUESTION_COLUMNS)
+        .from(DemographicsQuestionIden::Table)
+        .and_where(Expr::col(DemographicsQuestionIden::Slug).eq(slug))
+        .build_sqlx(PostgresQueryBuilder);
+
+    let question = sqlx::query_as_with::<_, DemographicsQuestion, _>(&sql, values)
+        .fetch_one(db)
+        .await
+        .map_err(|e| ComhairleError::DatabaseError(e))?;
```
rosswilson2306: @alpinegeorge I made a trait to help with error handling in the models as `DatabaseError` comes through to the frontend as 500 so isn't very helpful. For cases like this it will account for `not in db -> 404` https://github.com/crownshy/comhairle/blob/3f14a8f1b38d7cf6e0eee1e260ef71da60fe8a41/api/src/models.rs?plain=1#L93

We've started standardising this elsewhere so might be worth updating here.

ui/packages/comhairle/src/lib/components/AdminNav.svelte
```diff
@@ -162,6 +163,17 @@
...
+					<!-- <SideBar.MenuItem>
+						<SideBar.MenuButton>
+							{#snippet child({ props: btnProps })}
+								<a {...btnProps} href="/admin/dev/demographics">
+									<FlaskConical class="size-4" />
+									Demographics (dev)
+								</a>
+							{/snippet}
+						</SideBar.MenuButton>
+					</SideBar.MenuItem> -->
```
rosswilson2306: @alpinegeorge Is this meant to stay commented out?

ui/packages/comhairle/src/routes/(admin)/admin/dev/demographics/+page.svelte
```diff
@@ -0,0 +1,779 @@
...
+			};
+		}
+		if (responseType === 'string' && options.length > 0) {
+			return { type: 'string' as const, options };
+		}
+		return undefined;
+	}
+
+	// --- New question form -------------------------------------------------
+	let newSlug = $state('');
```
rosswilson2306: @alpinegeorge Is this ever going to be a page accessible by non-devs? If so, could it use superform to be consistent with other forms?

api/src/models/demographics.rs
alpinegeorge: Will do. Thank you for that!

ui/packages/comhairle/src/lib/components/AdminNav.svelte
alpinegeorge: I was going to leave it commented out for testing in dev until we get a proper design and further the discussion of using demographics on the comhairle side. I can remove it and the developer demographics page entirely if need be

ui/packages/comhairle/src/routes/(admin)/admin/dev/demographics/+page.svelte
alpinegeorge: We'll probably need some more design work before that happens, so I imagine we'd be rewriting the whole page and possibly splitting it up into separate sub-pages within conversations at that time.

api/src/models/demographics.rs
alpinegeorge: Implemented!

ui/packages/comhairle/src/lib/components/AdminNav.svelte
alpinegeorge: Moved dev UI changes to the "dev-demographics-ui"

ui/packages/comhairle/src/routes/(admin)/admin/dev/demographics/+page.svelte
alpinegeorge: Moved dev UI changes to the "dev-demographics-ui"

rosswilson2306 (approved):

PR #1152, opened 2026-09-11
991/24 - Admin interface Optimization: moderation page

ADSmith-0:


ui/packages/comhairle/src/lib/tools/types.ts
```diff
@@ -1,9 +1,9 @@
-import type { ToolConfig, WorkflowStepWithTranslations } from '@crownshy/api-client/api';
+import type { ToolConfig, WorkflowStepWithTranslationsDto } from '@crownshy/api-client/api';
```
ADSmith-0, line 1: This type doesn't exist anymore in the API

Issue #1080, opened 2026-09-03
Configurable list of rejection reasons, required before launch


dtodorowa:
Let the policy team define the list of reasons a statement can be rejected for, and require the list to exist before a conversation launches. **P3.**

There is a list today, hardcoded: five preset labels in `ui/packages/comhairle/src/lib/tools/polis/polis-moderation/rejectReasons.ts`, stored verbatim into `polis_statement_aux.moderation_reason` alongside an optional free-text note (ADR-0015). Nobody outside the codebase can change it.

This ask goes further in two ways, and both are the point:

**Customisable rather than hardcoded.** The policy team writes the reasons, so rejections are consistent between moderators and across conversations, and the reasons can be counted. Today's list is a reasonable default to start from, not something every engagement should be stuck with.

**Required before launch.** Deciding the rules while already rejecting statements is how moderation becomes arbitrary. Making the list a launch prerequisite forces the policy conversation to happen first, which is the actual benefit.

## What this needs

- Somewhere in the admin interface to edit the list of rejection reasons.
- The moderation tab reading from that list rather than the constant.
- A launch check that the list is set.

## To settle

- Platform default list, per organisation, per conversation, or a default that conversations extend. The current five are the obvious seed for whatever the default is.
- Is picking a reason mandatory on every reject? Mandatory is better data and slows bulk moderation down, which is the same tension #834 raised.
- Does bulk reject take one shared reason?
- What happens to reasons already recorded when the list changes. `moderation_reason` stores the label as text, so editing the list does not rewrite history, which is right for the audit trail and awkward for counting.
- Do these reasons ever reach the participant whose statement was rejected? That decision belongs with #1046 and #1052, and it changes how the text needs to be written: a reason written for an audit log reads badly to the person it is about.
- What blocking launch looks like. A hard gate, or a warning in a pre-launch checklist alongside the strict moderation setting in #1042.

## Related

- #901 moderation logic epic, #892 and #899 (CSV export, which is where a controlled list pays off).
- #834 the closed free-text version.

PR #1162, opened 2026-09-15
Feat/moderation policy and statement export

### LINKED ISSUE #892: Export moderated statements (incl. rejection reasons) as CSV
Follow-up to #834 (optional reason field on reject).

Once moderators can record *why* a statement was rejected, they need to get that out of the platform. Today conversation owneres have to log rejection by hand in an docs. The Polis Report is no help here: rejected statements never enter voting, so they never appear in report data.

## What

A CSV export of the moderated statements for a Polis step, so the platform can replace the hand-kept log.

Rather than a reject-only export, do one export of **all** statements with their moderation state, which the moderator can filter to just rejects in a spreadsheet. Suggested columns:

- statement id
- statement text
- moderation status (accepted / rejected / pending / seeded)
- moderation reason (populated on rejects; also carries the auto reason the split flow writes)
- is_seed
- lineage (original / replaced-by), if cheap

## Notes

- Data is already on `polis_statement_aux` (`moderation_status`, `moderation_reason`). This is mostly an export endpoint + a download button in the moderation UI.
- Not for participants; admin/moderator only.
- Split/reword already writes its own `moderation_reason` (ADR-0011), so those rows come through for free.

## Priority

Flagged at the 1 September planning day as important for the next SG use case, in the #1047 bundle. **P1.** The policy team is keeping the rejection log by hand today, which is the thing this removes. #899 has the scope decision and needs settling first.

### LINKED ISSUE #899: Decide what the moderated-statements CSV export includes
Scoping question for the CSV export in #892.

The export is meant to replace hand-kept rejection logs, which means we need to pin down **which rows and columns** actually go in the file:

- **Which statuses?** Accepted only, or everything (pending / accepted / rejected)? For a rejection log you'd want rejected rows *and* their `moderation_reason` at minimum. Likely: all statuses, so the log is complete.
- **Which columns?** At least statement text, status, and `moderation_reason`. Do we also want `polis_statement_id`, themes, submitted-by, timestamps?
- **Seeds and split-originals?** Do moderator seeds and the auto-rejected originals from a split show up in the export, or get filtered out? The split original is rejected with an auto-generated reason ("Reworded/split by moderator into N statement(s)"), so it *will* have a reason.

Data source is `polis_statement_aux` (has `moderation_status`, `moderation_reason`, `is_seed`, `original_statement_id`). Just need a decision on scope before #892 is built.

Flagged at the 1 September planning day as important for the next SG use case, in the #1047 bundle, so this decision blocks #892. **P1.**

Note #1080, which makes the rejection reason list configurable. If that lands, a reason column becomes countable rather than free text, which changes what the export is for.

dtodorowa:
Adds a moderation policy to each conversation: the list of reasons a moderator can pick when rejecting a statement. The policy team edits it in Configure > Moderation policy. The Moderation tab gets a Download CSV button that exports every statement, with the reject reason split back out so rejections can be counted.

The design is written up in ADR-0037. It's marked as a proposal, so shout if anything in there looks off.

## What's in it

- **Moderation policy editor** in Configure. Each reason is a label plus an optional description of what counts under it. With nothing saved, a conversation uses the five default reasons. The first edit makes the list its own, and "Reset to default" goes back. It lives in `conversation.metadata.moderation_policy`, so there's no migration.
- **Reason picker** in the reject popover is now a searchable list, with each description under its label. A reject still stores `"Label: note"` in `moderation_reason`, same as before.
- **CSV export** of every statement, seeds and split originals included. Columns: `statement_id`, `created_at`, `statement_text`, `moderation_status`, `is_seed`, `themes`, `reject_reason`, `reject_note`, `edited_from_statement_id`, `replaced_by_statement_ids`.
- **Shared autosave** for the glossary and moderation policy editors. Saves run one at a time now, so an older save can't land last and overwrite a newer list. This also fixes the glossary editor, which had the same code.
- `csv.ts` is shared with the Insights export, which now gets the same formula escaping and byte order mark.

## Left out on purpose

- **No launch check.** #1080 asked for one, but it would bring back the launch gating ADR-0001 removed.
- **No server export endpoint.** The CSV is built in the browser from the rows the Moderation tab already loads. Worth revisiting if that list ever gets paginated.
- **No per-status download.** Filtering the file by status is in a follow-up PR stacked on this one.

## Known limits

- Counting by reason only works while labels stay the same. After a rename, older rejections show the old wording in `reject_note` with an empty `reject_reason`.
- A label can't contain ": " because that's what separates the label from the note. The editor flags it and doesn't save it.

## Trying it

1. Edit or add a reason in Configure > Moderation policy.
2. Go straight to a Polis step's Moderation tab and reject a statement with that reason.
3. Hit Download CSV and check the `reject_reason` and `reject_note` columns.

Closes #892
Closes #899
Part of #1080


dtodorowa:
Split into three smaller stacked PRs: autosave, moderation policy, statement export.

PR #1164, opened 2026-09-15
Per-conversation moderation policy for reject reasons

dtodorowa:
Stacked on `fix/configure-autosave-order`. Review that one first.

Adds a moderation policy to each conversation: the list of reasons a moderator can pick when rejecting a statement. The policy team edits it in Configure > Moderation policy, and the reject popover offers those reasons.

The design is written up in ADR-0037. It's marked as a proposal, so shout if anything in there looks off. The ADR also covers the CSV export, which is in the next PR.

## What's in it

- **Moderation policy editor** in Configure. Each reason is a label plus an optional description of what counts under it. With nothing saved, a conversation uses the five default reasons. The first edit makes the list its own, and "Reset to default" goes back. It lives in `conversation.metadata.moderation_policy`, so there's no migration.
- **Reason picker** in the reject popover is now a searchable list, with each description under its label. A reject still stores `"Label: note"` in `moderation_reason`, same as before.
- **Label checks.** Repeated labels and labels containing ": " are flagged in the editor and not saved. ": " is what separates the label from the note in the stored reason.
- The conversation layout load reads the policy once, so the Moderation tab picks up an edit without a reload.

## Left out on purpose

- **No launch check.** #1080 asked for one, but it would bring back the launch gating ADR-0001 removed.

## Known limits

- Renaming or deleting a reason doesn't touch rejections already recorded. They keep the old wording.

## Trying it

1. Edit or add a reason in Configure > Moderation policy.
2. Go straight to a Polis step's Moderation tab and reject a statement.
3. The new reason should be in the picker, with its description under it.

Part of #1080


stuartlynn:
Lets merge this in so we dont delay it's use for the upcoming engagement. I think we want to make moderation policies their own actual entity on the backend however as it feels like a core component here.

stuartlynn (approved):

PR #1193, opened 2026-09-16
Fixed configure/details page tests

### LINKED ISSUE #1179: Fix e2e tests for configure/details
### Description
Currently the e2e tests for the details page fail on the headers, we should fix them

ADSmith-0:
Fixes https://github.com/crownshy/comhairle/issues/1179


ui/packages/comhairle/e2e/utils/navigation/Conversation.ts
```diff
@@ -29,7 +29,6 @@ const Conversation = {
 
 		// Make conversation
 		await page.getByRole('button', { name: 'New conversation', exact: true }).click();
-		await page.getByText('Start from blank').click();
```
ADSmith-0, line 32: Duplicate identifier so it was breaking when it was called

dtodorowa (approved):

Issue #1195, opened 2026-09-17
Use per-conversation permission checks on step, translation, event and media routes


dtodorowa:
Org members can now open and edit a conversation shared with their org (see the linked PR), but they still can't edit its steps unless they're also a site-wide Admin.

That's because a bunch of routes use `RequiredAdminUser`, which only checks the site-wide Admin role and never looks at the conversation:

- `api/src/routes/workflow_steps.rs`
- `api/src/routes/workflows.rs`
- `api/src/routes/translations.rs`
- `api/src/routes/events.rs` (and `events/breakout.rs`)
- `api/src/routes/media.rs`
- `api/src/tools/polis.rs`, `prioritization.rs`, `thinking_space.rs`
- some handlers in `api/src/routes/conversations.rs`

It cuts both ways: an org member without the site role gets a 401 on step edits, and any site Admin can edit any conversation's steps by id.

## What to do

Switch the conversation-scoped handlers to `RequiredUser` plus `authorize(&state, &user, Action::ConversationUpdate, &resource)` (or `ConversationRead` for reads), the same way `update_conversation` and the moderation policy routes already work. Keep `RequiredAdminUser` only on routes that really are site-wide (creating a conversation, the media library, email template configs).

## Done when

- A member of a co-hosting or primary host org who isn't a site Admin can edit steps, translations and events on that conversation.
- A user with no access to the conversation gets a 401 on those routes, even if they're a site Admin. Super admins still bypass.
- Tests cover one route per file.

PR #1196, opened 2026-09-17
Let org members edit conversations shared with their org

dtodorowa:
If you added someone to an org and shared a conversation with that org, they could see it in the sidebar, but opening it bounced them back to `/admin`. The admin layout loads the moderation policies, those routes needed `ConversationUpdate`, and the co-host role only gave `ConversationRead`. Even without the redirect they couldn't edit anything unless the owner added them on the Team tab.

## Changes

- **Co-host role grants edit.** `Role::ConversationCoHost` now has `ConversationRead` + `ConversationUpdate`, so every member of a co-hosting org can open and edit. Org admins and members get the same access for now. If we want members to be view-only later, drop `ConversationUpdate` from that role. Co-hosts still can't manage co-hosts or delete (`ConversationAdmin`).
- **Primary host org counts as a co-host.** `can_perform_resource_action` checks `conversation.organization_id` against the user's org when no granted role allows the action, and `list_for_permitted_user` includes those conversations. Nothing gets written to `resource_permissions`, so there's no backfill.
- **Moderation policy reads need `ConversationRead`.** Create, replace and delete still need `ConversationUpdate`.
- The Team tab is unchanged. We can remove it later.
- ADR-0040 and CONTEXT.md updated (the co-host role was documented as read-only).

## Not in this PR

Step, translation, event, media and workflow routes still use `RequiredAdminUser` (site-wide Admin role only), so an org member who isn't a site Admin can open the conversation and edit its details but not its steps. That's #1195.

## Testing

- New tests: co-host org member can update but not admin, primary host org member can read and update with no grant (and an unrelated org can't), primary host org conversations show up in the permitted list.
- `test_list_permissions_by_action` updated, since co-host now shows up for `conversation_update`.
- `cargo test` on `models::permissions`, `models::conversation`, `routes::conversations`, `routes::moderation_policies`: 57 pass. `should_be_able_to_get_conversation_with_media` fails, but it fails on staging too (needs bulk storage config).


dtodorowa:
closing, bc this is premature and we need to align on behaviour with the team

